PR #11, opened 2012-10-11
Yaml based answers file

GregSutcliffe:
Test using the installer class:

vi ./foreman-installer.yaml
echo include installer | puppet apply -v --modulepath ./

### DIFF
--- a/README.debian
+++ b/README.debian
@@ -7,7 +7,8 @@ is a list of gotchas that will be needed to work with this module
 # Limitations
 
 At present, this will install foreman with sqlite3 support. Using a different database
-is left as a post-install task exercise for the reader.
+is left as a post-install task exercise for the reader. This is normally simply installing
+foreman-<dbtype> as a package and altering database.yml
 
 # Setup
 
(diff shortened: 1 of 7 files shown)


installer/manifests/init.pp
```diff
@@ -0,0 +1,15 @@
+# This class is called from the foreman install script (or Vagrantfile)
+# and expects a yaml file to exist in either $modulepath/foreman-installer.yaml
+# or /etc/foreman-installer.yaml
+#
+class installer {
+
+  $params=loadyaml('/etc/foreman-installer.yaml',
```
skottler: What do you think about putting the answers file into /etc/foreman-installer/answers.yaml or something along those lines? That will make it easier when we start doing releases of the installer modules so we can put the installer into /etc/foreman-installer.

GregSutcliffe:
I agree completely. I've moved everything under a foreman_installer module to keep it tidy. I had to change to an underscore as the hyphen is not permitted in class names. Seems to work for me.

PR #12, opened 2012-10-12
Use https instead of git for submodule protocol

GregSutcliffe:
Some users (FriedBob springs to mind) have requested this change, since their stricter network setup won't allow access to git:// URLs - I don't have a problem with that, so here's the pull request.

I've tested this with git clone --recursive https://github.com/GregSutcliffe/foreman-installer -b use_https

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,33 +1,33 @@
 [submodule "tftp"]
 	path = tftp
-	url = git://github.com/theforeman/puppet-tftp.git
+	url = https://github.com/theforeman/puppet-tftp.git
 [submodule "apache"]
 	path = apache
-	url = git://github.com/theforeman/puppet-apache.git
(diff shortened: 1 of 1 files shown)


llowder:
I tested this today, and it did allow me to clone the repos successfully, using the same command from the PR.

I am also known as FriedBob, so very excited about this PR.

skottler:
+1 since smart HTTPS is going to be the road forward for Github.

skottler:
I just tested upgrading from a git protocol based checkout to a https-based one. Since there isn't an direct upgrade path (apart from `sed` ;)) and git-based checkouts will continue to work just fine, I am going to merge this.

ekohl:
Actually there is: `git submodule sync`.

skottler:
Woah, that is seriously awesome @ekohl! Thanks for the tip.

PR #17, opened 2012-11-09
Throw error when YAML type is unknown, fix YAML location comment

domcleal:
I accidentally put a list into my YAML answers file and lack of errors stumped me for a while.

### DIFF
--- a/foreman_installer/manifests/yaml_to_class.pp
+++ b/foreman_installer/manifests/yaml_to_class.pp
@@ -1,7 +1,8 @@
 # Takes a key to lookup in the installation answers file
 # - If it's a hash, declare a class with those parameters
 # - If it's true or "true" declare the default parameters for that class
-# - Otherwise ignore it
+# - If it's false or "false" ignore it
+# - Otherwise fail with error
 #
 define foreman_installer::yaml_to_class (
   $classname = ''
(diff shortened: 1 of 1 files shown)


foreman_installer/manifests/init.pp
```diff
@@ -1,11 +1,12 @@
...
-# or /etc/foreman-installer.yaml
+# and expects a yaml file to exist at either:
+#   $modulepath/foreman_installer/answers.yaml
+#   /etc/foreman_installer/answers.yaml
 #
 class foreman_installer {
 
-  $params=loadyaml('/etc/foreman_installer/answers.yaml',
-                  "${settings::modulepath}/foreman_installer/answers.yaml")
+  $params=loadyaml("/etc/${module_name}/answers.yaml",
```
GregSutcliffe: Is that really necessary? There was debate over changing the module name, so that would then break the location of the user's answer file...

GregSutcliffe:
Looks good to me, apart from a minor comment. Testing now.

foreman_installer/manifests/init.pp
domcleal: Good point, I'll change it back to the fixed name.  The only niggle I have is the underscore in the path `/etc/foreman_installer` compared to `/etc/foreman-proxy` - despite it being the exact module name, do you think this should be standardised?

It might also be good to make foreman_installer parameterised itself, with these two paths becoming the default locations.

GregSutcliffe: I have no issue with /etc/foreman-installer/answers.yaml. I don't mind the parameterizaton either, but lets do that in a separate pull.

domcleal:
Refreshed the branch with only the YAML type detection bit, will resubmit the foreman_installer part.

GregSutcliffe:
Tests fine here with both valid and invalid answers files (Squeeze, Puppet3). Merging.

GregSutcliffe:
Merged in 81894ba371514834a

PR #23, opened 2012-11-30
Rename loadyaml to loadanyyaml to prevent stdlib conflict

domcleal:
loadyaml currently conflicts with puppetlabs-stdlib since they're different.

err: Could not retrieve catalog from remote server: Error 400 on SERVER: loadyaml(): wrong number of arguments (3; must be 1) at /etc/puppet/modules/production/foreman_installer/manifests/init.pp:11 on node foreman.fm.example.net


GregSutcliffe:
+1

PR #26, opened 2013-01-23
Version bump + yamlgen

GregSutcliffe:


generate_answers.rb
```diff
@@ -0,0 +1,169 @@
...
+# Bonus, per-module, questions defined up top where they're easy to find :)
+
+def foreman_questions
+  {
+    "Should Foreman require SSL to be enabled? (default: true) " \
+    => 'menu_helper("agree", "foreman", "ssl")',
+    "Should Foreman run under apache & passenger? (default: true) " \
+    => 'menu_helper("agree", "foreman", "passenger")',
+    "Should Foreman be installed from the unstable repo? (default: false) " \
+    => 'menu_helper("agree", "foreman", "use_testing")',
```
domcleal: This should change `repo` instead of `use_testing` now, ditto for proxy.

generate_answers.rb
```diff
@@ -0,0 +1,169 @@
...
+    => 'menu_helper("agree", "foreman_proxy", "tftp")',
+    "Should Foreman_proxy manage DNS? (default: false) " \
+    => 'menu_helper("agree", "foreman_proxy", "dns")',
+    "Should Foreman_proxy manage DHCP? (default: false) " \
+    => 'menu_helper("agree", "foreman_proxy", "dhcp")',
+  }
+end
+
+def puppet_questions
+  # this is only really here to provide a non-boolen example...
```
domcleal: Useful example, but probably best to remove it.

generate_answers.rb
GregSutcliffe: Yep, will replace this as the non-boolean exmple ;)

GregSutcliffe:
@domcleal done

PR #30, opened 2013-02-04
Add create_resources from puppetlabs to make this work on 2.6

GregSutcliffe:
New PR to merge this to develop instead of master

PR #31, opened 2013-02-05
Use Puppet's own create_resources function if available

domcleal:
Tested on 2.6.17 and 3.1.0.


ekohl:
Why not use the create_resources submodule?

domcleal:
@ekohl yep, we could have done that instead of GH-30.  I think this is better though, since we get to use Puppet's built in function if available (which may have bug fixes etc), while if we simply included the create_resources module then it overrides Puppet's own function.

ekohl:
@domcleal maybe fork create_resources to do that instead? Sounds like the wider community could benefit from it. If that's too much effort I agree that the function should be renamed.

domcleal:
@ekohl I was thinking the same, but I haven't found a way of doing it yet within a function called create_resources.  As far as I can see, renaming it is necessary.  The method to check whether a function is already defined is private, the only public method causes it to be autoloaded, which causes stack overflows etc.  Hence I renamed it so we can reliably check if a differently named method is present.

ekohl:
@domcleal fair enough.

GregSutcliffe:
Based on the above discussion, :+1: to merge

PR #33, opened 2013-02-12
Allow for modulepaths other than pwd

skottler:


generate_answers.rb
```diff
@@ -79,8 +79,10 @@ def puppetmaster_questions
 # helper methods
 
 def save_or_run_and_exit
+  # If the foreman_installer module exists then just use it. Otherwise, ask.
+  $modulepath = File.exists?("/usr/share/foreman-installer/foreman_installer") ? "/usr/share/foreman-installer" : ask("<%= color('Where are the installer Puppet modules?', :question) %>")
```
domcleal: The default might be better as `File.dirname(__FILE__)` so it works from a checkout too.

domcleal:
The $outfile also needs fixing as it's relative, shall I do that separately?

skottler:
I've updated $outfile and made that changes mentioned in the review. Thanks @domcleal!

domcleal:
Merged, thanks!

PR #38, opened 2013-03-08
Update gitmodules to point to OnyxPoint's concat module

GregSutcliffe:
My concat module was originally a fork of theirs, but they've done some pretty nice improvements since, so we should use theirs (if only so more people are aware of it as a native-type concat implementation :P)

Tested on Debian - produces the same concat output for Dns / Dhcp modules. Small PRs incoming for those modules as well.


domcleal:
:+1: tested, go ahead and merge.

PR #39, opened 2013-03-22
Add database management support

domcleal:
Adding the puppet-postgresql, puppetlabs-mysql, stdlib and concat modules.
The existing concat module is renamed to prevent a conflict and since it's
types/providers, the name doesn't matter.

Clarifying the licensing statement as these are Apache 2 licensed modules.

This depends on https://github.com/theforeman/puppet-foreman/pull/51.  It also renames a couple of submodules, which is a bit hairy, so be careful.  To test it, do the following:
1. fetch my puppet-foreman branch in the foreman submodule
2. fetch my foreman-installer branch here
3. run `git submodule sync`
4. run `git submodule update`
5. hope that git doesn't blow up

I've only tested this on RHEL6, so would appreciate some Debian and Ubuntu testing - though I expect both of the modules are in good shape for those systems.

Currently the postgresql submodule points at my fork, but I'll fix that once one more PR is merged into upstream for Puppet 2.6 compat.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -28,6 +28,18 @@
 [submodule "dns"]
 	path = dns
 	url = https://github.com/theforeman/puppet-dns
+[submodule "concat_native"]
+	path = concat_native
+	url = https://github.com/onyxpoint/pupmod-concat.git
+[submodule "mysql"]
+	path = mysql
+	url = https://github.com/puppetlabs/puppetlabs-mysql.git
+[submodule "postgresql"]
+	path = postgresql
+	url = https://github.com/puppetlabs/puppet-postgresql.git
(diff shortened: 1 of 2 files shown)


GregSutcliffe:
Tested on squeeze, both psql and mysql seem fine. I can do ubuntu tomorrow if an ack for that is wanted.

Addtional note for other testers, a `git submodule sync` seems to be required to get git to pickup the new submodules - otherwise they remain empty directories.

domcleal:
`git submodule init`, no?

GregSutcliffe:
Oops, yeah. Too much multitasking ;)

domcleal:
Merged as b3b6559.  Thanks!

PR #47, opened 2013-05-20
Add Modulefile and pre-reqs for Puppet Forge, add build script

domcleal:


GregSutcliffe:
Despite a certain amount of cu'n'paste in my comments on all the other Forge PRs, I've read the Modulefiles and they all look fine. `build_modules` is untested but reads ok to me.

Happy for all these to be merged on the basis that it's new content and we can roll forward if there are any small fixes. Ship it :)

domcleal:
@GregSutcliffe thanks for the reviews, all merged!

PR #49, opened 2013-05-23
adding spec and temp release script

lzap:
This is SCL enabled RHEL SPEC file and a temporary release script until we
sort out how to use tito with repositories with git submodules.

Here is a build: http://koji.katello.org/koji/taskinfo?taskID=36605

### DIFF
--- a/foreman-installer.spec
+++ b/foreman-installer.spec
@@ -1,0 +1,72 @@
+
+# uncomment to disable foreman-generate-answers script (and dependencies)
+#global skip_generator 1
+
+%if "%{?scl}" == "ruby193"
+    %global scl_prefix %{scl}-
+    %global scl_ruby /usr/bin/ruby193-ruby
+%else
+    %global scl_ruby /usr/bin/ruby
+%endif
+
+%global alphatag RC1
+
+Name:       foreman-installer
+Version:    1.2.0
+Release:    0.1.%{alphatag}%{?dist}
+Summary:    Puppet-based installer for The Foreman
+Group:      Applications/System
+License:    GPLv3+ and ASL 2.0
+URL:        http://theforeman.org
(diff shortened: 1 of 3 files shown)


foreman-installer.spec
```diff
@@ -0,0 +1,72 @@
...
+    %global scl_prefix %{scl}-
+    %global scl_ruby /usr/bin/ruby193-ruby
+%else
+    %global scl_ruby /usr/bin/ruby
+%endif
+
+%global alphatag RC1
+
+Name:       foreman-installer
+Version:    1.2.0
```
domcleal, line 15: This should be 1.2.9999 in develop.

domcleal:
Please target develop, not master.

foreman-installer.spec
```diff
@@ -0,0 +1,72 @@
...
+BuildArch:  noarch
+
+Requires:   %{?scl_prefix}ruby(abi)
+Requires:   %{?scl_prefix}puppet >= 0.24.4
+%if %{?skip_generator:0}%{!?skip_generator:1}
+Requires:   %{?scl_prefix}rubygem-highline
+%endif
+
+%description
+Script to generate answers for puppet-based installer of The Foreman life-cycle
```
domcleal, line 36: More than just a script, it's the installer too.

domcleal:
I just noticed a bug in our current foreman-installer package which affects this.  The package has epoch set to 1 due to a bug in fpm (https://github.com/jordansissel/fpm/issues/381), so we need to make sure it's set to 1 in this spec file too.

lzap:
aah sorry about PR into master - will close this PR and make a new one.

1.2.9999 - ok but with tito we will probably change this as tito modifies version number.

description fixed yeah - I just wanted not to have summary same as description because then rpmlink throws a warning :-)

epoch - oh right, the only way is to add Epoch: 1 then. adding this line

PR #50, opened 2013-05-27
adding spec and temp release script

lzap:
This is SCL enabled RHEL SPEC file and a temporary release script until we
sort out how to use tito with repositories with git submodules.

Here is a build: http://koji.katello.org/koji/taskinfo?taskID=36605

Replaces: https://github.com/theforeman/foreman-installer/pull/49

### DIFF
--- a/foreman-installer.spec
+++ b/foreman-installer.spec
@@ -1,0 +1,77 @@
+
+# uncomment to disable foreman-generate-answers script (and dependencies)
+#global skip_generator 1
+
+%if "%{?scl}" == "ruby193"
+    %global scl_prefix %{scl}-
+    %global scl_ruby /usr/bin/ruby193-ruby
+%else
+    %global scl_ruby /usr/bin/ruby
+%endif
+
+# set and uncomment all three to set alpha tag
+#global alphatag RC1
+#global dotalphatag .%{alphatag}
+#global dashalphatag -%{alphatag}
+
+Name:       foreman-installer
+Epoch:      1
+Version:    1.2.9999
+Release:    1%{?dotalphatag}%{?dist}
+Summary:    Puppet-based installer for The Foreman
+Group:      Applications/System
+License:    GPLv3+ and ASL 2.0
+URL:        http://theforeman.org
+Source0:    %{name}-%{version}%{?dashalphatag}.tar.gz
+
+%if 0%{?rhel} && 0%{?rhel} == 5
+BuildRoot:  %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
+%endif
+
+BuildArch:  noarch
+
+Requires:   %{?scl_prefix}puppet >= 0.24.4
+
+%if %{?skip_generator:0}%{!?skip_generator:1}
+Requires:   %{?scl_prefix}ruby(abi)
+Requires:   %{?scl_prefix}rubygem-highline
+%endif
+
+%description
+Complete installer for The Foreman life-cycle management system based on puppet and
+script to generate answers for puppet manifests.
+
+%prep
+%setup -q -n %{name}-%{version}%{?dashalphatag}
+
+%build
+echo "%{version}" > VERSION
+#replace shebangs for SCL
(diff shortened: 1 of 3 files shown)


foreman-installer.spec
```diff
@@ -0,0 +1,73 @@
...
+#global skip_generator 1
+
+%if "%{?scl}" == "ruby193"
+    %global scl_prefix %{scl}-
+    %global scl_ruby /usr/bin/ruby193-ruby
+%else
+    %global scl_ruby /usr/bin/ruby
+%endif
+
+%global alphatag RC1
```
domcleal: Can we remove all the %alphatag bits?  The release should just be 1%{?dist} for now, the RCs will be done on the stable branches only.

foreman-installer.spec
```diff
@@ -0,0 +1,73 @@
...
+URL:        http://theforeman.org
+Source0:    %{name}-%{version}-%{alphatag}.tar.gz
+
+%if 0%{?rhel} && 0%{?rhel} == 5
+BuildRoot:  %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
+%endif
+
+BuildArch:  noarch
+
+Requires:   %{?scl_prefix}ruby(abi)
```
domcleal: This should be in the skip_generator block too.

foreman-installer.spec
domcleal: Ah, I see it's needed for the source tarball name.  Hmm, that's a pain as it'll break when we do or don't have an RC.

foreman-installer.spec
lzap: Pushed, keeping the alphatag. Conforms Fedora Pacakging Guidelines.

LZ

On Mon, May 27, 2013 at 06:16:51AM -0700, Dominic Cleal wrote:

> > +Epoch:      1
> > +Version:    1.2.9999
> > +Release:    0.1.%{alphatag}%{?dist}
> > +Summary:    Puppet-based installer for The Foreman
> > +Group:      Applications/System
> > +License:    GPLv3+ and ASL 2.0
> > +URL:        http://theforeman.org
> > +Source0:    %{name}-%{version}-%{alphatag}.tar.gz
> > +
> > +%if 0%{?rhel} && 0%{?rhel} == 5
> > +BuildRoot:  %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
> > +%endif
> > +
> > +BuildArch:  noarch
> > +
> > +Requires:   %{?scl_prefix}ruby(abi)
> 
> This should be in the skip_generator block too.
> 
> ---
> 
> https://github.com/theforeman/foreman-installer/pull/50/files#r4400356

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

lzap:
Fixed, it is now develop-branch friendly.

domcleal:
Couple of things, the release + changelog number were out of sync and I had an issue with it putting double quotes in the release string and then rpmbuild failing, both fixed with this (if you're happy, squash it in):
https://github.com/domcleal/foreman-installer/commit/ccd5124e8557fa3e563222214ce1f636dc3ef57a

One last point, could you add an arg to release so we can specify a different dist?

Otherwise :+1:, thanks!

lzap:
Done, pushed, thanks!

LZ

On Tue, May 28, 2013 at 04:55:07AM -0700, Dominic Cleal wrote:

> Couple of things, the release + changelog number were out of sync and I had an issue with it putting double quotes in the release string and then rpmbuild failing, both fixed with this (if you're happy, squash it in):
> https://github.com/domcleal/foreman-installer/commit/ccd5124e8557fa3e563222214ce1f636dc3ef57a
> 
> One last point, could you add an arg to release so we can specify a different dist?
> 
> Otherwise :+1:, thanks!
> 
> ---
> 
> https://github.com/theforeman/foreman-installer/pull/50#issuecomment-18545748

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

domcleal:
Thanks for the updates @lzap, merged as e603d99!

PR #51, opened 2013-06-13
Fix db_type section name

domcleal:


GregSutcliffe:
Good catch! Merged, thanks.

PR #52, opened 2013-06-17
Version bump

domcleal:
Primarily to get the F18 support working.


ekohl:
Rather annoying that you can't easily diff the changes submodules, but :+1:.

PR #54, opened 2013-07-04
FQDN checks

GregSutcliffe:
A couple of usability commits - we have so many issues related to a missing FQDN, so I added a test for it. In testing it worked, I noticed the "You're all set!" message is printed even though it bombs out with the fqdn error, so I fixed that too.

### DIFF
--- a/generate_answers.rb
+++ b/generate_answers.rb
@@ -88,8 +88,12 @@ def save_or_run_and_exit
   # If the foreman_installer module exists then just use it. Otherwise, ask.
   $modulepath = File.exists?("#{$workdir}/foreman_installer") ? $workdir : ask("<%= color('Where are the installer Puppet modules?', :question) %>")
   if Process::UID.eid == 0 && agree("\n<%= color('Do you want to run Puppet now with these settings?', :question) %> (y/n)", false)
-    system("echo include foreman_installer | puppet apply --modulepath #{$modulepath} -v")
-    parting_message
+    system("echo include foreman_installer | puppet apply --detailed-exitcodes --modulepath #{$modulepath} -v")
+    if [1,4,6].include? $?.exitstatus
+      parting_error
+    else
+      parting_message
+    end
   else
     parting_message
   end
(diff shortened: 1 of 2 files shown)


domcleal:
Untested, but :+1: in principle.

ekohl:
:+1: to the principle.

GregSutcliffe:
Would one of you care to test it too? ;)

domcleal:
In a bit!

generate_answers.rb
```diff
@@ -102,6 +106,14 @@ def parting_message
   exit 0
 end
 
+def parting_error
+  say("\n <%= color('There was an error!', :red) %>")
+  say("\n Check the puppet error messages above, and <%= color('#{$outfile}', :cyan) %> for your config.")
+  say("\n Once you have fixed the problem, you can re-run it in the future as root with:")
+  say("  echo include foreman_installer | puppet apply --modulepath #{$modulepath} -v")
+  exit 0
```
domcleal: `exit 1`?

generate_answers.rb
```diff
@@ -89,7 +89,11 @@ def save_or_run_and_exit
   $modulepath = File.exists?("#{$workdir}/foreman_installer") ? $workdir : ask("<%= color('Where are the installer Puppet modules?', :question) %>")
   if Process::UID.eid == 0 && agree("\n<%= color('Do you want to run Puppet now with these settings?', :question) %> (y/n)", false)
     system("echo include foreman_installer | puppet apply --modulepath #{$modulepath} -v")
-    parting_message
+    if $?.exitstatus == 0
```
domcleal: Perhaps use `--detailed-exitcodes` and then check for either 1, 4 or 6 to indicate failure (or not 0 or 2)?  That way it'll also pick up on errors during the run and not just catalog compilation errors.

GregSutcliffe:
@domcleal sorry for the delay, updated with your suggestions

generate_answers.rb
```diff
@@ -99,6 +103,14 @@ def parting_message
   say("\n Okay, you're all set! Check <%= color('#{$outfile}', :cyan) %> for your config.")
   say("\n You can apply it in the future as root with:")
   say("  echo include foreman_installer | puppet apply --modulepath #{$modulepath} -v")
+  exit 1
```
domcleal: exit codes for sucess/error appear to be mixed up

GregSutcliffe: Curses, changed the wrong one ;)

GregSutcliffe:
@domcleal oops, fixed.

domcleal:
Thanks, merged!

PR #55, opened 2013-07-04
adding foreman_api dependency

lzap:
SSIA

### DIFF
--- a/foreman-installer.spec
+++ b/foreman-installer.spec
@@ -17,7 +17,7 @@
 Name:       foreman-installer
 Epoch:      1
 Version:    1.2.9999
-Release:    1%{?dotalphatag}%{?dist}
+Release:    2%{?dotalphatag}%{?dist}
 Summary:    Puppet-based installer for The Foreman
 Group:      Applications/System
 License:    GPLv3+ and ASL 2.0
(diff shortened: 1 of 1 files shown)


foreman-installer.spec
```diff
@@ -31,6 +31,7 @@ BuildRoot:  %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
 BuildArch:  noarch
 
 Requires:   %{?scl_prefix}puppet >= 0.24.4
+Requires:   rubygem-foreman_api
```
domcleal: Should have the scl_prefix, in case it is built for SCL.  We build into a non-SCL tag normally.

lzap: Right, was in the wrong impression I must not add it.

LZ

On Thu, Jul 04, 2013 at 09:43:54AM -0700, Dominic Cleal wrote:

> > @@ -31,6 +31,7 @@ BuildRoot:  %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
> >  BuildArch:  noarch
> > 
> >  Requires:   %{?scl_prefix}puppet >= 0.24.4
> > +Requires:   rubygem-foreman_api
> 
> Should have the scl_prefix, in case it is built for SCL.  We build into a non-SCL tag normally.
> 
> ---
> 
> https://github.com/theforeman/foreman-installer/pull/55/files#r5034043

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

foreman_proxy
```diff
@@ -1 +1 @@
-Subproject commit 5a7eedca05d5969c88fb5941551f55af44fcb00b
+Subproject commit 145e7f8b754a17e616e9f0fc7864575b473eee06
```
domcleal: You don't want to do this yet, this commit isn't in our repo.

domcleal:
@lzap could you just add a %changelog entry and bump the release please?

lzap:
D0N3!

domcleal:
Thanks @lzap, merged as e236b54.

PR #59, opened 2013-08-30
Kafoized version of foreman-installer

ares:
Foreman installer is now based on kafo gem
Submodules moved to modules/
Updated spec for new version
Update puppet, foreman, foreman-proxy modules
Don't include helper scripts into RPM


foreman-installer.spec
```diff
@@ -16,8 +16,8 @@
 
 Name:       foreman-installer
 Epoch:      1
-Version:    1.2.9999
-Release:    2%{?dotalphatag}%{?dist}
+Version:    1.3
```
domcleal: This shouldn't be changed, it's version 1.2.9999 at the moment as we haven't branched to 1.3.

config/kafo.yaml
```diff
@@ -0,0 +1,23 @@
...
+# :log_dir: '/var/log/kafo'
+# :log_level: :info
+
+## Kafo tuning, customization of core functionality
+# :no_prefix: false
+:mapping:
+  :puppetmaster:
+    :dir_name: puppet
+    :manifest_name: server
+:order:
```
domcleal: What's this needed for?  (same question for https://github.com/theforeman/kafo/commit/4ccc52657ca2461e22994b8a1c10814f2f615f9a I guess)

foreman-installer.spec
```diff
@@ -61,7 +61,6 @@ mkdir -p %{buildroot}/%{_datadir}/%{name}
 cp -dpR * %{buildroot}/%{_datadir}/%{name}
 %if %{?skip_generator:0}%{!?skip_generator:1}
   mkdir -p %{buildroot}/%{_sbindir}
-  ln -sf %{_datadir}/%{name}/generate_answers.rb %{buildroot}/%{_sbindir}/foreman-generate-answers
```
domcleal, line 64: Shouldn't foreman-install get installed to sbin?

modules/passenger
```diff
@@ -0,0 +1 @@
+Subproject commit 86d5c27141a8bb9a7c2cd871f4eedbabb2ac4e69
```
domcleal: Old commit SHA?

modules/puppet
```diff
@@ -0,0 +1 @@
+Subproject commit a610829945442947c765259593729c7e4083c78f
```
domcleal: Old commit

config/kafo.yaml
```diff
@@ -0,0 +1,23 @@
...
+# note current configuration is written tokafo.yaml every time kafo is run
+
+## Useful for development, e.g. when you want to move log files to local directory
+# :log_dir: '/var/log/kafo'
+# :log_level: :info
+
+## Kafo tuning, customization of core functionality
+# :no_prefix: false
+:mapping:
+  :puppetmaster:
```
domcleal: Should this be the case?  Direct instantiation of puppet::server has been removed in puppet-puppet.

config/kafo.yaml
```diff
@@ -0,0 +1,23 @@
+# kafo main configuration file example
+# you can rename it to kafo.yaml so it overwrite default kafo settings
+# note current configuration is written tokafo.yaml every time kafo is run
+
+## Useful for development, e.g. when you want to move log files to local directory
+# :log_dir: '/var/log/kafo'
```
domcleal: Perhaps set this to /var/log/foreman-install, or /var/log/foreman?

domcleal:
Haven't tested the actual process or kafo, just general comments on this PR.

config/kafo.yaml
GregSutcliffe: The class_name function in kafo can go too, for the same reason ;)

config/kafo.yaml
GregSutcliffe: :+1: for `/var/log/foreman-install`

config/kafo.yaml.example
```diff
@@ -0,0 +1,16 @@
+# kafo main configuration file example
```
domcleal: I guess remove this example file, we just need the actual one

config/kafo.yaml
```diff
@@ -0,0 +1,23 @@
+# kafo main configuration file example
```
domcleal: Should probably change this to say it's the foreman installer

config/kafo.yaml
ares: I encountered various errors when puppet modules were executed in random order. It's probably a problem of puppet modules themself but before they got fixed Kafo provides this feature to specify in which order should the modules be executed.

This is core functionality in kafo that does not have to be necessarily used by foreman (as soon as puppet modules are fixed this can be removed from config).

foreman-installer.spec
ares, line 64: Yes, this is one thing to resolve. Currently you have to actually cd to foreman installer dir to run it. Puppet modules wouldn't be located correctly right now. Definitely something to improve soon.

config/kafo.yaml
domcleal: By executed do you mean parsed, or actually executed?  It appears to just affect the parse order rather than execution.  I don't mind leaving this in, but it'd be interesting to get to the bottom of it.

I know a parse order issue used to exist between puppet and puppet::server, but since the latter no longer directly exists (and kafo shouldn't instantiate it), that shouldn't be the issue.

ekohl: If it's actual parse order, it might help with things like `include` and `class {'...':}` because the latter should be done before the former. An example is puppetdb vs our foreman module and the order matters because of include postgresql being in one of the two (and I forgot which one).

modules/passenger
ares: Old? I actually updated it, but I'll double check we have the latest (same with others).

config/kafo.yaml
ares: This is needed for parsing the parameters from parametrized class. By default kafo must know that puppetmaster specified in answer file has params not in puppetmaster/params.pp but in puppet/server.pp (aka $dir_name/$manifest_name.pp)

config/kafo.yaml
ares: correct, I'll fix that

config/kafo.yaml.example
ares: Actually, this is just tempate which should stay there so people can see possible options and can be regenerated everytime you run kafofy helper. You can run kafofy when you upgrade kafo so you have latest config template.

config/kafo.yaml
ares: sure thing, it was just a copy of default to proof that it works, I'll change that, thank you

config/kafo.yaml
GregSutcliffe: the "puppetmaster" class hack no longer exists. to install a puppetmaster, it's now `class {'puppet': server => true }` so this can go.

That said, from the point of view of kafo as a general library for building installers from puppet modules, it may be useful to others.

foreman-installer.spec
```diff
@@ -72,10 +71,11 @@ cp -dpR * %{buildroot}/%{_datadir}/%{name}
...
+%exclude %{_datadir}/%{name}/release
+%exclude %{_datadir}/%{name}/update_submodules
+%exclude %{_datadir}/%{name}/foreman-installer.spec
 %{_datadir}/%{name}
-%if %{?skip_generator:0}%{!?skip_generator:1}
-  %{_sbindir}/foreman-generate-answers
-%endif
 
 %changelog
 * Mon Jul 22 2013 Lukas Zapletal <lzap+rpm[@]redhat.com> - 1.2.9999-2
```
domcleal, line 96: Could you add a -3 changelog entry please?

foreman-installer.spec
domcleal, line 64: I'd quite like to resolve this before release, is that possible?

config/kafo.yaml
```diff
@@ -0,0 +1,19 @@
+# foreman-installer main configuration file
+# note current configuration is written to kafo.yaml every time kafo is run
+
+## Useful for development, e.g. when you want to move log files to local directory
+:log_dir: '/var/log/foreman-installer'
```
domcleal: You might need to add this to the RPM.

config/kafo.yaml
domcleal: +1

config/kafo.yaml
ares: It's created after first kafo (foreman-installer) run and also user can change it so I don't think it should be mentioned in specfile. Unless we want to precreate it so people may find out from which package this comes from.

foreman-installer.spec
ares, line 64: addressed in latest version, you can now run installer from anywhere you want

domcleal:
I still see out of date SHAs on the puppet and foreman submodules, could you recheck those please?

The README could do with a refresh too, as it's referring to the old answers file and foreman_installer.  Maybe we need a Gemfile specifying kafo for source checkouts?

Lastly, could you submit scratch builds for EL6/F18/F19 to check the new spec works?  Jenkins is an easy way to do this: http://projects.theforeman.org/projects/foreman/wiki/RPM_Packaging#Use-Jenkins  (ping me and I'll set a password for you)

ares:
@domcleal commits are up to date now, I modified README, please review. Working on scratch builds, will post links to koji in here if that's ok.

domcleal:
:+1: once scratch builds are up, thanks.

ares:
EL6 http://koji.katello.org/koji/taskinfo?taskID=51627
F18 http://koji.katello.org/koji/taskinfo?taskID=51629
F19 http://koji.katello.org/koji/taskinfo?taskID=51631

GregSutcliffe:
I'm still hitting some errors, I've raised one bug and I'll discuss the others with you tomorrow @ares

domcleal:
Merged, thanks @ares!  This is going to be great.

PR #61, opened 2013-09-10
Kafo > 0.0.12 allows to specify custom log filename

ares:
So let's discuss log filename.

The project is called `foreman-installer`, the script is named according to action `foreman-install`, later we may add other scripts like `foreman-check` or whatever 

I prefer having log files in `/var/log/$project_name/$script_name` in this case `/var/log/foreman-installer/foreman-install.log`

Also note that this won't have any effect until Kafo 0.0.13 is released.


GregSutcliffe:
:+1:

ares:
@domcleal @lzap you wanted to discuss this?

domcleal:
Fine by me.  I kind of prefer foreman-installer to foreman-install for the name of the script + log, but it's fine as-is.

lzap:
Hmmmm...

domcleal:
@lzap insightful!

ekohl:
Why the rename of the script to foreman-install anyway? Isn't it currently foreman-installer? I feel foreman-installer is a better name since foreman-install can (AFAIK) just write the config and not install. That's something I expect of an installer, but not an install command. This may just be bike shedding and since I don't use the installer, feel free to ignore it :)

lzap:
My only opinion about that is it should be one command, one package name, one repo name. History shows we mis-type and cannot remember this. At least I am life-sentenced to make these mistakes. I don't care if this is foreman-install, installer or whatever.

ares:
I renamed script to foreman-installer and changed the log filename accordingly. Please update your bookmarks if accepted :-)

ekohl:
:+1: on changing to foreman-installer.

domcleal:
:+1:

PR #62, opened 2013-09-10
Add own hostname system checker

ares:
New KAFO is going to drop this hardcoded check. This is extracted and improved from Kafo. We require rubygems but facter can be loaded from libs as well. I assume rubygems are installed on every ruby installation.

### DIFF
--- a/checks/hostname.rb
+++ b/checks/hostname.rb
@@ -1,0 +1,39 @@
+#!/usr/bin/env ruby
+
+begin
+  require 'rubygems'
+rescue LoadError
+end
+require 'facter'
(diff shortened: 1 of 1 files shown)


ares:
@GregSutcliffe would you please review? since you're the author of original check. Also I'd like to know your opinion on require 'rubygems' there.

GregSutcliffe:
I assume we'll be calling this from foreman-install? I don't see anything wrong with the script, I just don't see anything which calls it :)

make it `require 'rubygems' rescue nil` to make it debian-proof :)

PR #65, opened 2013-09-13
Change digest type to MD5 to stop unpack errors

domcleal:
I don't know if this solves it, but I have it on @lzap's authority that this is what tito does :)


lzap:
Yes this is what tito does.

domcleal:
Actual fix is in #76 I believe.

lzap:
Ok we just need to remember that if we will be building EL5 packages (maybe client API or something), we must enable this. But yeah, in general - not recommended.

PR #67, opened 2013-09-26
Fixes #3131 - Output information at the end of the foreman-installer

GregSutcliffe:
Produces output like:

```
# foreman-installer
  Success!
    You can now login at https://test.domain.org
    A proxy is running at https://test.domain.org:8443
    A puppetmaster has been set up on port 8140
  There is a full log /var/log/foreman-installer/foreman-installer.log
```

or on error:

```
# foreman-installer
  Something went wrong! Check the log below for ERROR level lines
  There is a full log /var/log/foreman-installer/foreman-installer.log
```

### DIFF
--- a/bin/foreman-installer
+++ b/bin/foreman-installer
@@ -1,6 +1,56 @@
 #!/usr/bin/env ruby
 require 'rubygems'
+require 'highline/import'
+require 'yaml'
 require 'kafo'
 
-CONFIG_FILE = "config/foreman-installer.yaml"
-KafoConfigure.run
+# Run the install
+CONFIG_FILE = "/usr/share/foreman-installer/config/foreman-installer.yaml"
+@result = KafoConfigure.run
+exit 0 if @result.nil? # --help invocation
+
+# Parse the results
+HighLine.color_scheme = HighLine::ColorScheme.new do |cs|
+  cs[:info] = [:bold, :cyan, :on_black]
+  cs[:bad]  = [:bold, :red, :on_black]
+  cs[:good] = [:bold, :green, :on_black]
+end
+
+def module_enabled? name
+  mod = @result.module(name)
+  return false if mod.nil?
+  mod.enabled?
+end
+
+def get_param mod, name
+  @result.param(mod,name).value
+end
+
+# Puppet status codes say 0 for unchanged, 2 for changed succesfully
+if [0,2].include? @result.exit_code
+  say "  <%= color('Success!', :good) %>"
+
+  # Foreman UI?
+  if module_enabled? 'foreman'
+    say "  * <%= color('Foreman', :info) %> is running at <%= color('#{get_param('foreman','foreman_url')}', :info) %>"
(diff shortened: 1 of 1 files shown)


ekohl:
Looks good, but maybe you can check if authentication is enabled and mention default username + password?

bin/foreman-installer
```diff
@@ -1,6 +1,38 @@
...
+require 'yaml'
 require 'kafo'
 
-CONFIG_FILE = "config/foreman-installer.yaml"
-KafoConfigure.run
+CONFIG_FILE = "/usr/share/foreman-installer/config/foreman-installer.yaml"
+result = KafoConfigure.run
+exit 0 if result.nil? # --help invocation
+
+answers = YAML.load_file(result.config.app[:answer_file])
```
ares: Why don't you use parameter values that are already parsed? Something like

``` ruby
if result.config['foreman_proxy']
```

reloading and reparsing answer file seems wrong, also becase params can have some extra parsing logic

bin/foreman-installer
```diff
@@ -1,6 +1,38 @@
...
+
+# Puppet status codes say 0 for unchanged, 2 for changed succesfully
+unless [0,2].include? result.exit_code
+  puts "  Success!"
+
+  # Foreman UI?
+  if ( answers.include? 'foreman' and answers['foreman'] != false)
```
ares: Could you please change `and` for `&&`? `and` and `or` are usually used for flow control, `&&` and `||` are used in boolean expressions

GregSutcliffe:
@ekohl done

@ares Updated to use https://github.com/theforeman/kafo/issues/15

ares:
Kafo 0.0.16 released. This versions contains patch from https://github.com/theforeman/kafo/pull/15

GregSutcliffe:
Tested with kafo 0.0.16, works as expected (on Wheezy)

bin/foreman-installer
```diff
@@ -1,6 +1,56 @@
...
+  if module_enabled? 'foreman_proxy'
+    say "  * <%= color('Foreman Proxy', :info) %> is running at <%= color('#{get_param('foreman_proxy','registered_proxy_url')}', :info) %>"
+  end
+
+  # Puppetmaster?
+  if ( module_enabled?('puppet') && ( get_param('puppet','server') != false ) )
+    say "  * <%= color('Puppetmaster', :info) %> is running at <%= color('port #{get_param('puppet','server_port')}', :info) %>"
+  end
+else
+  say "  <%= color('Something went wrong!', :bad) %> Check the log for ERROR-level output"
```
lzap, line 49: Minor suggestion: Can we add which log file to check? We will be doing this million times, I'd love just to copy and paste it into grep command... :-)

Nice thanks!

GregSutcliffe, line 49: It does, see line 54

GregSutcliffe:
bump, are we including this in rc3?

domcleal:
Looks OK to me.  Can you add highline to the Gemfile as we now explicitly require it?

lzap:
Merging as there is no Gemfile in the repo :-) Thanks guys

PR #69, opened 2013-10-02
Revert config file location, should be relative

domcleal:


GregSutcliffe:
Thanks @domcleal, merged!

PR #70, opened 2013-10-07
fixes #3209 - exit code only zero by default

lzap:
We also need to merge this

https://github.com/theforeman/kafo/pull/27

in kafo to make this work.

I hope it is now clear what I mean.

With this patch, you can run installer and by default you always get zero on
success. Although if you want to parse detailed puppet exit codes (0, 2 on
success), you can still provide extra option for that.

### DIFF
--- a/bin/foreman-installer
+++ b/bin/foreman-installer
@@ -45,12 +45,14 @@ if [0,2].include? @result.exit_code
   if ( module_enabled?('puppet') && ( get_param('puppet','server') != false ) )
     say "  * <%= color('Puppetmaster', :info) %> is running at <%= color('port #{get_param('puppet','server_port')}', :info) %>"
   end
+  exit_code = 0
 else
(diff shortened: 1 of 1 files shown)


domcleal:
Surely get rid of the `KafoConfigure.config.app[:detailed_exitcodes]` reference and you don't need to change kafo's behaviour?  It'll also then give an accurate exit code when errors occur etc.

lzap:
Okay this is what I have now. Simple solution.

GregSutcliffe:
I think this is ok now, @domcleal?

lzap:
Sorry for the confusion, I was doing my best to give users possibility to rely on default puppet exit codes. As I have learned today - it is useless, so my effort was pretty useless... :-P

GregSutcliffe:
@lzap heh, it happens :)

domcleal:
@GregSutcliffe yep, looks good

lzap:
Merged as a287dc9 thanks all.

PR #76, opened 2013-11-08
Fix concurrent execution of the release script during Jenkins builds

domcleal:
Should fix the invalid SRPMS and tarballs being uploaded to koji, which were
failing with CRC errors as Jenkins runs 2/3 of these in parallel for
different dists.

e.g. http://koji.katello.org/koji/getfile?taskID=65594&name=build.log

### DIFF
--- a/release
+++ b/release
@@ -14,13 +14,15 @@ DIST=$2
 
 git describe $VERSION >/dev/null || echo "WARNING! There is no such tag: $VERSION!"
 
+RPMBUILD=$(pwd)/pkg/rpmbuild
 TEMPDIR=$(mktemp -d)
 trap "rm -rf $TEMPDIR" EXIT
-mkdir $TEMPDIR/$PROJECT-$VERSION
+mkdir -p $TEMPDIR/$PROJECT-$VERSION $RPMBUILD/SOURCES
 cp -Rad . $TEMPDIR/$PROJECT-$VERSION
+
 pushd $TEMPDIR
 rm -rf `find -name ".git*"`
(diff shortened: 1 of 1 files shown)


release
```diff
@@ -14,13 +14,15 @@ DIST=$2
 
 git describe $VERSION >/dev/null || echo "WARNING! There is no such tag: $VERSION!"
 
+RPMBUILD=$(pwd)/pkg/rpmbuild
```
lzap, line 17: I wonder if we want to use mktemp here just to have some random things here to be absolutely sure this will not happen again. Something like:

```
RPMBUILD=$(mktemp -d)
trap "rm -rf $RPMBUILD" EXIT
```

Stunning, this was driving me crazy.

lzap:
But if this is tested, then you have +1 from me...

release
domcleal, line 17: I did start by adding the rpmbuild dir under $TEMPDIR but then realised if the rpmbuild dir is deleted when the script exits, we delete the SRPM and can't submit it to koji :)

In Jenkins it creates a separate workspace for every job running, so we can safely run this in parallel provided the files are created within the checkout.

lzap:
Oh maybe to add .gitignore too.

domcleal:
pkg/ is already in .gitignore here as we use it for the tar.gzs.

release
lzap, line 17: Oh in that case you could use:

```
RPMBUILD=$(mktemp --tmpdir $(pwd)/pkg-XXXXXXX/rpmbuild)
```

But I assume Jenkins already creates unique PWD right?

domcleal, line 17: Good idea, though as you say, this PWD is unique.

lzap:
Merged with 3897a1a..aee6be6 ty

PR #80, opened 2013-12-12
Bumping modules to fix proxy registration

lzap:
Hopefully not bringing any issues. Do not merge, I am testing that. Feel free
to join me with testing on other platforms than RHEL 6.4. :-D


modules/postgresql
```diff
@@ -1 +1 @@
-Subproject commit 80d44718e93ea1417f1c84cd730c41a2187bd39c
+Subproject commit 8ed663fff55549844e678ba09a03607ce408bd18
```
domcleal: Don't update this yet, we're using the older version still compatible with 2.6

lzap:
Fixed, but errors reported:

```
[ERROR 2013-12-12 11:51:36 verbose]  /File[/var/lib/puppet/concat]/ensure: change from absent to directory failed: Could not set 'directory on ensure: Could not find user postgres at /usr/share/foreman-installer/modules/concat/manifests/setup.pp:49
```

Holding off.

lzap:
Ping - anyone?

modules/concat
```diff
@@ -1 +1 @@
-Subproject commit 57ce1c70c822ae96b2fe17512913981485160539
+Subproject commit 63b5aba9fd8aee29ae2461d1819abc92095afc50
```
domcleal: This updates us from ripienaar-concat to puppetlabs-concat, while we still have postgresql pinned at an older version which points at the former.  I wouldn't bump it yet.

lzap:
Thanks Dom, updated, this is working on my RHEL6. :+1:

domcleal:
Tested + merged as 39bac59, thanks @lzap!

PR #83, opened 2013-12-30
Add a Gemfile

ekohl:


lzap:
Will the RPM build? +1

Gemfile
```diff
@@ -0,0 +1,5 @@
+source 'https://rubygems.org'
+
+gem 'mime-types', '<2.0'  # Needs ruby 1.9.2+
```
ekohl, line 3: This is an indirect dependency btw. Not sure if you can check for the ruby version in a Gemfile and only then set it.

ekohl:
I did find out that running the installer under bundler can give bad results. Things like puppet fail to run properly so I'm not sure if this is still needed.

domcleal:
The RPM builds: http://ci.theforeman.org/job/packaging_build_matrix_rpm/854/

Gemfile
domcleal, line 3: `gem 'mime-types', :platforms => :ruby_18`

ekohl:
I don't think this is needed anymore and needs a rebase anyway.

PR #88, opened 2014-01-10
Bump module versions to develop

lzap:
For @domcleal


modules/stdlib
```diff
@@ -1 +1 @@
-Subproject commit 224b8f9a191f635b03ee900a9bf87bfdb0f1a6ed
+Subproject commit 01c9c3c97998c220c41b386e64468a1efff37127
```
domcleal: Please drop this, it triggers a kafo bug: http://projects.theforeman.org/issues/3990

modules/concat
```diff
@@ -1 +1 @@
-Subproject commit 57ce1c70c822ae96b2fe17512913981485160539
+Subproject commit 650a73244df54dbe24c58ace438e2726f0c7c7b1
```
domcleal: Could you please revert to 04356974f72b90a1d0f57346a00e95a717924e43 (the v1.0.0 release)?

I hit https://github.com/puppetlabs/puppetlabs-postgresql/pull/343 when testing.

domcleal:
Verified \o/

el6: http://ci.theforeman.org/job/systest_foreman/29/ PASS
f19: http://ci.theforeman.org/job/systest_foreman/33/tapResults FAIL
precise: http://ci.theforeman.org/job/systest_foreman/39/tapResults PASS, with http://projects.theforeman.org/issues/3977
wheezy: http://ci.theforeman.org/job/systest_foreman/37/tapResults PASS, with http://bugs.debian.org/cgi-bin/bugreport.cgi?bug=734444

F19 seems to be in a bit of a mess.  There's an issue with rgen missing from the puppet RPM, which breaks the proxy (since we merged future parser support, https://bugzilla.redhat.com/show_bug.cgi?id=1051613).

My system also fails to start PostgreSQL, as it seems to already be running.  More debugging required, this could have been introduced in the new module.  I think it's getting a bad "status" back or something.

I also have a Puppet SSL issue, unsure if that's just me or not.  Will re-test.

domcleal:
Ack, I ran two instances of the installer.  Let me re-test F19.

domcleal:
F19 is PASS, it was only the rgen issue above, worked with PL puppet.

domcleal:
Tested squeeze too, PASS.  Thanks @lzap!

lzap:
> My system also fails to start PostgreSQL, as it seems to already be running.  More debugging required, this could have been introduced in the new module.  I think it's getting a bad "status" back or something.
> 
> There is a known bug in PSQL init script in RHEL6:

https://bugzilla.redhat.com/show_bug.cgi?id=800534

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

PR #89, opened 2014-01-10
Fixes #3977 - Don't re-open a Highline colourscheme

GregSutcliffe:
Re-opening a colourr scheme crashes on Squeeze/Precise, so just define a hash and use it twice.


GregSutcliffe:
Tested on squeeze

domcleal:
Looks good, merged as cc00eed, thanks @GregSutcliffe.

PR #91, opened 2014-01-16
Bump foreman module to include symlink fix

domcleal:
systest: http://ci.theforeman.org/job/systest_foreman_matrix/3/ (though F19 is currently disabled due to multiple unrelated issues)


skottler:
Merged as fbfa397480f1524a2eae5577463ee0c3efd81631. thanks!

PR #92, opened 2014-01-16
Fully drop 2.6 support

ekohl:
Now that we require puppet 2.7+ we no longer need the create_resources module and can rely on the built in. The other module bumps update travis to drop 2.6 support.


domcleal:
Thanks @ekohl!

PR #93, opened 2014-01-17
Add new script to tag and build modules consistently

domcleal:
I'm adding this so that all of us maintaining the installer can quickly and consistently release modules to the forge (I'll send out credentials).  Once we can do this easily for the modules we have, it'll be easier to switch the installer to use librarian-puppet.


lzap:
This was confusing for the author himself :-) Thanks, merging.

lzap:
Merged as a9fd624

PR #94, opened 2014-01-20
fixes #4030 - build via librarian-puppet

domcleal:
Putting this up for comments.

This supports running librarian-puppet in either the build (via the Rakefile) which we'll probably use for Debian packages once they're using the Rakefile, or in the "release" script which is used to build SRPMs.

Both work reasonably well I think, though the release script suffers from a bug when running concurrently as librarian-puppet calls Puppet's module tool, which gets confused when the vardir (~/.puppet/var) is used concurrently on Jenkins.  The Jenkins job is running under RVM like our others, so it installs gems in 'release'.

I got one successful el6 build out of it: http://koji.katello.org/koji/taskinfo?taskID=75163 (http://ci.theforeman.org/job/packaging_build_matrix_rpm/1092/).


modules/xinetd
```diff
@@ -1 +0,0 @@
-Subproject commit 3a640d8418090431ac469bfdb02a7d06a35c597f
```
domcleal, line 1: We'll have a regression here as there's no recent release of pl-xinetd containing Amazon support.

ekohl, line 1: Time to bug @apenney I suppose. In general I think we should look at a vagrant image of Amazon Linux so we can claim actual support instead of just best effort based on bug reports.

skottler, line 1: AL isn't made available outside of AWS - it's one of their added value things. Maybe we can force the facts to look like AL using `FACTER_*` environment variables in the system tests?

domcleal, line 1: I've not come across enough users to think it warrants more than best effort support at the moment.

skottler, line 1: Agreed that it's not terribly important. It's also "just" conditionals since it's very similar to EL in everything but name.

GregSutcliffe, line 1: :+1: for best effort, I don't regard it as a _core_ OS

GregSutcliffe:
@domcleal apart from the rpmbuild line at the end of the release script, I assume this would work pretty well for the debs too. The current installer job on jenkins just checks out foreman-installer, deletes **/.git and then tars it, so the fact that you're putting out a tarball seems very suitable...

I'd test it but I'm lacking infrastructure this week :)

modules/apache
```diff
@@ -1 +0,0 @@
-Subproject commit c9fef6c882cb57c059920cf347b92e1ed77ac2e6
```
domcleal, line 1: Regression here too, now that I think about it, as we rely on @stbenjam's fork with Apache 2.4 fixes.  The pl-apache team have been talking about getting the branch merged and released soon, so it won't be long.

modules/xinetd
domcleal, line 1: https://github.com/puppetlabs/puppetlabs-xinetd/issues/32

domcleal:
@GregSutcliffe perhaps as something separate to this PR, we should make the "release" scripts generic and add a "-r" option to invoke the rpmbuild step, if you're able to work with this for now.

modules/apache
domcleal, line 1: Hooray, it's merged.  I've added their git repo to the Puppetfile temporarily, until there's a fully fledged release available, then we'll drop that line.

domcleal:
Rebased, and I'd like to merge sooner rather than later :)

I've confirmed the package builds on Jenkins ([green](http://ci.theforeman.org/job/packaging_build_matrix_rpm/1288/)), and have disabled concurrent builds until I can fix the underlying Puppet bug.

I've checked the RPM package contents and ran BATS against it on EL6 successfully.  F19 nearly works, once https://github.com/puppetlabs/puppetlabs-apache/pull/612 is merged.

Since the Debian package now uses the Rakefile, we can rely on the librarian calls inside this to populate modules/, provided we can get librarian-puppet into the build environment... @GregSutcliffe thoughts?

GregSutcliffe:
@domcleal in general, works for me. Some scratch builds on nightly would be excellent - I'll try to get to it soon if you don't.

domcleal:
@GregSutcliffe if you could please, I don't understand how to use Jenkins to do that

ekohl:
@domcleal maybe you should update the title of this PR to remove the [DO NOT MERGE] part then.

domcleal:
done :)

ekohl:
:+1:

modules/xinetd
domcleal, line 1: Added puppetlabs-xinetd to the Puppetfile, pulling from git in the meantime.

domcleal:
EL6 and F19 BATS both successful now.

GregSutcliffe:
@domcleal Ok, using something like https://github.com/GregSutcliffe/foreman-packaging/commit/9231caba5651a9e8c7a4d6241245ac066722aeb3 allows the package to build. That's hacky, and we should improve our tarball generation process, but at least we can move forward...

domcleal:
Thanks for the help, merged and have enabled nightly builds of the installer packages for now on Jenkins so the modules are fresh :)

PR #96, opened 2014-01-21
Allow to skip man page generation

lzap:
With `rake NO_MAN_PAGE=1` command it will be possible to build also on squeeze
where asciidoc is somehow screwed.


ekohl:
:+1: because eventually I want to publish release tarballs anyway. We can just ship a generated version there.

domcleal:
Merged, thanks @lzap.

PR #98, opened 2014-01-23
Bump puppet module for F19 fix

domcleal:
Pulls in https://github.com/theforeman/puppet-puppet/pull/129.

F19 build with PL repo (non-PL fails now due to RHBZ#1051613, rgen): http://ci.theforeman.org/job/systest_foreman/204/ at least gets rid of that error.  A foreman_api 0.1.10 issue remains causing the whole run to fail, please ignore that.


lzap:
Fixes as c5b47c1 thanks @domcleal

PR #100, opened 2014-02-06
moving the foreman installer's bin file into a library we can reuse between both Katello and Foreman

mccun934:
This allows Katello and others to re-use the code in Foreman's kafo
launcher.  Originally I had thought we could move this into kafo
directly but almost _all_ of the code in the old bin/foreman-installer
and now lib/foreman_installer/launcher.rb was specific to Foreman and
not really well suited to Kafo directly.

### DIFF
--- a/foreman_installer.gemspec
+++ b/foreman_installer.gemspec
@@ -1,0 +1,20 @@
+# -*- encoding: utf-8 -*-
+require File.expand_path('../lib/foreman_installer/version', __FILE__)
+
+Gem::Specification.new do |gem|
+  gem.authors       = ["Foreman"]
+  gem.email         = ["foreman-dev@googlegroups.com"]
+  gem.license       = "MIT"
+  gem.description   = %q{Helps you call The Foreman's installer via ruby}
+  gem.summary       = %q{Helps you call The Foreman's installer via ruby}
+  gem.homepage      = "https://github.com/theforeman/foreman-installer"
+
+  gem.files = %w(foreman_installer.gemspec LICENSE README.md) + Dir.glob("{lib}/**/*")
+
+  gem.name          = "foreman_installer"
+  gem.require_paths = ["lib"]
+  gem.version       = ForemanInstaller::VERSION
+
+  gem.add_dependency 'highline', '~> 1.6'
+  gem.add_dependency 'kafo', '~> 0.3'
+end
--- a/bin/foreman-installer
+++ b/bin/foreman-installer
@@ -1,94 +1,9 @@
 #!/usr/bin/env ruby
 require 'rubygems'
-require 'highline/import'
-require 'yaml'
-require 'kafo'
+require 'foreman_installer'
 
 # where to find answer file
 CONFIG_FILE = "config/foreman-installer.yaml"
 
-# helpers
-def module_enabled?(name)
-  mod = @result.module(name)
-  return false if mod.nil?
-  mod.enabled?
-end
-
-def get_param(mod, name)
-  @result.param(mod, name).value
-end
-
-# functions specific to foreman installer
-Kafo::KafoConfigure.app_option '--reset-foreman-db', :flag,
-                               "Drop foreman database first? You will lose all data! Unfortunately we\n" +
-                               "can't detect a failure at the moment so you should verify the success\n" +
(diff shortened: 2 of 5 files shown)


foreman_installer.gemspec
```diff
@@ -0,0 +1,20 @@
...
+
+Gem::Specification.new do |gem|
+  gem.authors       = ["Foreman"]
+  gem.email         = ["foreman-dev@googlegroups.com"]
+  gem.license       = "MIT"
+  gem.description   = %q{Helps you call The Foreman's installer via ruby}
+  gem.summary       = %q{Helps you call The Foreman's installer via ruby}
+  gem.homepage      = "https://github.com/theforeman/foreman-installer"
+
+  gem.files = %w(foreman_installer.gemspec LICENSE README.md) + Dir.glob("{lib}/**/*")
```
lzap, line 12: Hey Mike. Don't you miss "bin" directory here? Does the gem actually works, or is this the scratch version?

We want to switch the SPEC file and make it a gem-based. I guess I can do that if you haven't prepared the SPEC already.

Thanks

mccun934, line 12: I suppose we could make the entire installer be install-able via gem but I was thinking this would be for _just_ the lib.  

actually, that does make some good sense to just have the entire thing be a gem:

$ gem install foreman-installer
$ foreman-installer -h 

will update!

domcleal:
:-1: to splitting this into two packages, it's unnecessary just for a small script and will give us a lot of packaging and release overhead.  We already try to keep what's here to a minimum.

A single gem is better, but I still question how much is actually relevant to Katello anyway, since different top level modules are used.  The API exposed here doesn't look particularly suitable for extension.

mccun934:
the entire contents of the foreman-installer launcher is relevant.  Since Katello is Foreman + Katello we want to be able to reuse:
- the status output at the end of the installer (urls, admin login info)
- the additional option for resetting the db
- the puppet exit code parsing

we could of course just copy-paste all this code into katello's installer but that is a horrible way to try and re-use this logic that we both need.   If there is some better way for both the katello-installer and foreman-installer to share this logic, I'm happy to look into it.

GregSutcliffe:
What's not in the current script that Katello needs? Given that the contents of `modules/` is whatever puppet modules you like, and the config/answer files are separate, is there anything that prevents the full Foreman+Katello installer from using this script as-is? I agree with re-using code, but it seems simpler to extend the current script nicely, rather than add a whole new package for 7 OSes.

mccun934:
@GregSutcliffe we actually want _everything_ except the location of the configuration file which kafo requires to be a global variable:

CONFIG_FILE = "config/foreman-installer.yaml"

for katello we need:

CONFIG_FILE = "./config/katello_installer.yaml"

with our own separate set of modules since we install way more stuff than vanilla Foreman.

if there is some easy way to inline/require/reuse the code in the foreman-installer/bin in katello's binary:

https://github.com/Katello/katello-installer/blob/master/bin/katello-installer

 without pulling it out to a library/class that I can use, I'm totally happy to do it.  I just couldn't see a way around it with that global var declaration requirement.

GregSutcliffe:
@mccun934 so lets refactor to make that an optional argument to the installer script, with the default staying as is. Would something like:

```
if config_file ARGS.find {|a| a ~= /--config-file/}
  CONFIG_FILE = config_file.split('=').last
else
  CONFIG_FILE = "config/foreman-installer.yaml"
end
```

And call it as `foreman-installer --config-file=config/katello_installer.yaml`. Obviously we can make that nicer, but you get the idea...

mccun934:
ah yes, way better.  katello's installer can just call that

thanks, this simplifies the change required to foreman-installer a bunch

GregSutcliffe:
@mccun934 the only thing to watch out for is to strip the option from ARGV before we pass it to Kafo I guess.

ekohl:
You could even use the file name as a basis for the config file. Then you just need to rename (or symlink) to katello-installer.

GregSutcliffe:
yeah, basing the default on basename[0] seems sensible.

mccun934:
@GregSutcliffe @ekohl thanks for the ideas, will update this (or a new pr) with this new approach

PR #101, opened 2014-02-07
base the answer file location on the basename of the script 

mccun934:
This change allows katello-installer and others to use its own configuration file
but re-use the logic contained within foreman-installer.

Originally proposed this here:

https://github.com/theforeman/foreman-installer/pull/100

but re-did this change here with an entirely different approach.  We also no longer need the sed script to munge the location during the build which becomes hard to maintain when the CONFIG_FILE variable is now based on some logic

### DIFF
--- a/bin/foreman-installer
+++ b/bin/foreman-installer
@@ -4,8 +4,17 @@ require 'highline/import'
 require 'yaml'
 require 'kafo'
 
+# base the answer file name on the file calling the
+# script.  This allows other programs to use their own
+# files if named based on their executable
+config_filename = File.basename($0) + ".yaml"
+
 # where to find answer file
-CONFIG_FILE = "config/foreman-installer.yaml"
+if File.exist?('config/' + config_filename)
+  CONFIG_FILE = 'config/' + config_filename   
+else
+  CONFIG_FILE = '/etc/foreman/' + config_filename
+end
 
 # helpers
 def module_enabled?(name)
--- a/Rakefile
+++ b/Rakefile
@@ -33,7 +33,7 @@ end
 
 file "#{BUILDDIR}/foreman-installer" => 'bin/foreman-installer' do |t|
   cp t.prerequisites[0], t.name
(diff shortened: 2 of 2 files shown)


bin/foreman-installer
```diff
@@ -4,8 +4,17 @@ require 'highline/import'
...
+# script.  This allows other programs to use their own
+# files if named based on their executable
+config_filename = File.basename($0) + ".yaml"
+
 # where to find answer file
-CONFIG_FILE = "config/foreman-installer.yaml"
+if File.exist?('config/' + config_filename)
+  CONFIG_FILE = 'config/' + config_filename   
+else
+  CONFIG_FILE = '/etc/foreman/' + config_filename
```
lzap, line 16: Mike can you keep the sed in the Rakefile and change the /etc/foreman path according to the SYSCONFDIR setting? We now allow users to use Rakefile in tarball distribution, this is hardcoded now. It would be something like (not tested):

```
sh 'sed -i "s#\(.*CONFIG_FILE\) = '/etc.*#\1 = \"%s\" + config_filename#" %s' % ["#{SYSCONFDIR}/foreman/", t.name]
```

Does not look nice, if you coud introduce something like BASE_CONFIG_DIR variable and sed that, would be much nicer.

mccun934, line 16: updated to something similar to what you mentioned so you now get after running Rake outside the RPM's build:

```
if File.exist?('config/' + config_filename)
  CONFIG_FILE = 'config/' + config_filename   
else
  CONFIG_FILE = '/usr/local/etc/foreman/' + config_filename
end
```

lzap, line 16: Yeah thats fine, in RPM we set these dirs correctly. :+1:

GregSutcliffe:
Tested on Wheezy, appears to work fine. Copying both the binary and the config file to foo-installer works as expected.

:+1:

GregSutcliffe:
@domcleal @ekohl any concerns?

domcleal:
:+1:

GregSutcliffe:
merged as 310dffb with the commit message tidied up a little ;)

Thanks @mccun934!

PR #102, opened 2014-02-11
refs #4030 - use latest dependency modules from git

domcleal:
It just occurred to me that we're not pulling the latest dependency modules from git, just from the Forge.

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -1,11 +1,20 @@
 forge 'http://forge.puppetlabs.com'
 
 # Temporary for Apache 2.4 support (post-0.11.0)
-mod 'apache', :git => 'https://github.com/puppetlabs/puppetlabs-apache'
+mod 'puppetlabs/apache', :git => 'https://github.com/puppetlabs/puppetlabs-apache'
 
 # Temporary for Amazon Linux support (https://github.com/puppetlabs/puppetlabs-xinetd/issues/32)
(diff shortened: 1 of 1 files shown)


Puppetfile
```diff
@@ -6,6 +6,14 @@ mod 'apache', :git => 'https://github.com/puppetlabs/puppetlabs-apache'
 # Temporary for Amazon Linux support (https://github.com/puppetlabs/puppetlabs-xinetd/issues/32)
 mod 'xinetd', :git => 'https://github.com/puppetlabs/puppetlabs-xinetd'
 
+# Dependencies
+mod 'concat', :git => 'https://github.com/theforeman/puppet-concat'
```
ekohl: Won't this give conflicts with puppetlabs-concat?

domcleal: Good catch, thanks!

domcleal:
Updated and tested this time, needs https://github.com/theforeman/puppet-tftp/pull/17 to work.

ekohl:
:+1:

GregSutcliffe:
From testing, we need the `puppetlabs/` prefix adding back in - http://pastie.org/8725745. While we're at it, the mysql module should be explicitly listed.

domcleal:
Thanks for the testing, updated.

ekohl:
:+1:

GregSutcliffe:
Merged, thanks @domcleal!

PR #104, opened 2014-02-14
Fix compatibility with Facter 2 API

domcleal:
Tested on 1.6, 1.7 and master.  This is the usual way to reference fact values within facter's source too.


ares:
Tested, works nicely, maybe just modify commit message so it will close http://projects.theforeman.org/issues/4372

domcleal:
Message updated, thanks.

ekohl:
Merged, thanks

PR #108, opened 2014-03-14
Reuse color support from Kafo

ares:
Installer output now uses color scheme selected by user. Also we don't
have to check color configuration manually. HighLine is already set by
Kafo.


domcleal:
Should we set a minimum kafo version?

ares:
Good catch, fixed and rebased

domcleal:
Merged, thanks!  Please also submit a PR to deb/develop to require kafo 0.4.0.

ares:
done as https://github.com/theforeman/foreman-packaging/pull/196

PR #110, opened 2014-04-18
Depend on latest kafo and curl

ares:
We need it because of foreman_discovery plugin installation

Merge this before or at the same time as https://github.com/theforeman/puppet-foreman/pull/174


domcleal:
:+1: and ditto for Debian in foreman-packaging :)

ares:
Thanks, I opened PR https://github.com/theforeman/foreman-packaging/pull/221

domcleal:
Merged as eeb5b42, thanks @ares.

PR #111, opened 2014-04-23
Add support for discovery params class

ares:
We need this merged after https://github.com/theforeman/puppet-foreman/pull/174 is merged


domcleal:
Merged as 8e217f9.

iNecas:
For further tracking of issues in the installer, it would be nice to have the "every commit with an Remine issue" as well, to make it easy when hunting the issues in katello-installer, that already have been resolved, like this one http://projects.theforeman.org/issues/5576

domcleal:
This is the wrong forum for that discussion.

PR #116, opened 2014-05-13
refs #3272 - print new user/password after installation

domcleal:
Depends on https://github.com/theforeman/puppet-foreman/pull/190.

The change to print the details all the time might be needed if we had a partial failure, but enough that the web UI works - so as not to leave the user with no idea what their credentials are.

### DIFF
--- a/bin/foreman-installer
+++ b/bin/foreman-installer
@@ -11,7 +11,7 @@ config_filename = File.basename($0) + ".yaml"
 
 # where to find answer file
 if File.exist?('config/' + config_filename)
-  CONFIG_FILE = 'config/' + config_filename   
+  CONFIG_FILE = 'config/' + config_filename
 else
   CONFIG_FILE = '/etc/foreman/' + config_filename
 end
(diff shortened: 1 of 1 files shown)


bin/foreman-installer
```diff
@@ -57,28 +57,28 @@ exit 0 if @result.nil? # --help invocation
...
   exit_code = 0
 else
   say "  <%= color('Something went wrong!', :bad) %> Check the log for ERROR-level output"
   exit_code = @result.exit_code
 end
 
+# Foreman UI?
+if module_enabled? 'foreman'
+  say "  * <%= color('Foreman', :info) %> is running at <%= color('#{get_param('foreman','foreman_url')}', :info) %>"
+  say "      Credentials are '<%= color('#{get_param('foreman', 'admin_user')} / #{get_param('foreman', 'admin_password')}', :info) %>'" if get_param('foreman','authentication') == true
```
ekohl: One problem I do see is rerunning the installer to upgrade for example, but these will still only be the default parameters and the user could have changed the password.

domcleal: That's possible - the rake task is designed not to overwrite or change the admin user if one already exists, so it's actually an initial password.  I've added the word "initial", which should clarify.

ekohl:
:+1: pending other patch.

domcleal:
screenshot from 2014-05-14 14 21 44

lzap:
Nice, our installer is pure awesomness.

dLobatog:
Depends on https://github.com/theforeman/puppet-foreman/pull/192 at the moment.

GregSutcliffe:
Tested, works great, :+1: will merge when the core PR goes in.

ohadlevy:
PR is merged, merge at will

PR #118, opened 2014-05-28
refs #5793 - add pkg:generate_source rake task to create tar.bz2

domcleal:


lzap:
Tested, merged as a040723 thanks.

PR #119, opened 2014-05-29
refs #5987 - remove unused packaging files

domcleal:
https://github.com/theforeman/foreman-packaging/tree/rpm/develop/foreman-installer


lzap:
Merged as a4c7784, thanks.

PR #120, opened 2014-06-09
Add foreman_ipa integration module

ares:
as a part of a review, the puppet module itself should be reviewed, it lives at https://github.com/theforeman/puppet-foreman_ipa

I tested the installer, if you have any questions regarding setup, feel free to ask me

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -19,3 +19,4 @@ mod 'theforeman/tftp',          :git => 'https://github.com/theforeman/puppet-tf
 mod 'theforeman/foreman',       :git => 'https://github.com/theforeman/puppet-foreman'
 mod 'theforeman/foreman_proxy', :git => 'https://github.com/theforeman/puppet-foreman_proxy'
 mod 'theforeman/puppet',        :git => 'https://github.com/theforeman/puppet-puppet'
+mod 'theforeman/foreman_ipa',   :git => 'https://github.com/theforeman/puppet-foreman_ipa'
--- a/config/answers.yaml
+++ b/config/answers.yaml
@@ -32,3 +32,4 @@ foreman::compute::openstack: false
 foreman::compute::ovirt: false
 foreman::compute::rackspace: false
 foreman::compute::vmware: false
+foreman_ipa: false


domcleal:
foreman_ipa has a lot of issues, how should it be reviewed, since there's no PR?

ares:
create a new PR with fixes? :-) I can either create blank branch and create a PR to track reviews or I could move it to my fork, clean theforeman repo and create a PR, it's likely that no one have used it yet. Which one do you prefer?

domcleal:
I think I prefer the latter, thanks.

ares:
I created a new PR at https://github.com/theforeman/puppet-foreman_ipa/pull/1

domcleal:
Please note that @purpleidea is also working on a similar module - we should probably work on only one or the other.  https://github.com/purpleidea/puppet-ipa

purpleidea:
Hey... I'm a bit confused... Maybe I've misunderstood something. Maybe someone can provide some background...
- You need a puppet module for each separate component you want to "puppet".
  Eg: one puppet-ipa module, one puppet-gluster module, one puppet-pulp module, and perhaps even a puppet-foreman module. Why there would be a "puppet-foreman-ipa" module isn't quite clear to me.
- I've got a puppet-ipa module that I'm actively working on (and I've got a large feature branch with new magic landing soon) which deploys FreeIPA servers... It obviously needs hosts provisioned first (foreman) or similar...

purpleidea:
@domcleal the module you referenced is quite a good base, and fully works. Missing features coming soon include true multi-master replication...

purpleidea:
09:52 < Dominic> foreman_ipa doesn't seem to configure IPA at all, it's only 
                 foreman configuration

AFAICT, if that is correct, then we're talking about apples and flamethrowers

ekohl:
@purpleidea I agree it makes more sense for puppet-foreman_api to be merged into puppet-foreman since. Maybe if it needs to set up the host as IPA client, it can leverage that from another module though.

ares:
This module is for (foreman + ipa) integration only. It does not install IPA and it does not enroll machine to IPA. It installs authentication modules, configures apache and foreman to use IPA auth and adds some pam and sssd stuff. @adelton could bring more light I guess.

Maybe we could move this integration part to puppet-ipa but it does not make much sense to me. Do all users of puppet-ipa want to have also a class to integrate it with foreman?

adelton:
> ```
> You need a puppet module for each separate component you want to "puppet".
> Eg: one puppet-ipa module, one puppet-gluster module, one puppet-pulp module, and perhaps even a puppet-foreman module. Why there would be a "puppet-foreman-ipa" module isn't quite clear to me.
> ```

Because it configures Foreman to authenticate against IPA. It's not a module to deploy or manage IPA itself. At this point, the primary use case is Foreman installer, to avoid the need for (majority of the) manual steps in http://theforeman.org/manuals/1.5/index.html#5.7ExternalAuthentication

purpleidea:
> Maybe we could move this integration part to puppet-ipa but it does not make much sense to me. Do all users of puppet-ipa want to have also a class to integrate it with foreman?

Agreed, doesn't make sense. I was misled into thinking that this was another FreeIPA module to setup FreeIPA core, not Foreman integration... Now that I know my work isn't in vain, I'll continue with my puppet-ipa module.... Phew.

There will be a time for better integration, but at the moment, both teams seem as if they're doing what's right for now.

Cheers,
James

purpleidea:
@adelton 

> At this point, the primary use case is Foreman installer, to avoid the need for (majority of the) manual steps in http://theforeman.org/manuals/1.5/index.html#5.7ExternalAuthentication

Useful link, thanks!

domcleal:
As I mentioned on the other PR, it sounds like the useful thing to have in puppet-ipa would be the generic _client_ bits, i.e. enrolment and keytab fetching.  The useful things in puppet-foreman would be httpd configuration (since we already do this), then use the two together.

adelton:
> useful thing to have in puppet-ipa would be the generic client bits, i.e. enrolment and keytab fetching

But puppet-ipa as I understand it is primarily for setting up the FreeIPA server, _not_ client. Yes, there are client parts there as well but I'm not really fond of having them in the same module.

purpleidea:
> But puppet-ipa as I understand it is primarily for setting up the FreeIPA server, not client. Yes, there are client parts there as well but I'm not really fond of having them in the same module.

Actually it does both. It's up to the user about which they use, but there is a strong use case of offering them in the same module: automatic configuration. Puppet can automatically export the OTP to the client to get it to enroll without any manual fooling around. Example:
https://github.com/purpleidea/puppet-ipa/blob/master/examples/simple-usage2.pp

ares:
To sum up, each module solves something different, logic in foreman_ipa is wanted in our installer and later it could hopefully use some parts from puppet-ipa and/or puppet-ipaclient. Next step is to move foreman_ipa to puppet-foreman. We'll get to that in following days.

ares:
Closing in favor of https://github.com/theforeman/puppet-foreman/pull/199

PR #122, opened 2014-07-07
Add --detailed-exitcodes arg

domcleal:


ares:
ACK (I should move code from bin script to hooks though since it's getting bigger and bigger)

domcleal:
I guess that'd also fix http://projects.theforeman.org/issues/5488

bin/foreman-installer
```diff
@@ -34,6 +34,10 @@ Kafo::KafoConfigure.app_option '--reset-foreman-db', :flag,
                                'manually. e.g. dropping can fail when DB is currently in use.',
     :default => false
 
+Kafo::KafoConfigure.app_option '--detailed-exitcodes', :flag,
+                               "Provide transaction information via exit codes, see puppet-agent(8)\n" +
+                               'full details.', :default => false
```
GregSutcliffe: nitpick, should read "for full details" ?

ares:
@domcleal yes, you're right

bin/foreman-installer
domcleal: @GregSutcliffe thanks, fixed

GregSutcliffe:
Merged as 601047d, thanks!

PR #126, opened 2014-07-28
Switch back to official pl-apache release & pl-postgresql upstream git

domcleal:
pl-apache was released the other day, and pl-postgresql was released (for EL7), but remained broken for F19.


ekohl:
:+1:

PR #127, opened 2014-08-19
Revert to released postgresql module for Fedora fix

domcleal:
We can now rely on https://github.com/puppetlabs/puppetlabs-postgresql/blob/3.4.x/CHANGELOG.md#2014-07-31---supported-release-342.


ekohl:
Yay for cleaning up :)

PR #128, opened 2014-10-13
Fixes #5488 - fixed hook deprecation warning

mbacovsky:
- code from installer executable was moved to hooks
- services details in post-install message printed out on success only

### DIFF
--- a/hooks/post/10-post_install_message.rb
+++ b/hooks/post/10-post_install_message.rb
@@ -1,0 +1,26 @@
+# Puppet status codes say 0 for unchanged, 2 for changed succesfully
+if [0,2].include? @kafo.exit_code
+  say "  <%= color('Success!', :good) %>"
+
+  # Foreman UI?
+  if @kafo.config.module_enabled? 'foreman'
+    say "  * <%= color('Foreman', :info) %> is running at <%= color('#{param('foreman','foreman_url').value}', :info) %>"
+    say "      Initial credentials are <%= color('#{param('foreman', 'admin_username').value}', :info) %> / <%= color('#{param('foreman', 'admin_password').value}', :info) %>" if param('foreman','authentication').value == true
+  end
+
+  # Proxy?
+  if @kafo.config.module_enabled? 'foreman_proxy'
+    say "  * <%= color('Foreman Proxy', :info) %> is running at <%= color('#{param('foreman_proxy','registered_proxy_url').value}', :info) %>"
+  end
+
+  # Puppetmaster?
+  if ( @kafo.config.module_enabled?('puppet') && ( param('puppet','server').value != false ) )
+    say "  * <%= color('Puppetmaster', :info) %> is running at <%= color('port #{param('puppet','server_port').value}', :info) %>"
+  end
+else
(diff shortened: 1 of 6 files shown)


ares:
s/fixed/fixes/ in commit message (AFAIK our redmine don't match fixed)

config/foreman-installer.yaml
```diff
@@ -7,6 +7,7 @@
 :installer_dir: .
 # Uncomment if you want to load puppet modules from a specific path, $pwd/modules is used by default
 :modules_dir: ./_build/modules
+:hooks_dir: "./hooks"
```
ares: why is this needed? kafo adds `root_dir + '/hooks'` by default

EDIT:
also the option name is hook_dirs and is expected to be an array

mbacovsky:
Commit message fixed, `hooks_dir` removed.

hooks/post/10-post_install_message.rb
```diff
@@ -0,0 +1,37 @@
+# helpers
+def module_enabled?(name)
```
ares: this is no longer needed, see https://github.com/theforeman/kafo/blob/master/lib/kafo/configuration.rb#L126,  https://github.com/theforeman/kafo/blob/master/lib/kafo/hook_context.rb#L13

you should be able to use `@kafo.config.module_enabled?(name)`

hooks/post/10-post_install_message.rb
```diff
@@ -0,0 +1,37 @@
...
+  mod = @kafo.module(name)
+  return false if mod.nil?
+  mod.enabled?
+end
+
+def get_param(mod, name)
```
ares: this is no longer needed, see https://github.com/theforeman/kafo/blob/master/lib/kafo/hook_context.rb#L47

you should be able to use `param('foreman', 'admin_username').value` directly

domcleal:
While I remember, we should check (or just bump to $current... either way) the minimum kafo version and update the packaging to match.

mbacovsky:
Thanks for the hints @ares, helpers were removed.
I'll update the packaging @domcleal, I wouldn't mind if there are volunteers for debs though ;)

domcleal:
Oh sure, that's easy enough.  We have almost up-to-date kafo packages, so it's just bumping the min version.

mbacovsky:
Found out that build task in the Rakefile didn't included hooks directory plus one bug in the exit_code handling. Fixed both, updated

mbacovsky:
Bumped kafo version in rpm package dependences https://github.com/theforeman/foreman-packaging/pull/393

ares:
lgtm I'll have to do the testing

ares:
@mbacovsky would you mind uploading links to scratch builds you used for testing?

ares:
nvm, built myself one and testing with it

mbacovsky:
@ares One is linked in the foreman-packaging PR I mentioned above

ares:
Merged as https://github.com/theforeman/foreman-installer/commit/40fc22eecd3f313dfa63870f9513d54f355b1f86, thanks @mbacovsky!

PR #130, opened 2014-11-04
adding ovirt_provision plugin to installer options

bronhaim:
Adding an option to install ovirt_provision plugin with foreman-installer 

Signed-off-by: Yaniv Bronhaim ybronhei@redhat.com


config/foreman-installer.yaml
```diff
@@ -52,6 +52,9 @@
     :dir_name: foreman
     :manifest_name: plugin/discovery
     :params_name: plugin/discovery/params
+  :foreman::plugin:ovirt_provision:
```
domcleal: missing colon

PR #132, opened 2014-11-12
fixes #8008 - display foreman, foreman-proxy and puppet information

shlomizadok:
display foreman, foreman-proxy and puppet regradless to success or fail


domcleal:
Merged, thanks @shlomizadok!

PR #135, opened 2014-12-09
Disable foreman_bootdisk by default

domcleal:
I don't think this was a good idea, so would like to remove it again from the default install so it's in the same boat as other plugins.

### DIFF
--- a/config/answers.yaml
+++ b/config/answers.yaml
@@ -16,7 +16,7 @@ foreman_proxy:
 puppet:
   server: true
 foreman::cli: true
-foreman::plugin::bootdisk: true
+foreman::plugin::bootdisk: false
 foreman::plugin::chef: false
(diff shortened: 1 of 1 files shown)


ekohl:
Not against it, but what made you think it's not a good idea? Too many issues?

domcleal:
No, not really, it's been pretty much stable.  I think I was anticipating it moving into core sooner than it is (that is, I have no plan to do the work any time soon) and don't feel that it's any more special than other plugins.

I'm also on a bit of a drive to minimise the default installation as every rubygem slows down boot times.

ekohl:
Fair enough :+1:

mmoll:
merged as 0a4ba0c :boot: :floppy_disk: , thanks @domcleal

PR #136, opened 2015-01-01
check for 127.0.1.1 in /etc/hosts

mmoll:
This was added to the installation instructions, but still people are showing up in IRC with this problem.

### DIFF
--- a/checks/debian_hosts.rb
+++ b/checks/debian_hosts.rb
@@ -1,0 +1,19 @@
+#!/usr/bin/env ruby
+
+MESSAGE = %q(Ensure that
+ping $(hostname -f)
+shows the real IP address, not 127.0.1.1.
+
+Change or remove this entry from /etc/hosts
+)
+
+def match_hosts_entry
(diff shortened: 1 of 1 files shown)


checks/debian_hosts.rb
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env ruby
+
+begin
```
ekohl: Why this optional import?

mmoll: Copied over from hostname.rb, I think that's because Ruby 1.8 needs that `require 'rubygems'` explicitly.

mmoll: hrm, looking again, I don't use any gems in that file, so that can indeed go

ekohl:
:+1:

domcleal:
Thanks @mmoll!

lzap:
Nice, I think this could be named differently as this is also useful for non-Debian hosts, but whatever :-)

PR #141, opened 2015-01-27
Revert "check for 127.0.1.1 in /etc/hosts"

domcleal:
This reverts commit fbec366eb67faac000b6b15938cbd1bfcc2fd52a.

The smart proxy now uses SSL CNs for authorisation, via ticket [#7849](http://projects.theforeman.org/issues/7849), so the
reverse DNS now isn't critical for installation.

---

Thoughts?  I pushed the introduction of debian_hosts.rb to 1.7-stable for 1.7.2 as it's still important there.


ekohl:
:+1:

mmoll:
makes sense, thanks @domcleal

PR #143, opened 2015-03-08
Refs #7529 - sets compute and plugins to low priority

ares:
Depends on Kafo having https://github.com/theforeman/kafo/pull/103


mmoll:
Maybe also set proxy plugins to low priority?

ares:
Good catch, added

mmoll:
:+1:

domcleal:
Merged as 6d5c890, thanks @ares!

PR #147, opened 2015-06-07
Update forge URL

ekohl:


mmoll:
:+1:, working fine here.

However, this might introduce a minimum version of librarian-puppet on the Jenkins host.

domcleal:
Thanks @ekohl.

I think that's fine, we install librarian-puppet from gem each time so it's picking up whatever the latest is.

PR #148, opened 2015-06-09
[do not merge] Validate modules/classes against supported OSes

domcleal:
Allow each module to ship a foreman-installer.json file containing:

```
{"example::class":{"operatingsystem_support":{ .. }}
```

and validate each kafo module (class) against the OS list.  If it isn't
listed then disable the module before execution and log a warning.

The operatingsystem_support hash matches Puppet's metadata.json format,
with a hash of facts to supported values.

---

Marked as DNM as this requires [a newer kafo](https://github.com/theforeman/kafo/pull/115), some metadata in our modules and probably some discussion.

It attempts to fix the problem described here: https://github.com/theforeman/foreman-bats/pull/72#issuecomment-110033421

Using the installer to install all of our plugins doesn't always work as some aren't shipped on certain OSes, either because they're unsupported or it hasn't been done yet.  I think this means we ought to put the support matrix into a data file somewhere, either in foreman-installer or the modules themselves, which lets us either disable the options or advise the user somehow.  (In this PR, I only advise the user if they enable/select it, and ignore their choice.)

We could put the data in foreman-installer, which might be easier to handle with branching and releasing, or by putting it in the modules, we can change it at the same time as we make any changes to the plugin::\* classes.

Either way, there's always a risk that the data file will be out of date compared to the contents of our repo.  To get around this, I added the --ignore-os-support switch in case a user knows better.

### DIFF
--- a/hooks/boot/12-ignore_os_support.rb
+++ b/hooks/boot/12-ignore_os_support.rb
@@ -1,0 +1,5 @@
+app_option(
+  '--ignore-os-support', :flag,
+  "Ignore checks for supported operating systems and attempt to install anyway"
+  :default => false
+)
--- a/hooks/pre_commit/10-apply_os_support.rb
+++ b/hooks/pre_commit/10-apply_os_support.rb
@@ -1,0 +1,46 @@
+require 'facter'
+require 'json'
+
+# Check OS support for top-level modules and disable those that aren't known
+# to work on the current OS.
+def apply_os_support
+  kafo.config.modules.each do |mod|
+    metajson = File.join(KafoConfigure.modules_dir, mod.dir_name, 'foreman-installer.json')
+    next unless File.exist?(metajson)
+
+    meta = JSON.load(File.read(metajson))
+    if meta[mod.identifier] && meta[mod.identifier]['operatingsystem_support']
+      supported = meta[mod.identifier]['operatingsystem_support'].any? do |os|
+        os.all? { |fname,fvalue| os_fact_matches?(fname, fvalue) }
+      end
+
+      unless supported
+        logger.warn("Skipping module #{mod.name} as it isn't supported on this operating system. Please check the plugin documentation for its supported OSes, or use --ignore-os-support to attempt to run this module.")
+        mod.disable
+      end
+    end
+  end
+end
+
+def os_fact_matches?(name, value)
+  os_value = [os_fact(name)].flatten.map(&:to_s)
+  [value].flatten.any? do |value|
+    os_value.include? value.to_s
+  end
+end
+
+def os_fact(name)
+  # metadata style uses osrelease when it really means the major release
+  if name == 'operatingsystemrelease'
+    maj = Facter.value(:operatingsystemmajrelease)
+    return maj if maj
+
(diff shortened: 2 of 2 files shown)


ekohl:
I'd like to see some example in a README somewhere. Otherwise this looks like a sane solution (as a user, can't judge the ruby code that well).

ares:
I wonder if it makes sense and how hard it would be to read this information directly from puppet manifest rather than from a new file `foreman-installer.json`. We already introduced rdoc DSL so maybe we could just enhance it and maybe make it even core part of kafo/kafo_parsers? (I know I told you to use kafo hooks at first). At least I could make Kafo parse such information and the check itself could still remain as a foreman-installer hook.

ekohl:
@ares I considered the same thing but I wonder if we want to clutter our modules with that info. Another benefit of doing it in the installer is that typically you package things async from the modules. That allows you to package a plugin in a .1 release instead of .0 without having to update the module.

domcleal:
@ekohl sorry, I'm not sure I understand your last comment?

To be clear, the current implementation stores a foreman-installer.json inside each module, which would require re-releasing the module to update the metadata.

ekohl:
@domcleal I'll blame that on not being fully awake when I wrote that.

What I meant was that I think I prefer the installer.json approach because I expect that it has a lower overhead due to releasing. Often a class can't be used because something isn't packaged (yet) for Debian. By marking that in the installer.json, we can package it, bump the installer and keep the modules clean. Most forge users will not care about it and if we bump there they will need to spend time to look at the changelogs. Also, if we bump a module then we need to bump the installer anyway.

Is that a bit more coherent? :)

domcleal:
Yeah, so you're saying keep the file in this repo rather than modules, OK.

domcleal:
> I wonder if it makes sense and how hard it would be to read this information directly from puppet manifest rather than from a new file foreman-installer.json. We already introduced rdoc DSL so maybe we could just enhance it and maybe make it even core part of kafo/kafo_parsers?

I guess kafo already has some support for evaluating conditionals: https://github.com/theforeman/kafo#conditional-parameters-in-interactive-mode

So what you're proposing could be similar, allowing it to be expressed at the class level with facts (rather than other parameters) - though the conditionals would get rather long for the number of OSes we express today in metadata.json.  Or do it differently and more like the list of OSes/versions that's proposed here or used in metadata.json.

I can't say I'm a big fan of a large conditional expression living in the class documentation, it makes me feel very uneasy.  Probably a list of OSes like metadata.json is acceptable though?

It also means having module-level metadata rather than foreman-installer level.

ares:
Just a side note, conditions are written in ruby so we could introduce some helpers to make condition nearly just a list of supported OSes. But we'd probably duplicate info from metadata.json (unless we combine it somehow).

If we want to disable modules because we don't have packages, it makes more sense to keep it in installer rather than modules. Also metadata.json does not always correlate with what we support in our installer. So yeah, I agree one config file in installer. Still could be useful if built-in in Kafo

ekohl:
So is this still something we want?

domcleal:
I think so, but I'm uncertain how to implement it at present. I'll close this PR until I can reconsider it, thanks for checking.

PR #149, opened 2015-06-09
Fixes #10766 - fail nicely if puppet certs were already generated

mbacovsky:
Puppet is not able to generate proper CA certificates when the client certificate was already generated. In practice it checks if the `/var/lib/puppet/ssl` exists and if so it skips the CA cert generation.

With the existing possibility of having custom ssl dir and existing puppet CA elsewhere it is difficult to detect the situation. 
With the default installation the installer fails to start httpd on missing 
- revocation list for foreman apache (`/var/lib/puppet/ssl/ca/ca_crl.pem`)
- ca cert when installed with passenger and puppet ran as master (`/var/lib/puppet/ssl/ca/ca_crt.pem`)

My proposal is to add hook guessing if these  these two files will be needed and failing with suggestion of possible fixes.

### DIFF
--- a/hooks/pre_commit/05-puppet_certs_exist.rb
+++ b/hooks/pre_commit/05-puppet_certs_exist.rb
@@ -8,20 +8,22 @@ def param_value(mod, name)
   param(mod, name).value if param(mod, name)
 end
 
+puppet_ca_enabled = param_value('puppet', 'server_ca')
 server_crl_path = param_value('foreman', 'server_ssl_crl')
 puppet_server_implementation = param_value('puppet', 'server_implementation')
 puppet_passenger = param_value('puppet', 'server_passenger')
 cert_dir = param_value('puppet', 'server_ssl_dir')
+key_path = param_value('foreman', 'server_ssl_key')
 puppet_ca_cert = File.join(cert_dir, 'ca/ca_crt.pem')
 puppet_enabled = kafo.module('puppet').enabled?
 
 client_message = "- is Puppet already installed without Puppet CA? You can remove the existing certificates with 'rm -rf #{cert_dir}' to get Puppet CA properly configured."
 
-if cert_dir && File.directory?(cert_dir)
-  if !server_crl_path.empty? && !File.exists?(server_crl_path)
-    error "The file #{server_crl_path} does not exist.\n #{client_message}\n - if you set custom revocation list (--foreman-server-ssl-crl) make sure the file exists."
-  end
+if puppet_ca_enabled && key_path && File.exists?(key_path) && !kafo.skip_checks_i_know_better?
   if puppet_enabled && puppet_passenger && puppet_server_implementation == 'master' && !File.exists?(puppet_ca_cert)
     error "The file #{puppet_ca_cert} does not exist.\n #{client_message}\n - if you use custom Puppet SSL directory (--puppet-server-ssl-dir) make sure the directory exists and contain the CA certificate.\n   Also make sure --enable-puppet, --puppet-server-passenger and --puppet-server-implementation are set properly"
+  end
+  if !server_crl_path.empty? && !File.exists?(server_crl_path)
+    error "The file #{server_crl_path} does not exist.\n #{client_message}\n - if you set custom revocation list (--foreman-server-ssl-crl) make sure the file exists."
   end
 end
--- a/hooks/pre_commit/05-puppet_certs_exist.rb
+++ b/hooks/pre_commit/05-puppet_certs_exist.rb
@@ -1,0 +1,27 @@
+def error(message)
+  say message
+  logger.error message
+  kafo.class.exit 101
+end
+
+def param_value(mod, name)
+  param(mod, name).value if param(mod, name)
+end
+
+server_crl_path = param_value('foreman', 'server_ssl_crl')
+puppet_server_implementation = param_value('puppet', 'server_implementation')
+puppet_passenger = param_value('puppet', 'server_passenger')
+cert_dir = param_value('puppet', 'server_ssl_dir')
+puppet_ca_cert = File.join(cert_dir, 'ca/ca_crt.pem')
+puppet_enabled = kafo.module('puppet').enabled?
+
+client_message = "- is Puppet already installed without Puppet CA? You can remove the existing certificates with 'rm -rf #{cert_dir}' to get Puppet CA properly configured."
+
+if cert_dir && File.directory?(cert_dir)
+  if !server_crl_path.empty? && !File.exists?(server_crl_path)
+    error "The file #{server_crl_path} does not exist.\n #{client_message}\n - if you set custom revocation list (--foreman-server-ssl-crl) make sure the file exists."
+  end
+  if puppet_enabled && puppet_passenger && puppet_server_implementation == 'master' && !File.exists?(puppet_ca_cert)
+    error "The file #{puppet_ca_cert} does not exist.\n #{client_message}\n - if you use custom Puppet SSL directory (--puppet-server-ssl-dir) make sure the directory exists and contain the CA certificate.\n   Also make sure --enable-puppet, --puppet-server-passenger and --puppet-server-implementation are set properly"
+  end
+end


hooks/pre_validations/05-puppet_certs_exist.rb
```diff
@@ -0,0 +1,27 @@
...
+puppet_ca_cert = File.join(cert_dir, 'ca/ca_crt.pem')
+puppet_enabled = kafo.module('puppet').enabled?
+
+client_message = "- is puppet already installed without puppet CA? You can remove the existing certificates with 'rm -rf #{cert_dir}' to get puppet CA properly configured."
+
+if cert_dir && File.directory?(cert_dir)
+  if !server_crl_path.empty? && !File.exists?(server_crl_path)
+    error "The file #{server_crl_path} does not exist.\n #{client_message}\n - if you set custom revocation list (--foreman-server-ssl-crl) make sure the file exists."
+  end
+  if puppet_enabled && puppet_passenger && puppet_server_implementation == 'master' && !File.exists?(puppet_ca_cert)
```
domcleal: Why does this depend on puppet_passenger and server_implementation?  I think having it depend on puppet::server_ca would be more accurate, as that's when having client certs only would be a problem.

hooks/pre_validations/05-puppet_certs_exist.rb
```diff
@@ -0,0 +1,27 @@
...
+end
+
+server_crl_path = param_value('foreman', 'server_ssl_crl')
+puppet_server_implementation = param_value('puppet', 'server_implementation')
+puppet_passenger = param_value('puppet', 'server_passenger')
+cert_dir = param_value('puppet', 'server_ssl_dir')
+puppet_ca_cert = File.join(cert_dir, 'ca/ca_crt.pem')
+puppet_enabled = kafo.module('puppet').enabled?
+
+client_message = "- is puppet already installed without puppet CA? You can remove the existing certificates with 'rm -rf #{cert_dir}' to get puppet CA properly configured."
```
domcleal: capitalise "Puppet" throughout

hooks/pre_validations/05-puppet_certs_exist.rb
```diff
@@ -0,0 +1,27 @@
...
+
+if cert_dir && File.directory?(cert_dir)
```
domcleal: The Puppet exec checks for existence of the SSL cert, not the dir, so it'd be best to mirror that.

domcleal:
Perhaps use a pre_commit hook (new feature) rather than pre_validations, so it runs after the wizard?

hooks/pre_validations/05-puppet_certs_exist.rb
mbacovsky: @domcleal, from my experiments it seems that the default installation fails only when the apache starts the puppet master. I'm not sure if Foreman can be installed without Puppet CA or puppet master so I limited the CA cert check to just when the puppet master is used. Wasn't it valid concern?

mbacovsky:
Yes, `pre_commit` phase is better place to run the hook. Updated, @domcleal

mbacovsky:
@domcleal is anything I can help with preventing merging of this PR?

domcleal:
Sorry, had the tab open and trying to get around to it.  I need to have a think about the parameters that the check's on.

hooks/pre_commit/05-puppet_certs_exist.rb
```diff
@@ -0,0 +1,27 @@
...
+server_crl_path = param_value('foreman', 'server_ssl_crl')
+puppet_server_implementation = param_value('puppet', 'server_implementation')
+puppet_passenger = param_value('puppet', 'server_passenger')
+cert_dir = param_value('puppet', 'server_ssl_dir')
+puppet_ca_cert = File.join(cert_dir, 'ca/ca_crt.pem')
+puppet_enabled = kafo.module('puppet').enabled?
+
+client_message = "- is Puppet already installed without Puppet CA? You can remove the existing certificates with 'rm -rf #{cert_dir}' to get Puppet CA properly configured."
+
+if cert_dir && File.directory?(cert_dir)
```
domcleal: I don't think this is precise enough, as any time the agent has run, you might have the directory (since it'll create a CSR).  This will particularly be the case on a Debian installation where the package install usually starts the Puppet agent service.

hooks/pre_commit/05-puppet_certs_exist.rb
```diff
@@ -0,0 +1,27 @@
+def error(message)
+  say message
+  logger.error message
+  kafo.class.exit 101
```
domcleal, line 4: I'm not certain where the bug is, but I get this error:

<pre>
/usr/lib/ruby/gems/1.8/gems/kafo-0.6.11/lib/kafo/logger.rb:104:in `error': wrong number of arguments (3 for 1) (ArgumentError)
        from /usr/lib/ruby/gems/1.8/gems/kafo-0.6.11/lib/kafo/logger.rb:104:in `send'
        from /usr/lib/ruby/gems/1.8/gems/kafo-0.6.11/lib/kafo/logger.rb:104:in `dump_buffer'
        from /usr/lib/ruby/gems/1.8/gems/kafo-0.6.11/lib/kafo/logger.rb:104:in `each'
        from /usr/lib/ruby/gems/1.8/gems/kafo-0.6.11/lib/kafo/logger.rb:104:in `dump_buffer'
        from /usr/lib/ruby/gems/1.8/gems/kafo-0.6.11/lib/kafo/logger.rb:103:in `each'
        from /usr/lib/ruby/gems/1.8/gems/kafo-0.6.11/lib/kafo/logger.rb:103:in `dump_buffer'
        from /usr/lib/ruby/gems/1.8/gems/kafo-0.6.11/lib/kafo/logger.rb:94:in `dump_errors'
        from /usr/lib/ruby/gems/1.8/gems/kafo-0.6.11/lib/kafo/logger.rb:99:in `dump_errors'
        from /usr/lib/ruby/gems/1.8/gems/kafo-0.6.11/lib/kafo/exit_handler.rb:26:in `exit'
        from /usr/lib/ruby/gems/1.8/gems/kafo-0.6.11/lib/kafo/kafo_configure.rb:125:in `exit'
        from /usr/share/foreman-installer/hooks/pre_commit/05-puppet_certs_exist.rb:4:in `error'
        from /usr/share/foreman-installer/hooks/pre_commit/05-puppet_certs_exist.rb:22:in `load'
        from /usr/lib/ruby/gems/1.8/gems/kafo-0.6.11/lib/kafo/hook_context.rb:13:in `instance_exec'
        from /usr/lib/ruby/gems/1.8/gems/kafo-0.6.11/lib/kafo/hook_context.rb:13:in `execute'
        from /usr/lib/ruby/gems/1.8/gems/kafo-0.6.11/lib/kafo/hooking.rb:50:in `execute'
        from /usr/lib/ruby/gems/1.8/gems/kafo-0.6.11/lib/kafo/hooking.rb:48:in `each'
        from /usr/lib/ruby/gems/1.8/gems/kafo-0.6.11/lib/kafo/hooking.rb:48:in `execute'
        from /usr/lib/ruby/gems/1.8/gems/kafo-0.6.11/lib/kafo/kafo_configure.rb:105:in `execute'
        from /usr/lib/ruby/gems/1.8/gems/clamp-1.0.0/lib/clamp/command.rb:68:in `run'
        from /usr/lib/ruby/gems/1.8/gems/clamp-1.0.0/lib/clamp/command.rb:133:in `run'
        from /usr/lib/ruby/gems/1.8/gems/kafo-0.6.11/lib/kafo/kafo_configure.rb:119:in `run'
        from /usr/sbin/foreman-installer:18
</pre>

hooks/pre_commit/05-puppet_certs_exist.rb
```diff
@@ -0,0 +1,27 @@
...
+puppet_ca_cert = File.join(cert_dir, 'ca/ca_crt.pem')
+puppet_enabled = kafo.module('puppet').enabled?
+
+client_message = "- is Puppet already installed without Puppet CA? You can remove the existing certificates with 'rm -rf #{cert_dir}' to get Puppet CA properly configured."
+
+if cert_dir && File.directory?(cert_dir)
+  if !server_crl_path.empty? && !File.exists?(server_crl_path)
+    error "The file #{server_crl_path} does not exist.\n #{client_message}\n - if you set custom revocation list (--foreman-server-ssl-crl) make sure the file exists."
+  end
+  if puppet_enabled && puppet_passenger && puppet_server_implementation == 'master' && !File.exists?(puppet_ca_cert)
```
domcleal, line 23: Probably related to the if cert_dir check, but this conditional might be true even when the installer is going to generate the certificate.  I think it should check if the host certificate is present, as in that case the installer won't set up a new CA.  It's possible for the CA cert to be missing and for the installer to then generate a new CA.

hooks/pre_commit/05-puppet_certs_exist.rb
```diff
@@ -0,0 +1,27 @@
...
+end
+
+server_crl_path = param_value('foreman', 'server_ssl_crl')
+puppet_server_implementation = param_value('puppet', 'server_implementation')
+puppet_passenger = param_value('puppet', 'server_passenger')
+cert_dir = param_value('puppet', 'server_ssl_dir')
+puppet_ca_cert = File.join(cert_dir, 'ca/ca_crt.pem')
+puppet_enabled = kafo.module('puppet').enabled?
+
+client_message = "- is Puppet already installed without Puppet CA? You can remove the existing certificates with 'rm -rf #{cert_dir}' to get Puppet CA properly configured."
```
domcleal, line 20: I'd only suggest rm -rf when the user intends this to be a CA (::puppet::ca_server == true), as when it's not a CA you actually want to copy the missing files over instead of deleting the ones you have.

mbacovsky:
@domcleal, I totaly forget on this PR, sorry for that. In the new update I tried to address your comments:
- the checks are only performed when server CA is true
- existence of puppet is tested on existence of host priv key rether then on presence of ssl directory 
- I tested the hook with foreman 1.11 and was not able to reproduce the issue with logger. Did you used some custom logging setup?
- the checks can be skipped with --skip-checks-i-know-better switch

mmoll:
what's the status here?

mbacovsky:
@mmoll, I believe all @domcleal 's comments were addressed. and now it is waiting for re-review/re-testing.

ehelms:
Due to the age of the PR, I'm going to opt to close. If folks would like to see this PR get in please re-open.

PR #150, opened 2015-07-03
Pin puppetlabs/postgresql to avoid MODULES-2185

domcleal:
Affects config of PostgreSQL on Puppet 2.7.

---

My PR's been merged into master, but it's unlikely to result in another 4.4.x release to fix the regression.  Since 4.5.0's not likely to be for a month or two, pin temporarily?


ekohl:
Thanks!

PR #151, opened 2015-07-17
puppetlabs-puppetdb 5.x is incompatible

mmoll:
with theforeman-puppet

this should also go into 1.8 and 1.9


domcleal:
Thanks @mmoll!

domcleal:
No need to backport this to 1.8/1.9 as I pin all modules to minor versions after branching.

mmoll:
Indeed. :bow:

PR #155, opened 2015-09-03
Add new foreman::plugin::*::params classes to kafo config

domcleal:
As added in https://github.com/theforeman/puppet-foreman/commit/a64c58e85e0b1e50fbf4e8ee838951a53c687d2e


ekohl:
Thanks!

PR #156, opened 2015-09-08
DO NOT MERGE: fixes #11735 - adding initial config and answers for smart-proxy scenario

bbuckingham:
This PR contains an initial answer and config file to support 'smart-proxy' as a scenario.  It is based upon the 'scenario' feature that is being introduced in to kafo with the following: https://github.com/theforeman/kafo/pull/117

### DIFF
--- a/config/smart-proxy-answers.yaml
+++ b/config/smart-proxy-answers.yaml
@@ -1,0 +1,22 @@
+# Format:
+# <classname>: false - don't include this class
+# <classname>: true - include and use the defaults
+# <classname>:
+#   <param>: <value> - include and override the default(s)
+#
+# Every support plugin/compute class is listed, so that it
+# shows up in the interactive menu
+#
+# See params.pp in each class for what options are available
+---
+foreman_proxy:
+  custom_repo: true
+puppet: false
+foreman_proxy::plugin::abrt: false
+foreman_proxy::plugin::chef: false
+foreman_proxy::plugin::dns::powerdns: false
+foreman_proxy::plugin::dynflow: false
+foreman_proxy::plugin::openscap: false
+foreman_proxy::plugin::pulp: false
+foreman_proxy::plugin::remote_execution::ssh: false
+foreman_proxy::plugin::salt: false
--- a/config/smart-proxy.yaml
+++ b/config/smart-proxy.yaml
@@ -1,0 +1,152 @@
+# foreman-installer main configuration file
+# note current configuration is written to foreman-installer.yaml every time foreman-install is run
+
+## Installer configuration
+# Path to answer file
+:answer_file: /etc/foreman/installer-scenarios.d/smart-proxy-answers.yaml
+:installer_dir: /usr/share/foreman-installer
+# Uncomment if you want to load puppet modules from a specific path, $pwd/modules is used by default
+:module_dirs: /usr/share/foreman-installer/modules
+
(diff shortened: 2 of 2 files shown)


bbuckingham:
The following is an example:

foreman-installer --scenario smart-proxy \
  --foreman-proxy-tftp=false \
  --foreman-proxy-foreman-base-url=https://$FOREMAN \
  --foreman-proxy-trusted-hosts=$FOREMAN \
  --foreman-proxy-oauth-consumer-key=GzssNTJwax2mn84iguTB6NJw7xboiD5U \
  --foreman-proxy-oauth-consumer-secret=BzcKZXRAE8zjdQ466J87iFUK4cRREjXN

config/smart-proxy-answers.yaml
```diff
@@ -0,0 +1,42 @@
...
+# <classname>: true - include and use the defaults
+# <classname>:
+#   <param>: <value> - include and override the default(s)
+#
+# Every support plugin/compute class is listed, so that it
+# shows up in the interactive menu
+#
+# See params.pp in each class for what options are available
+---
+foreman: false
```
ehelms: I think this could be omitted.

bbuckingham: @ehelms, That appears correct.  I was able to remove all of the foreman specific params, since they all default to false.  The updated list now only has params for foreman_proxy.

domcleal:
Is this just a one-arg way to change the default answers?  e.g. no default enabling of Foreman?

There seems to be a lot of duplication with the config files we ship by default - can that be minimised?

bbuckingham:
@domcleal, You are correct that there is a lot of overlap (~90+%) between the smart-proxy.yaml and the base foreman configuration.  The main changes in this example are the location of the answers file and the name of the scenario.

This is something that I wanted to discuss with @mbacovsky when he returns.  It is possible that there is a simplification that can be done to this with the updated kafo or additional changes that can be made to it to reduce duplication.  For example, perhaps the installer could have (or already supports) a base config and then each scenario can simply add to or override it.

config/smart-proxy.yaml
```diff
@@ -0,0 +1,133 @@
+# foreman-installer main configuration file
+# note current configuration is written to foreman-installer.yaml every time foreman-install is run
+
+## Installer configuration
+# Path to answer file
+:answer_file: /etc/foreman/installer-scenarios.d/smart-proxy-answers.yaml
+:installer_dir: /usr/share/foreman-installer
+# Uncomment if you want to load puppet modules from a specific path, $pwd/modules is used by default
+:modules_dir: /usr/share/foreman-installer/modules
```
mbacovsky: `modules_dir` was renamed to `module_dirs` and can be a list. The old version is still supported but deprecated.

config/smart-proxy.yaml
```diff
@@ -0,0 +1,133 @@
...
+# :log_owner: root
+# :log_group: root
+# :config_header_file:
+# :dont_save_answers:
+
+# If using the Debian ruby-kafo package, uncomment this
+# :kafo_modules_dir: /usr/share/foreman-installer/modules
+
+## Kafo tuning, customization of core functionality
+:name: Smart Proxy
```
mbacovsky, line 29: `description` field describing purpose of the scenario would help. This string is displayed in `foreman-installer --list-scenarios`

PR #158, opened 2015-10-30
Remove .git and spec directories from tarballs

domcleal:
Reduces the size of the uncompressed tarball from about 6.3MiB to 2.0MiB, which will hopefully have an effect on the installed size too.

https://gist.github.com/domcleal/b00d61499a57483cbe5c is a diff of the file lists.


ehelms:
:+1:

ekohl:
Thanks!

PR #160, opened 2015-11-16
DO NOT MERGE: fixes #12498 - adding initial config and answers for capsule scenario.

bbuckingham:
This PR contains an initial answer and config file to support 'capsule' as a scenario. It is based upon the 'scenario' feature that is being introduced in to kafo with the following: theforeman/kafo#117

The goal is to enable the user to use the foreman-installer to install a capsule versus the existing capsule-installer.  This will help to remove duplication between the 2 installers and support the 'Katello on Foreman' feature.

The additional features that are added by a capsule are the 'pulp' (to support content management) and 'reverse proxy' (to support capsule isolation).

### DIFF
--- a/config/capsule.yaml
+++ b/config/capsule.yaml
@@ -1,0 +1,155 @@
+# foreman-installer main configuration file
+# note current configuration is written to foreman-installer.yaml every time foreman-install is run
+
+## Installer configuration
+# Path to answer file
+:answer_file: /etc/foreman/installer-scenarios.d/capsule-answers.yaml
+:installer_dir: /usr/share/foreman-installer
+# Uncomment if you want to load puppet modules from a specific path, $pwd/modules is used by default
+:module_dirs: /usr/share/foreman-installer/modules
+
+## Useful for development, e.g. when you want to move log files to local directory
+:log_dir: '/var/log/foreman-installer'
+:log_name: 'foreman-installer.log'
+:log_level: :debug
+
+# Change if you want to debug default answers for you modules, this directory holds default answers
+# :default_values_dir: /tmp
+
+## Advanced configuration - if not set it's ignored
+# :log_owner: root
+# :log_group: root
+# :config_header_file:
+# :dont_save_answers:
+
+# If using the Debian ruby-kafo package, uncomment this
+# :kafo_modules_dir: /usr/share/foreman-installer/modules
+
+## Kafo tuning, customization of core functionality
+:name: Capsule
+:description: Install a stand-alone Capsule.
+# :no_prefix: false
+:order:
+  - certs
+  - foreman
+  - capsule
+  - foreman_proxy
+  - puppet
+:low_priority_modules:
+  - foreman_proxy_plugin
+  - foreman_compute
+  - foreman_plugin
+
+# The mapping hash provides Kafo with the information to find plugin classes
+:mapping:
+  :foreman::cli:
+    :dir_name: foreman
+    :manifest_name: cli
+    :params_name: cli/params
+  :foreman::plugin::bootdisk:
+    :dir_name: foreman
+    :manifest_name: plugin/bootdisk
+  :foreman::plugin::puppetdb:
+    :dir_name: foreman
+    :params_name: plugin/puppetdb/params
+    :manifest_name: plugin/puppetdb
+  :foreman::plugin::hooks:
+    :dir_name: foreman
+    :manifest_name: plugin/hooks
+  :foreman::plugin::dhcp_browser:
+    :dir_name: foreman
+    :manifest_name: plugin/dhcp_browser
+  :foreman::plugin::digitalocean:
+    :dir_name: foreman
+    :manifest_name: plugin/digitalocean
+  :foreman::plugin::discovery:
+    :dir_name: foreman
+    :manifest_name: plugin/discovery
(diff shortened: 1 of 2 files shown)


bbuckingham:
The following is an example:

foreman-installer --scenario capsule \
--capsule-parent-fqdn "katello-centos7-nightly.example.com"\
--foreman-proxy-registered-proxy-url "https://katello-centos7-scenarios-2.example.com:9090"\
--foreman-proxy-oauth-consumer-key "QSJ2qCGV3zWyohCJfKxQBDCCjceBiTMy"\
--foreman-proxy-oauth-consumer-secret "DGKBJjXWfkyroiyDHNhYAFn9uTnUcJxa"\
--foreman-proxy-foreman-base-url "https://katello-centos7-nightly.example.com"\
--foreman-proxy-trusted-hosts "katello-centos7-nightly.example.com"\
--foreman-proxy-trusted-hosts "katello-centos7-scenarios-2.example.com"\
--capsule-pulp-oauth-secret "bQ5zUPVd2qVDY987D4VtD6Va9pDj4Y7J"\
--capsule-certs-tar "~/katello-centos7-scenarios-2.example.com-certs.tar"

domcleal:
Looks like this is part of Katello, not foreman-installer.

bbuckingham:
@domcleal, It is to support katello on foreman using the kafo scenario-based install.  Are you suggesting that these 2 scenario files be placed in a separate git repo?

ekohl:
@bbuckingham it mentions modules that foreman-installer doesn't ship (certs and capsule) and I'd like to avoid shipping broken configs.

config/capsule.yaml
```diff
@@ -0,0 +1,155 @@
...
+  - foreman
+  - capsule
+  - foreman_proxy
+  - puppet
+:low_priority_modules:
+  - foreman_proxy_plugin
+  - foreman_compute
+  - foreman_plugin
+
+# The mapping hash provides Kafo with the information to find plugin classes
```
ekohl, line 43: I wonder if we could make this mapping reusable. It will save a lot of work if you can just say:

``` yaml
:mappings:
- foreman-installer-mappings.yaml
- katello-installer-mappings.yaml
```

The reason to make it a list is that someone might want to extend an installer with additional modules. I wouldn't mind if katello-installer would be a pure addon to foreman-installer with just additional modules and additional configs/scenarios. Maybe you could also accept a string or a list.

bbuckingham, line 43: @ekohl, Agreed.  It would be nice to minimize some of the duplication in the configuration files; however, that is not yet supported by kafo.  @mbacovsky is aware of the current duplication and has reducing it as one of goals for kafo in the future, it just isn't there yet.  Ideally, we'd really like to have the ability to add katello to existing foreman sooner than later, so we are working within current constraints.

bbuckingham:
@ekohl, Good point on the additional modules (certs and capsule).  Before this could be merged, we would need to add those in; however, I assume that we would want to do that versus continuing the separate path that we have today with (foreman-installer vs katello-installer & capsule-installer).  The hope was that adding the kafo scenario support would enable us to add katello in as an additional option without disrupting the existing core foreman or foreman-proxy scenarios.

ekohl:
@bbuckingham given the current constraints I think the short term path should be to converge katello-installer and capsule-installer into just katello-installer while we get the other modules in a mergeable shape.

Longer term I'd prefer to have all foreman-specific modules (so that excludes apache etc) under theforeman organization. I don't think we're ready for that just yet and we should have a good discussion about it, but I'm certainly open to it.

config/capsule.yaml
mbacovsky, line 43: @ekohl, @bbuckingham  my original plan was to have `installer-modules.d/` containing mapping for installer modules one module per file, the files would have to also  contain some dependency and visibility information. Scenario config would have no mapping section and the mappings would be loaded from `installer-modules.d`. This would make us possible to install additional installer modules (e.g. discovery or salt or whatever) and make it available in all scenarios containing modules it depends on. We decided not to include this feature yet as it would bring extra complexity too early. 
Is it the time to start thinking about this?

ekohl, line 43: @mbacovsky It may be time to at least be thinking about this. I like the `installer-modules.d` approach since it splits the problem up, but could we (optionally) ship them inside our own modules? That way when a plugin is added, you only need to submit a PR to one repo to have it supported. You would need to think about an answers file, but defaulting to off sounds like a reasonable approach.

mbacovsky:
The scenarios were designed to allow to be packaged separately. We can have "installer plugins" and let users to ship extensions to our installer. I think it would be also easier to maintain and make it possible to follow individual release cycles of the plugins. According to the original plan in case of capsule the user would do

``` bash
$ yum install foreman-installer-katello
$ foreman-installer --scenario capsule
```

where `foreman-installer-katello` is a installer plugin package containing extra modules, hooks, checks and scenarios specific for the `Katello`. Tha package should depend on `foreman-installer` and therefore install it as a dependency.

I'd prefer same approach for other plugins where it makes sense.

Does it make sense? @ekohl, @bbuckingham?

ekohl:
@mbacovsky I think that makes sense

bbuckingham:
@mbacovsky, makes sense.  Closing this PR.

PR #161, opened 2015-12-11
Add foreman::plugin::cockpit to config

domcleal:


ekohl:
:+1:

mmoll:
Merged, thanks @domcleal!
Maybe this should be on the list for 1.10.1?

domcleal:
That's a little tricky as the puppet-foreman support isn't in a minor release (and we'd have to split out the other changes).

PR #164, opened 2016-01-29
Refs #13405: Pin puppetlabs/apache module to 1.8.x

ehelms:
Current 1.8.x branch includes fix for passenger.conf that otherwise
leads to inability to reach web application.


domcleal:
Merged as 207af51, thanks @ehelms.

PR #166, opened 2016-02-21
Refs #13405: unpin puppetlabs/apache module again

mmoll:
This reverts commit 207af51faab25fe69f05456def6966bfae6ceb37.

puppetlabs-apache master now contains Dominic's fix


ekohl:
:+1: for the code change though I didn't test it.

domcleal:
Merged, thanks @mmoll.  We're actually testing against the latest release rather than master (maybe something to change/fix), but 1.8.1 is indeed out now and contains the fix.

PR #167, opened 2016-02-23

domcleal:
Reintroduce Puppetfile.lock back into the stable branch to keep all
dependencies pinned at current versions, and to be able to better
version-control updates for both theforeman and external modules in
stable releases.

---

I'd like to try and fix the weird behaviour whereby we're not properly tagging what's in each release, so now using Puppetfile.lock on stable branches will let us know precisely what's in each release, and pull changes in only when we want them.

1.11.0-RC1 will therefore contain:

<pre>
$ librarian-puppet show
adrien-alternatives (0.3.0)
nanliu-staging (1.0.3)
puppet-extlib (0.11.0)
puppetlabs-apache (1.8.1)
puppetlabs-apt (2.2.1)
puppetlabs-concat (2.1.0)
puppetlabs-firewall (1.8.0)
puppetlabs-inifile (1.4.3)
puppetlabs-mysql (3.6.2)
puppetlabs-postgresql (4.7.1)
puppetlabs-puppetdb (4.3.0)
puppetlabs-stdlib (4.11.0)
puppetlabs-xinetd (1.5.0)
theforeman-dhcp (2.3.0)
theforeman-dns (3.2.0)
theforeman-foreman (5.1.0)
theforeman-foreman_proxy (2.5.0)
theforeman-git (1.6.0)
theforeman-puppet (4.3.0)
theforeman-tftp (1.7.0)
</pre>


Puppetfile
```diff
@@ -4,13 +4,12 @@ forge 'https://forgeapi.puppetlabs.com'
 mod 'puppetlabs/mysql'
 mod 'puppetlabs/postgresql'
 mod 'puppetlabs/puppetdb',      '< 5.0.0'
-mod 'puppetlabs/apache',        :git => 'https://github.com/puppetlabs/puppetlabs-apache', :ref => '1.8.x'
-mod 'theforeman/dhcp',          :git => 'https://github.com/theforeman/puppet-dhcp'
-mod 'theforeman/dns',           :git => 'https://github.com/theforeman/puppet-dns'
-mod 'theforeman/git',           :git => 'https://github.com/theforeman/puppet-git'
-mod 'theforeman/tftp',          :git => 'https://github.com/theforeman/puppet-tftp'
+mod 'theforeman/dhcp',          '>= 2.3.0 < 2.4.0'
+mod 'theforeman/dns',           '>= 3.2.0 < 3.4.0'
```
mmoll: shouldn't that be `< 3.3.0`?

mmoll:
Puppetfile.lock will get updated at each RC and/or patch release, I presume?
:+1:

Puppetfile
domcleal: Thanks!  Fixed.

domcleal:
@mmoll exactly.  I think the way we should do it is to just run `librarian-puppet update` against specific modules when we need to (e.g. having made a patch release of theforeman/*, or having a bug report that some other module is causing a problem), but not against all modules and only if we have something to update.  This way we can minimise changes from third party modules without listing them all individually too.

ekohl:
Haven't looked at the code yet but my :+1: for the general idea assuming we only do this for release branches and not develop.

domcleal:
@ekohl yep, only on stable branches - it doesn't make much sense on develop where we want to continually pull the latest.  I added it to the branching instructions at http://projects.theforeman.org/projects/foreman/wiki/Release_Process#Main-code-repos.

mmoll:
merged, thanks @domcleal!

PR #169, opened 2016-04-05
fixes #14475 - add and install config file migrations

domcleal:
To keep the migration files and applied file meeting packaging standards
they're symlinked into place, using this layout:

```
/etc/foreman-installer/scenarios.d/foreman.migrations ->
  /usr/share/foreman-installer/config/foreman.migrations
/etc/foreman-installer/scenarios.d/foreman.migrations/.applied ->
  /etc/foreman-installer/scenarios.d/foreman-migrations-applied
```

Includes migrations for current 1.10-stable to develop changes.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,4 +1,5 @@
 require 'rake/clean'
+require 'yaml'
 
 BUILDDIR = File.expand_path(ENV['BUILDDIR'] || '_build')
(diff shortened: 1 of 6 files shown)


mmoll:
I don't have an environment handy to test this, but it's looking good to me.
Thanks @domcleal!

PR #170, opened 2016-04-07
refs #14475 - change module mapping keys to symbols

domcleal:
Module names in the config mapping section should be symbols, so the
new migrations caused extra sections to be added. Recombine them in
another migration.


mmoll:
merged, thanks @domcleal!

PR #172, opened 2016-04-18
fixes #14453 - build and install Kafo parser cache

domcleal:
The parser cache will provide support for Puppet in AIO in the near
future with Kafo by not loading Puppet or puppet-strings, and provides
a performance boost to the installer startup. Requires Kafo 0.7.3.

Be careful to preserve mtimes of manifest files to ensure the cache
validity. Extra kafo-export-params search dirs were removed as all
current packages install correctly into /usr/bin or PATH, and it is
now required to build.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -15,8 +15,14 @@ DATAROOTDIR = DATADIR = ENV['DATAROOTDIR'] || "#{PREFIX}/share"
 MANDIR = ENV['MANDIR'] || "#{DATAROOTDIR}/man"
 PKGDIR = ENV['PKGDIR'] || File.expand_path('pkg')
 
+exporter_dirs = ENV['PATH'].split(':').push('/usr/bin', ENV['KAFO_EXPORTER'])
+exporter = exporter_dirs.find { |dir| File.executable? "#{dir}/kafo-export-params" } or
+  raise("kafo-export-paths is missing from PATH, install kafo")
+
 directory BUILDDIR
 directory PKGDIR
+directory "#{BUILDDIR}/parser_cache"
+file "#{BUILDDIR}/parser_cache" => BUILDDIR
 
 file "#{BUILDDIR}/foreman.yaml" => 'config/foreman.yaml' do |t|
   cp t.prerequisites[0], t.name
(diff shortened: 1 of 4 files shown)


Rakefile
```diff
@@ -14,14 +14,23 @@ DATAROOTDIR = DATADIR = ENV['DATAROOTDIR'] || "#{PREFIX}/share"
 MANDIR = ENV['MANDIR'] || "#{DATAROOTDIR}/man"
 PKGDIR = ENV['PKGDIR'] || File.expand_path('pkg')
 
+exporter = ENV['PATH'].split(':').push(
+  '/usr/bin',
+  ENV['KAFO_EXPORTER']).find do |exporter|
```
ekohl: I don't like the reuse of exporter. Isn't dir (or something similar) a better naming?

domcleal: Yep, fixed the variable name and I've split up the line to make it more readable.

mmoll:
[test]

domcleal:
We don't have any tests in this repo, unfortunately.  It's on my TODO list...

mmoll:
Oh, I won't get a DEB package to test out of here :)

domcleal:
Yeah, unfortunately not.  I did build an RPM with this patch, you could always inspect the contents with:

```
curl https://copr-be.cloud.fedoraproject.org/results/domcleal/foreman-puppet-aio/epel-7-x86_64/00175544-foreman-installer/foreman-installer-1.12.0-0.2.develop.git.0.bf22cb5.el7.centos.noarch.rpm | rpm2cpio - | cpio -di
```

or run `rake build install PREFIX=/tmp/installer` locally with kafo+librarian-puppet installed to build it.

mmoll:
this is looking good :+1:

mmoll:
please rebase ;)

mmoll:
the required kafo version in the Gemfile should also be bumped then

domcleal:
Thanks, it looks like there's an issue loading the serialised Puppet::\* classes from the cache, which I need to look at in more detail - moving to WoC for the time being.

domcleal:
http://projects.theforeman.org/issues/14729 describes the issue, which is an interaction between kafo 0.7.3 and kafo_parsers 0.1.0 when using this parser cache feature.  I think I'll wait until kafo 0.8.0 is ready as it should make the issue obsolete.

Edit: kafo 0.8.0 plus the next version of kafo_parsers.

domcleal:
Rebased and updated to require kafo 0.8.0, which causes `puppet` to get loaded again with the newer kafo_parsers.  The next version of kafo_parsers (probably out today) will also get rid of Puppet::\* from the cache entirely too.

mmoll:
merging, as we do have all needed fixes in the packaged versions of kafo and kafo_parsers now

mmoll:
merged, thanks @domcleal!

PR #173, opened 2016-04-19
Add Travis CI test to build/install

domcleal:
Replace LOGDIR to work within the specified PREFIX, and replace
CONFIG_DIR to prevent config files from the local directory being used
in preference to those from the installation prefix.

---

This will exercise the Puppetfile as librarian-puppet's called during build, config migrations during the `--help` call, generation of man pages etc.

### DIFF
--- a/bin/foreman-installer
+++ b/bin/foreman-installer
@@ -2,11 +2,7 @@
 require 'rubygems'
 require 'kafo'
 
-if Dir.glob('config/*.yaml').any?
-  CONFIG_DIR = 'config/'
-else
-  CONFIG_DIR = '/etc/foreman-installer/scenarios.d/'
-end
+CONFIG_DIR = 'config/'
 
 # Run the install
 @result = Kafo::KafoConfigure.run
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ _build
 Puppetfile.lock
 modules/*
 .ruby-*
+Gemfile.lock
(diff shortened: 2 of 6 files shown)


bin/foreman-installer
```diff
@@ -2,11 +2,7 @@
 require 'rubygems'
 require 'kafo'
 
-if Dir.glob('config/*.yaml').any?
-  CONFIG_DIR = 'config/'
-else
-  CONFIG_DIR = '/etc/foreman-installer/scenarios.d/'
```
mmoll, line 8: will this work in the packaged version still?

domcleal, line 8: Yes, I changed the Rakefile too at https://github.com/theforeman/foreman-installer/pull/173/files/6fef25e5d3a41f46bff9794946b2c6faed6c84f5#diff-52c976fc38ed2b4e3b1192f8a8e24cffR34 to always replace this CONFIG_DIR line with the path it's installing to.

The packaged version used to still contain this conditional, so it would have:

<pre>
if Dir.glob('config/*.yaml').any?
  CONFIG_DIR = 'config/'
else
  CONFIG_DIR = '/etc/foreman-installer/scenarios.d/'
end
</pre>


Now it'll be a single `CONFIG_DIR = '/etc/foreman-installer/scenarios.d'` generated by the Rakefile.

I don't think there's a good reason to keep the `config/*.yaml` support in the packaged/installed version, and I've seen reports of it causing problems when config/*.yaml does exist.

mmoll:
perfect, thanks @domcleal!

PR #174, opened 2016-04-20
handle autosign change in puppet-puppet

mmoll:
disclaimer: I have no idea what I'm doing there :)


mmoll:
would something like that work?

``` ruby
current_autosign=answers['puppet']['autosign']
answers['puppet'].delete('autosign') unless !!current_autosign == current_autosign
```

domcleal:
Yes, that code snippet is probably preferable to keep any true/false values assigned to puppet::autosign.

Deleting and recomputing the value is probably OK, otherwise it means parsing it to split out the file mode into autosign_mode and expand `$confdir`.

mmoll:
@domcleal Yes, I thought about that, but at the end of the day, as any string can be given, there might be also a path without mode (and maybe even something that's not coming to our mind now), so there would be additional logic needed to check this and split it differently.

Also I would need to read up, how to expand `$confdir / $dir`. ;)

domcleal:
:+1: when tests pass.

domcleal:
Thanks @mmoll!

PR #175, opened 2016-04-29
refs #14788 - pull in Ubuntu/xenial cherry-picks in modules (1.11)

mmoll:
This is what would be needed to get successful installer runs on Ubuntu/xenial. For -apache and -postgresql this is what we have plus one commit each, for -mysql it's 3.6.x -> 3.7.x and one additional commit. Note that -mysql and -postgresql do not have versioned releases with the required fixes.

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -1,8 +1,9 @@
 forge 'https://forgeapi.puppetlabs.com'
 
 # Dependencies
-mod 'puppetlabs/mysql'
-mod 'puppetlabs/postgresql'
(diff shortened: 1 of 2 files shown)


domcleal:
:+1: looks OK to me.  I think it'd be good to update `develop` too with the fixes for pl-mysql/postgresql so we're able to test them in nightly too, before cutting the next 1.11 release.

domcleal:
Thanks @mmoll!

PR #177, opened 2016-05-11
Update theforeman/tftp for Ubuntu 16.04 service fix

domcleal:
The minor version introduces a couple of new parameters, but I think it's low risk.


mmoll:
merged, thanks @domcleal!

PR #178, opened 2016-05-16
Remove unmaintained ovirt_provision plugin

domcleal:
per https://github.com/theforeman/foreman-packaging/commit/60ca1687bc1588f547cfe722d4bbd99b0cf90d78

We can optionally remove the class later (https://github.com/theforeman/puppet-foreman/issues/443).


ekohl:
:+1:

mmoll:
Sad to see that unmaintained, but as there were some people already running into problem, it's better to delete it here, thanks @domcleal!

PR #179, opened 2016-05-20
Rename foreman_proxy::puppetrun* to puppet

domcleal:


ekohl:
`.delete()` returns the previous value so it's like `pop()`?

domcleal:
Yep: http://ruby-doc.org/core-1.8.7/Hash.html#method-i-delete

ekohl:
:+1:

mmoll:
merged, thanks @domcleal!

PR #180, opened 2016-05-21
Slight refactoring of hostname check

petems:
- Add hostnamectl for systemd machines
- Add note for Debian
- A few grammar tweaks

### DIFF
--- a/checks/hostname.rb
+++ b/checks/hostname.rb
@@ -1,15 +1,17 @@
 #!/usr/bin/env ruby
 
-BASE = %q(If needed, change the hostname permanently via 'hostname' command and editing 
-appropriate configuration file.
-(e.g. on Red Hat systems /etc/sysconfig/network).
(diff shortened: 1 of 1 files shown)


mmoll:
I'm more used to a 120 char limit per line, :+1: otherwise.

As a side note, on my Debian systems, /etc/hostname does contain only the hostname, not the FQDN. I have no idea about the respective file on RH based systems and what would happen, if the FQDN is put there, but that might be worth a clarification.

ekohl:
> As a side note, on my Debian systems, /etc/hostname does contain only the hostname, not the FQDN. I have no idea about the respective file on RH based systems and what would happen, if the FQDN is put there, but that might be worth a clarification.

For RH-systems it's common to put the FQDN in /etc/sysconfig/network

domcleal:
Merged, thanks @petems.

PR #183, opened 2016-06-07
Change foreman_proxy 'puppetssh' provider value to 'ssh'

domcleal:
For 1.12-stable as well please.


config/foreman.migrations/20160607133050_foreman_proxy_puppetssh_rename.rb
```diff
@@ -0,0 +1,3 @@
+# Rename foreman_proxy provider "puppetssh" to "ssh"
+# http://projects.theforeman.org/issues/15323
+answers['foreman_proxy']['puppetrun_provider'] = 'ssh' if answers['foreman_proxy'] && answers['foreman_proxy']['puppetrun_provider'] == 'puppetssh'
```
ekohl, line 3: Any chance `answers['foreman_proxy']['puppetrun_provider']` isn't set?

domcleal, line 3: Yes, it'll be unset with the default answers file, but it'll still work as `Hash#[]` returns `nil`, and `nil` can safely be compared.

mmoll:
merged, thanks @domcleal!
Also cherry-picked as e14da2b for 1.12.

PR #184, opened 2016-06-07
Fix foreman_proxy answers migration when module's disabled

domcleal:
When the foreman_proxy module is disabled, the value of the answer will
be false rather than a hash, causing a migration error:

```
20160518125928_foreman_proxy_puppetrun_rename.rb:3:in `block
(2 levels) in load_migrations': undefined method `has_key?' for
false:FalseClass (NoMethodError)
```

Also for 1.12-stable please, as that's where the migration was introduced.


ekohl:
:+1:

mmoll:
also cherry-picked as 561e8bc into 1.12, thanks @domcleal!

PR #185, opened 2016-06-09
fixes #14789 - disable apache::mod::passenger::manage_repo

domcleal:
Add Hiera configuration using Kafo 0.9.0's hiera_config option to set
the Passenger class' manage_repo parameter to false on "RedHat" OSes,
preferring the EPEL or Foreman packages.

Hiera data files are installed as data files rather than configuration
files, as they're shipped by the installer only to configure modules
rather than for the user to configure the installer.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'kafo', '>= 0.8.0'
+gem 'kafo', '>= 0.9.0'
 gem 'librarian-puppet'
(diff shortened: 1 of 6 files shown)


ekohl:
:+1: this is very cool!

beav:
:+1:, altering `apache::mod::passenger::manage_repo` and running a katello install laid down the expected EPEL version of passenger for me.

mmoll:
Very nice, thanks @domcleal!
I'll have a look now, how to use this to solve that puppetlabs-puppetdb problem...

PR #187, opened 2016-06-14
add foreman plugin foreman_expire_hosts

timogoebel:
@domcleal : As discussed on https://github.com/theforeman/puppet-foreman/pull/454 . Would you mind taking a close look into this? I've never even started foreman-installer, so this is just a lucky guess.

### DIFF
--- a/config/foreman-answers.yaml
+++ b/config/foreman-answers.yaml
@@ -25,6 +25,7 @@ foreman::plugin::dhcp_browser: false
 foreman::plugin::digitalocean: false
 foreman::plugin::discovery: false
 foreman::plugin::docker: false
+foreman::plugin::expire_hosts: false
 foreman::plugin::memcache: false
 foreman::plugin::openscap: false
 foreman::plugin::tasks: false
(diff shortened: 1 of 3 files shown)


domcleal:
Could you also add a migration file like this one? https://github.com/theforeman/foreman-installer/blob/develop/config/foreman.migrations/20160405135043_foreman_proxy_discovery.rb, but the second line should use a symbol for the second hash, e.g. `scenario[:mapping][:'foreman::plugin::expire_hosts']` (it was a bug.)

timogoebel:
@domcleal : Added the migration and finally found the documentation for this.

config/foreman.migrations/160614145332_foreman_expire_hosts.rb
```diff
@@ -0,0 +1,2 @@
+answers['foreman::plugin::expire_hosts'] ||= false
```
domcleal: This file should be named 20160614145332... so it orders correctly with the existing files.

timogoebel: @domcleal: Sorry for this, i just copied the command from the kafo docs. Pushed a new commit.

Is `date +%y%m%d%H%M%S` the correct command on your machine?
I need `date +%Y%m%d%H%M%S` both on OS X and RHEL...

domcleal: %Y's needed, I didn't notice the Kafo recommendation until after I added migrations here.  I used the full year to match Rails migrations and most date formats, and think I still prefer it.

domcleal:
Merged, thanks @timogoebel!

PR #188, opened 2016-06-15
Add foreman_azure to config

dLobatog:
Depends on https://github.com/theforeman/puppet-foreman/pull/456


domcleal:
Please also add a new config file migration, see ff04cd1 for an example.

config/foreman.migrations/20160615132930_foreman_azure.rb
```diff
@@ -0,0 +1,2 @@
++answers['foreman::plugin::azure'] ||= false
```
domcleal: There's a leading `+` on these lines causing the migration (and tests) to fail.

dLobatog:
Updated to fix the typo on the migration

domcleal:
Merged, thanks @dLobatog!

PR #191, opened 2016-07-22
rename foreman environment parameter to rails_env

igalic:
ref: https://github.com/theforeman/puppet-foreman_proxy/commit/d239f0b


domcleal:
Merged, thanks for doing that @igalic.

PR #196, opened 2016-09-02
unpin puppetlabs/puppetdb

mmoll:
See theforeman/puppet-puppet#361.

This should be mentioned in the release notes as this makes the PuppetDB integration Puppet 4 AIO only.


domcleal:
Merged, thanks @mmoll.

PR #197, opened 2016-09-05
Refs #12635 - added migration for grub tftp dirs

lzap:
Required change for https://github.com/theforeman/puppet-foreman_proxy/pull/276

### DIFF
--- a/config/foreman.migrations/20160831180000_foreman_proxy_tftp_grub2.rb
+++ b/config/foreman.migrations/20160831180000_foreman_proxy_tftp_grub2.rb
@@ -1,0 +1,5 @@
+root = answers['foreman_proxy']['tftp_root']
+dirs = answers['foreman_proxy']['tftp_dirs']
+dirs << "#{root}/grub"
+dirs << "#{root}/grub2"
+dirs.uniq!
--- a/config/foreman.migrations/20160831180000_foreman_proxy_tftp_grub2.rb
+++ b/config/foreman.migrations/20160831180000_foreman_proxy_tftp_grub2.rb
@@ -1,5 +1,9 @@
-root = answers['foreman_proxy']['tftp_root']
-dirs = answers['foreman_proxy']['tftp_dirs']
-dirs << "#{root}/grub"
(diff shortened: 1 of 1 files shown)


mmoll:
this seems to fail

config/foreman.migrations/20160831180000_foreman_proxy_tftp_grub2.rb
```diff
@@ -0,0 +1,5 @@
+root = answers['foreman_proxy']['tftp_root']
```
domcleal: `answers['foreman_proxy']` may be false if the module is disabled

config/foreman.migrations/20160831180000_foreman_proxy_tftp_grub2.rb
```diff
@@ -0,0 +1,5 @@
...
+dirs = answers['foreman_proxy']['tftp_dirs']
+dirs << "#{root}/grub"
```
domcleal: `dirs` may be nil on the first run, not necessarily an Array

lzap:
Damn, I am unable to get to my work tree this week. Upgrades will be likely broken for RC1 then, will fix this for RC2 as my top priority.

config/foreman.migrations/20160831180000_foreman_proxy_tftp_grub2.rb
lzap: Nil on the first run? Why is that?

lzap:
I made a change using Github editor. @mbacovsky we talked about this in Brno last week, is this correct?

config/foreman.migrations/20160831180000_foreman_proxy_tftp_grub2.rb
domcleal: The first run will be using the default answers file which doesn't contain much: https://github.com/theforeman/foreman-installer/blob/develop/config/foreman-answers.yaml#L14

domcleal:
Merged, thanks @lzap.

PR #198, opened 2016-09-06
handle API whitelist changes in puppet-puppet

mmoll:
see theforeman/puppet-puppet#428


config/foreman.migrations/20160906121542_puppetserver_whitelist.rb
```diff
@@ -0,0 +1,8 @@
+# Redetermine the value of API whitelists, as it changed
+# in puppet-puppet XXX
```
mmoll, line 2: actual commit ID will be here

domcleal: f9b4e87cd855d7d5d0bbf3a1831b5daf22cdb413

domcleal:
Merged with the commit ID added, thanks @mmoll.

PR #201, opened 2016-09-19
Add migration for infoblox plugins

iNecas:


ekohl (approved):

domcleal (changes requested):
I think these should also be listed in the default config files, config/foreman.yaml and foreman-answers.yaml.

iNecas:
Updated

domcleal (approved):

domcleal:
Merged, thanks @iNecas.

PR #203, opened 2016-10-18
Added foreman hooks check

lzap:
Added check:

There are foreman hooks installed on this instance, please either remove them
or make them non-executable in order to continue. Foreman hooks often change
behavior and are not tested across releases.

When upgrading, test all workflows before re-enabling any of these hooks to
verify clean installation. Remember, it can be required to remove all Foreman
hooks completely when requesting support.


ekohl:
What's the reasoning behind this?

sean797:
:-1: for this. I don't see a reason to stop users from upgrading if they have hooks installed, a note in the docs saying foreman_hooks scripts behavior may change after an upgrade should suffice imo

lzap:
The reason is UEFI workaround which break after Satellite 6.2 upgrade (jgrep). You can put it ten times in the docs, users won't notice it.

Anyway, I am fine changing this to warning only.

ekohl:
I'd prefer a warning that tells users to verify that they check their hooks. Since they can have custom hooks that they wrote we can't test them so it makes perfect sense that we can't guarantee they continue working. The current text is too strong though.

ohadlevy:
Can't we disable hooks while updating instead?

‏בתאריך יום רביעי, 19 באוקטובר 2016, Ewoud Kohl van Wijngaarden <
notifications@github.com> כתב:

ekohl:
> Can't we disable hooks while updating instead?

I honestly don't know what's the best solution here. Some users might expect that we respect their installation while others would like their hand to be held and have a safe upgrade. There may not be a silver bullet here.

That doesn't mean I can't have a preference ;) Personally I'm in the camp of the 'respect my install' hence why I think we should only present a warning. We should never just disable hooks. This will surprise users which is always a bad thing.

A compromise could be a parameter that users have to present if there are hooks installed. The installer would exit if hooks are active and the parameter is not passed.

sean797:
I think that someone who has written a hook should know that something may change during a upgrade, though displaying a message saying hooks may or may not break after an upgrade if they are using hooks wouldn't be a bad thing, but the upgrade should continue.

Disabling something because you cannot guarantee that it will work afterwards is not the solution. Would you be annoyed if #your-favorite-distro disabled the proprietary video card drivers every time you ran the equivalent of `dnf update -y` because they cannot guarantee that it will still work? I would, not that I run an proprietary video card drivers ;)

lzap:
The video card driver argument is not relevant here, we are not doing "yum update" but distribution upgrade. (And in case of Fedora, the upgrade process does disable thirdparty repos and documentation recommends to uninstall 3rd party software.) But anyway, Foreman with plugins is a very complex system and providing support can be challenging. Foreman hooks and plugins are things that can change behavior. This patch aims to solve two possible issues:

1) Foreman hooks breaking Foreman _during_ upgrade.

2) Foreman hooks breaking Foreman _after_ upgrade.

I think it's perfectly clean and valid to ask users to disable hooks, it's just one `chmod` command. But let's review opinions here:

A) Don't let user to carry on with installation until hooks are disabled (current proposal)

B) A but with possibility to skip this check (I like this idea very much)

C) Disable hooks automatically and keep them disabled informing user here and in the documentation to re-enable them

D) Disable hooks for the upgrade process, re-enable them afterwards.

E) Issue just warning message

I think disabling them is just too much, and nobody reads warning messages or documentation.(TM)

Therefore my opinion is to don't let user to carry on with opt-out with parameter. This can be clearly communicated in the error message.

ekohl:
I think @lzap perfectly summarized the options and we're fully in agreement on the conclusion.

domcleal:
> 1) Foreman hooks breaking Foreman during upgrade.

Upgrades happen prior to the user running foreman-installer, suggesting that whatever bug is caused during an upgrade should be fixed in foreman_hooks.

> 2) Foreman hooks breaking Foreman after upgrade.

Foreman installer is pretty unrelated to the foreman_hooks plugin and is not only used after upgrades, I don't think this is an appropriate place to warn users on each installer invocation. Upgrade documentation would be best.

lzap:
Added

```
Use -s or --skip-checks-i-know-better parameter to ignore the check.
```

to the message. Created a patch in kafo in order to execute hooks in sorted way: https://github.com/theforeman/kafo/pull/186

lzap:
> Upgrades happen prior to the user running foreman-installer, suggesting that whatever bug is caused during an upgrade should be fixed in foreman_hooks.

I am not trying to prevent against bugs in foreman_hooks itself, rather than against bugs introduced by hooks. Consider a hook that rolls out every single transaction, which can happen since during upgrade there are new versions of packages, dependencies, sometimes new ruby (SCL change). Then migrations can be problem.

> Foreman installer is pretty unrelated to the foreman_hooks plugin and is not only used after upgrades, I don't think this is an appropriate place to warn users on each installer invocation. Upgrade documentation would be best.

There is no doubt about documentation, but this is the installer. It's related to everything it installs or upgrade.

ekohl:
In [katello-installer](https://github.com/Katello/katello-installer/tree/master/hooks/pre) there are precedents on these sort of checks. See [17-memory_check.rb](https://github.com/Katello/katello-installer/blob/master/hooks/pre/17-memory_check.rb). I now wonder if those should be rewritten to checks.

domcleal:
> > Upgrades happen prior to the user running foreman-installer, suggesting that whatever bug is caused during an upgrade should be fixed in foreman_hooks.
> 
> I am not trying to prevent against bugs in foreman_hooks itself, rather than against bugs introduced by hooks. Consider a hook that rolls out every single transaction, which can happen since during upgrade there are new versions of packages, dependencies, sometimes new ruby (SCL change). Then migrations can be problem.

Adding a warning to the installer will not help during upgrade then, the upgrade will fail _prior_ to the user seeing this message. I suggest you fix this in foreman_hooks by disabling hooks during a DB migration if that's where it fails, or fix the DB migrations to not use the hooked models.

> >  Foreman installer is pretty unrelated to the foreman_hooks plugin and is not only used after upgrades, I don't think this is an appropriate place to warn users on each installer invocation. Upgrade documentation would be best.
> 
> There is no doubt about documentation, but this is the installer. It's related to everything it installs or upgrade.

My point though was that the installer should not fail or warn users on _every invocation_ that their common configuration (having this plugin installed and configured) is a problem. The installer may be used for other things than being run after an upgrade has taken place.

Relying on the user to see upgrade warnings from plugins by running the installer is also a bit backwards. If there's a warning associated with the plugin on upgrade, the plugin should be implementing it, it shouldn't be in this project.

lzap:
Okay, sure that's valid I misunderstood. I will take a look at `hooks` if I am able to write hook that will only trigger during upgrade to show the warning.

lzap:
> If there's a warning associated with the plugin on upgrade, the plugin should be implementing it, it shouldn't be in this project.

Do you have something in mind how to implement this? Is there an API to do that?

domcleal:
> > If there's a warning associated with the plugin on upgrade, the plugin should be implementing it, it shouldn't be in this project.
> 
> Do you have something in mind how to implement this? Is there an API to do that?

No, I don't.

PR #206, opened 2016-11-01
refs #16207 - move foreman-hiera.conf out of _build/config

domcleal:
Reverts the change I made in 9b69685 to building foreman-hiera.conf, which was being skipped as I accidentally removed it from the `build` target and it was also already present at `#{BUILDDIR}/config/foreman-hiera.conf` because `config/` was entirely copied to `_build/config`.

Fixes test failures seen in nightly installer packages where foreman-hiera.conf refers to `./config/foreman.hiera` instead of the full path.


mmoll:
thanks!

PR #211, opened 2016-11-23
Remove class mappings, now correct by default

domcleal:


ekohl (approved):
Cool! I wonder if we should remove mapping migrations since they end up getting cleared anyway but probably safe to leave them.

mmoll:
merged, thanks @domcleal!

PR #212, opened 2016-12-02
Fixes #16298 - Add a user is root check to the front  of installer

bkearney:
…ler command

### DIFF
--- a/bin/foreman-installer
+++ b/bin/foreman-installer
@@ -4,6 +4,13 @@ require 'kafo'
 
 CONFIG_DIR = 'config/'
 
+# Verify that this is run as root
+if Process.uid != 0
+  cmd_name = File.basename($0)
+  puts '%s must be run as root' % cmd_name
+  exit -1
+end
+
 # Run the install
 @result = Kafo::KafoConfigure.run
 


bkearney:
Ticket was moved.

ekohl (approved):
Makes sense to me

ekohl (changes requested):
The tests fail because --help isn't executed as root. Maybe there should be an exception for that?

ares:
I think it would be better to do the check in a hook to avoid parsing ARGV here.

bkearney:
@ekohl what tests are failing? I ran rake spec and all seemed good.

bkearney:
@ares the issue crops up when the user tries to run help, which needs to get the scenario files, which are only available to root users. So, doing the root check in a hook would be done after that step.

ekohl:
@bkearney The tests fails because on Travis we [run with ```--help```](https://github.com/theforeman/foreman-installer/blob/25ceccd02236a069df98e74a31a63df37bfb695f/.travis.yml#L24). This works on Travis because it all works on a temporary directory owned by the correct user, but tests aren't executed as root. The problem is that foreman-installer will unconditionally write the answers file, even when running ```--help```. The correct check would be if the answers file location is writeable.

ares:
@bkearney oh I see, thanks

ekohl:
@ares here's the error:

```
[vagrant@c7-fm-bind ~]$ foreman-installer --help
/usr/share/ruby/fileutils.rb:1142:in `utime': Permission denied - /etc/foreman-installer/scenarios.d/foreman.yaml (Errno::EACCES)
	from /usr/share/ruby/fileutils.rb:1142:in `block in touch'
	from /usr/share/ruby/fileutils.rb:1139:in `each'
	from /usr/share/ruby/fileutils.rb:1139:in `touch'
	from /usr/share/gems/gems/kafo-1.0.5/lib/kafo/configuration.rb:54:in `save_configuration'
	from /usr/share/gems/gems/kafo-1.0.5/lib/kafo/configuration.rb:61:in `configure_application'
	from /usr/share/gems/gems/kafo-1.0.5/lib/kafo/configuration.rb:38:in `initialize'
	from /usr/share/gems/gems/kafo-1.0.5/lib/kafo/scenario_manager.rb:214:in `new'
	from /usr/share/gems/gems/kafo-1.0.5/lib/kafo/scenario_manager.rb:214:in `load_configuration'
	from /usr/share/gems/gems/kafo-1.0.5/lib/kafo/scenario_manager.rb:143:in `scenario_enabled?'
	from /usr/share/gems/gems/kafo-1.0.5/lib/kafo/scenario_manager.rb:97:in `select_scenario'
	from /usr/share/gems/gems/kafo-1.0.5/lib/kafo/kafo_configure.rb:470:in `config_file'
	from /usr/share/gems/gems/kafo-1.0.5/lib/kafo/kafo_configure.rb:60:in `initialize'
	from /usr/share/gems/gems/clamp-1.0.0/lib/clamp/command.rb:133:in `new'
	from /usr/share/gems/gems/clamp-1.0.0/lib/clamp/command.rb:133:in `run'
	from /usr/share/gems/gems/kafo-1.0.5/lib/kafo/kafo_configure.rb:154:in `run'
	from /usr/sbin/foreman-installer:8:in `<main>'
```

bkearney:
I am closing this in favor of https://github.com/theforeman/kafo/pull/205

PR #218, opened 2017-01-18
fixes #18113 - change dhcp_range from false to undef

domcleal:
The module permits `Optional[String]` for this parameter, so it's
preferred to use undef/nil instead of false to disable it.


mmoll (approved):

ekohl:
Thanks!

PR #219, opened 2017-02-07
Add upgrade-puppet hook

ekohl:
Based on the upgrade-puppet hook in katello-installer.

### DIFF
--- a/hooks/boot/01-helpers.rb
+++ b/hooks/boot/01-helpers.rb
@@ -1,0 +1,44 @@
+class Kafo::Helpers
+  class << self
+    def module_enabled?(kafo, name)
+      mod = kafo.module(name)
+      return false if mod.nil?
+      mod.enabled?
+    end
+
+    def log_and_say(level, message)
+      style = case level
+              when :error
+                'bad'
+              when :debug
+                'yellow'
+              else
+                level
+              end
+
+      # \ and ' characters could cause trouble in ERB, make sure to escape them
+      escaped_message = message.gsub('\\', '\\\\\\').gsub("'", %q{\\\'})
+      say "<%= color('#{escaped_message}', :#{style}) %>"
+      Kafo::KafoConfigure.logger.send(level, message)
+    end
+
+    def execute(commands)
+      commands = commands.is_a?(Array) ? commands : [commands]
+      results = []
+
+      commands.each do |command|
+        output = `#{command} 2>&1`
+
+        if $?.success?
+          log_and_say(:debug, output.to_s)
+          results << true
+        else
+          log_and_say(:error, output.to_s)
+          results << false
+        end
+      end
+
+      !results.include? false
+    end
+  end
+end
--- a/hooks/boot/31-upgrade-puppet.rb
+++ b/hooks/boot/31-upgrade-puppet.rb
@@ -1,0 +1,6 @@
+app_option(
+  '--upgrade-puppet',
+  :flag,
+  "Run the steps neccessary to upgrade from Puppet 3 to Puppet 4.",
+  :default => false
+)
--- a/hooks/pre/31-upgrade-puppet.rb
+++ b/hooks/pre/31-upgrade-puppet.rb
@@ -1,0 +1,96 @@
+def puppet4_available?
+  osfamily = %x{facter osfamily}.strip
+  success = case osfamily
+  when 'RedHat'
+    ['puppet-agent', 'puppetserver'].map do |pkg|
+      %x{yum info #{pkg} > /dev/null}
+      $?.success?
+    end
+  when 'Debian'
+    ['puppet-agent', 'puppetserver'].map do |pkg|
+      %x{apt-cache show #{pkg} > /dev/null}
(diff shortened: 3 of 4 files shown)


ekohl:
So this can be considered a preview. The helper was copied verbatim from katello-installer. I'm not sure if we need/want all of it but this allows me to quickly show a PoC. Currently it's broken but I'm interested to get some initial feedback.

ekohl:
Verified to work on CentOS 7 now.

hooks/pre/31-upgrade-puppet.rb
```diff
@@ -0,0 +1,89 @@
+def puppet4_available?
+  success = []
+  # TODO: OS dependent + override
```
mmoll: what's the status here?

mmoll:
looking good visually, but I'm unsure if I really can test drive this properly for 1.15

ekohl:
Updated, now also tested with Debian Jessie and stripped some unused code.

Regarding inclusion in 1.15: what would give you sufficient confidence?

mmoll:
As it is a new feature that's not interferring with the default install, I'd even merge it when you say, you tested it sucessfully under the OSes we support. We can ask people in the RC announcements to test and fix things on the way, if needed....

mmoll:
@domcleal any opinion on this PR?

hooks/boot/31-upgrade-puppet.rb
```diff
@@ -0,0 +1,6 @@
+app_option(
+  '--upgrade-puppet',
```
domcleal, line 2: This should contain the target version number (4), so as to leave room for 5 and future versions.

hooks/boot/01-helpers.rb
```diff
@@ -0,0 +1,44 @@
+class Kafo::Helpers
```
domcleal, line 1: It shouldn't be extending Kafo's modules from here, it should use its own namespace.

hooks/boot/01-helpers.rb
```diff
@@ -0,0 +1,44 @@
...
+  class << self
+    def module_enabled?(kafo, name)
```
domcleal, line 3: Already provided in hooks: https://github.com/theforeman/kafo#hooks

hooks/boot/01-helpers.rb
```diff
@@ -0,0 +1,44 @@
...
+              when :debug
+                'yellow'
+              else
+                level
+              end
+
+      # \ and ' characters could cause trouble in ERB, make sure to escape them
+      escaped_message = message.gsub('\\', '\\\\\\').gsub("'", %q{\\\'})
+      say "<%= color('#{escaped_message}', :#{style}) %>"
+      Kafo::KafoConfigure.logger.send(level, message)
```
domcleal, line 22: Similarly, `logger` should be in the hook context

hooks/pre/31-upgrade-puppet.rb
```diff
@@ -0,0 +1,96 @@
...
+    'puppet'        => ['autosign', 'client_package', 'codedir', 'configtimeout', 'dir', 'logdir', 'rundir', 'ssldir', 'vardir', 'server_common_modules_path', 'server_default_manifest_path', 'server_dir', 'server_envs_dir', 'server_external_nodes', 'server_jruby_gem_home', 'server_package', 'server_puppetserver_dir', 'server_puppetserver_vardir', 'server_ruby_load_paths', 'server_ssl_dir'],
+    'foreman_proxy' => ['puppet_ssl_ca', 'puppet_ssl_cert', 'puppet_ssl_key', 'puppetca_cmd', 'puppetdir', 'ssl_ca', 'ssl_cert', 'ssl_key', 'ssldir'],
+  }
+
+  unless app_value(:noop)
+    params.each do |mod, names|
+      names.each do |name|
+        p = param(mod, name)
+        fail_and_exit "Failed to find parameter '#{name}' in '#{mod}'" if p.nil?
+        p.value = nil
```
domcleal, line 62: Should use `.unset_value`.

domcleal (changes requested):

domcleal:
> @domcleal any opinion on this PR?

I'm unsure if this is safe enough to use and ship with the assumptions it makes about the user's environment, so aside from the technical points above, I'm not inclined to review/merge it.

hooks/boot/01-helpers.rb
ekohl: I tried that, but didn't work

hooks/pre/31-upgrade-puppet.rb
ekohl: When I tried that it didn't work. It would revert back to the value that was set before.

mmoll:
I'll need more time dive into this, but IMHO the assumptions that are made are correct for almost any Puppet 3 RPM and DEB install?

ekohl:
Closing this since Puppet 3 -> 4 is no longer relevant. Will revisit this when merging katello-installer into foreman-installer.

PR #220, opened 2017-02-13
Enable foreman_proxy::realm_split_config_files for 1.15

domcleal:
After https://github.com/theforeman/puppet-foreman_proxy/pull/327.


ekohl:
I'm guessing the differing defaults (true here vs false in the module) are intentional

domcleal:
Yes, as the `develop` branch here is specifically to be used with version 1.15 (same as f777323).

The module default could be true for 1.15 too, but I think we've been too aggressive in defaulting parameters to support unreleased versions, so it's false.

ekohl (approved):
I had the same thought process

mmoll:
merged, thanks @domcleal!

PR #221, opened 2017-03-09
Fixes #18821 - set PassengerMaxRequests to keep processes healthy

lzap:
I have *no idea* how this should work, but the goal is to set
PassengerMaxRequests for all passenger processes.


mmoll (approved):

ekohl:
Thanks!

PR #222, opened 2017-03-17
Add dynflow_in_core to deploy dynflow executor

dLobatog:
puppet-foreman supports a parameter dynflow_in_core that tells the module 
whether to run the foreman-tasks service when foreman is installed (1.15+) or
when foreman-tasks is installed (1.14)

Depends on https://github.com/theforeman/puppet-foreman/pull/530

### DIFF
--- a/config/foreman-answers.yaml
+++ b/config/foreman-answers.yaml
@@ -12,6 +12,7 @@
 foreman:
   custom_repo: true
   email_config_method: database
+  dynflow_in_core: true
 foreman_proxy:
   custom_repo: true
 puppet:


config/foreman-answers.yaml
```diff
@@ -12,6 +12,7 @@
 foreman:
   custom_repo: true
   email_config_method: database
+  dynflow_in_core: true
```
ekohl, line 15: Given this is the default, is this actually needed?

dLobatog: Actually only for 1.14

mmoll: The whole dynflow_executor change won't end up in any component used by 1.14 releases, or am I getting something completely wrong?

ekohl: @dLobatog the current puppet code defaults to true in params.pp and like @mmoll pointed out: this won't end up in 1.14 so I don't think this is actually needed.

ekohl:
I think we can close this. The puppet PR only adds params with correct defaults which makes this redundant.

ekohl:
Going to close this since it should be redundant.

PR #225, opened 2017-04-03

mmoll:


mmoll:
This time I also restricted the versions of the puppetlabs modules, because this will be the last composition that will be Puppet 3 compatible.

domcleal (approved):

ekohl (approved):

PR #226, opened 2017-04-04

dLobatog:
Note this is exclusively for 1.15-stable


domcleal:
This conflicts with a939706.

dLobatog:
Correct - I was looking at my old fork and did not pull, this is unneeded. Thanks for the quick reply!

PR #229, opened 2017-04-06
Test foreman-installer under Puppet 4

domcleal:
Additionally requires the puppet-strings gem as a parser under 4, which
is a weak dependency of kafo_parsers.


ekohl (approved):

mmoll:
merged, thanks @domcleal!

PR #230, opened 2017-04-20
Add logic to check executable of facter command

NewpTone:
In some situation, facter is not installed or not exist in the system executable command path,  then 'facter fqdn' will fail and exit, this patch try to catch this exception.

### DIFF
--- a/checks/hostname.rb
+++ b/checks/hostname.rb
@@ -21,7 +21,10 @@ INVALID = %q(Output of 'hostname -f' does not seems to be valid FQDN
 
 Make sure above command gives fully qualified domain name. At least one
 dot must be present. )
- 
+
+MISSING = %q(Command 'facter' does not exist in system executable path
+
+Make sure the above command is installed and executable in your system. )
 
 def error_exit(message, code)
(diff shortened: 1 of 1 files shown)


checks/hostname.rb
```diff
@@ -21,14 +21,21 @@
 
 Make sure above command gives fully qualified domain name. At least one
 dot must be present. )
- 
+
+MISSING = %q(Command 'facter' does not exist in system excutable path
```
domcleal: 'executable' typo

checks/hostname.rb
```diff
@@ -21,14 +21,21 @@
...
+
+Make sure above command is installed and executable in your system. )
```
domcleal: "Make sure _the_ above command"

domcleal (changes requested):

mmoll (approved):

domcleal (approved):

domcleal:
Merged, thanks @NewpTone.

(It's a good idea to reply to comments so we know it's updated.)

PR #231, opened 2017-05-30
Refs #6945 - use autosign path for puppetca

mmoll:


mmoll:
to be merged after theforeman/puppet-foreman_proxy#350

config/foreman.migrations/20170530141108_foreman_proxy_autosignfile.rb
```diff
@@ -0,0 +1,7 @@
+if answers['foreman_proxy']
+  answers['foreman_proxy']['use_autosignfile'] = true
```
ekohl: This shouldn't happen unconditionally I think, only if it's nil.

config/foreman.migrations/20170530141108_foreman_proxy_autosignfile.rb
```diff
@@ -0,0 +1,7 @@
...
+  if answers['foreman_proxy'].has_key?('puppetdir')
+    puppetdir = answers['foreman_proxy']['puppetdir']
```
ekohl: Delete the answer as well?

ekohl (changes requested):

config/foreman.migrations/20170530141108_foreman_proxy_autosignfile.rb
mmoll: changed.

config/foreman.migrations/20170530141108_foreman_proxy_autosignfile.rb
mmoll: $puppetdir still used in `templates/puppet_proxy_legacy.yml.erb`

ekohl (approved):

mmoll:
added a slash before 'autosign.conf'

domcleal:
Thanks @mmoll.

PR #232, opened 2017-06-26
Pin postgres and concat

ehelms:
Should be unpinned once puppetlabs/puppetdb dependency on postgres
is updated.

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -2,7 +2,8 @@ forge 'https://forgeapi.puppetlabs.com'
 
 # Dependencies
 mod 'puppetlabs/mysql',         '>= 3.8.0'
-mod 'puppetlabs/postgresql',    '>= 4.8.0'
+mod 'puppetlabs/postgresql',    '>= 4.8.0', '< 5.0.0'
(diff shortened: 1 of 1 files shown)


Puppetfile
```diff
@@ -2,7 +2,8 @@ forge 'https://forgeapi.puppetlabs.com'
 
 # Dependencies
 mod 'puppetlabs/mysql',         '>= 3.8.0'
-mod 'puppetlabs/postgresql',    '>= 4.8.0'
+mod 'puppetlabs/postgresql',    '< 5.0.0'
```
mmoll: `mod 'puppetlabs/postgresql',    '>= 4.8.0', '< 5.0.0'` would be more correct

mmoll (changes requested):

ehelms:
@mmoll Good call -- updated

mmoll (approved):

mmoll:
merged, thanks @ehelms!

ekohl:
When https://github.com/puppetlabs/puppetlabs-puppetdb/pull/260 is merged and released I think we can remove the pinning again.

PR #233, opened 2017-07-06
remove pinning of puppetlabs/postgresql to <5

mmoll:
only puppetlabs/concat is currently a problem for librarian-puppet


mmoll:
As puppetlabs-puppetdb 6.x is out now, it turns out, puppetlabs-concat is pinned to <3 by puppetlabs-apache and libraian-puppet is running into problems with it, still.

However, puppetlabs-postgresql 5.x is doing fine.

ekohl (approved):

PR #234, opened 2017-07-06
use pl-mysql 3.11.x with a fix for Debian/stretch

mmoll:
This is 3.11.0 with one commit cherry-picked, as it seems there will be no puppetlabs-mysql release in the near future. This should fix the current systest_foreman_mysql error on stretch.


ekohl:
Thanks!

PR #235, opened 2017-07-07
Refs #20194 - add migration for JRuby profiler

mmoll:


config/foreman.migrations/20170708010605_puppet_profiler_metrics.rb
```diff
@@ -0,0 +1,8 @@
+# server_puppetserver_metrics also controls if the Ruby profiler gets enabled
+# in puppet-puppet since 94ff77740a27d458ce1444db016645ab763cba42
+if answers['puppet'] && answers['puppet'].has_key?('server_enable_ruby_profiler')
+  if answers['puppet']['server_enable_ruby_profiler'] == true
```
mmoll, line 4: the default was always `false`, so only set it, when it's true (and I think with " == true" it's more readable)

ekohl:
Thanks!

PR #237, opened 2017-07-17
Refs #20282 - add more ciphers

mmoll:
The JRE coming with Debian/jessie does expose problems with Puppetserver,
delivering reports to Foreman with this configuration.


mmoll:
This will be reverted after Debian/jessie support got dropped.

mmoll:
@tbrisker ☝️

ekohl:
Interesting solution. I think this will work in most environments but it does make me wonder how many more old clients there are. Say a user deploys foreman on RH but puppet servers on Debian Jessie then there will still be a problem. Maybe for now we should just make it a recommendation?

mmoll:
> a user deploys foreman on RH but puppet servers on Debian Jessie then there will still be a problem.

hm, true... reverting that ciphers now...

mmoll:
added the cipher, that's used when setting no options

tbrisker:
@mmoll looks like the list is missing a few, apparently my copy+paste of the string got cut off in the middle. This should be:
`ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA256:DHE-RSA-AES256-SHA:ECDHE-ECDSA-DES-CBC3-SHA:ECDHE-RSA-DES-CBC3-SHA:EDH-RSA-DES-CBC3-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:DES-CBC3-SHA:!DSS`
could you please change to the complete list?

mmoll:
aha :)

tbrisker (approved):

ekohl:
Thanks!

PR #238, opened 2017-07-28
unpin puppetlabs/concat again

mmoll:
this is possible with puppetabs-apache 2.0.0 now


ekohl (approved):
This may break katello-installer but I'll deal with that later.

PR #241, opened 2017-09-11
Fixes #20889 - Use the Apache MPM event worker

ekohl:


mmoll:
works for me and is even the recomendation from the passenger guys.

merged, bedankt @ekohl!

PR #242, opened 2017-09-12
[WIP] Add checks on the system based on the scenario

ekohl:
Currently katello installer has a check_memory hook[1] that isn't very generic. This allows us to replace that hook with generic code and overrides in the scenario.

[1]: https://github.com/Katello/katello-installer/blob/a0a897dd62b5e0b57f105d073569b66a21564fe4/hooks/pre/17-memory_check.rb

### DIFF
--- a/hooks/pre/05-check_scenario_requirements.rb
+++ b/hooks/pre/05-check_scenario_requirements.rb
@@ -1,0 +1,76 @@
+require 'pathname'
+
+PREFIX = %W(PiB TiB GiB MiB KiB B).freeze
+
+# From https://codereview.stackexchange.com/questions/9107/printing-human-readable-number-of-bytes
+def as_size(s)
+  s = s.to_f
+  i = PREFIX.length - 1
+  while s > 512 && i > 0
+    i -= 1
+    s /= 1024
+  end
+  ((s > 9 || s.modulo(1) < 0.1 ? '%d' : '%.1f') % s) + ' ' + PREFIX[i]
+end
+
+def available_space(directory)
+  mountpoints = facts[:mountpoints]
+  until (mountpoint = mountpoints[directory.to_sym])
+    directory = File.dirname(directory)
+  end
+  mountpoint[:available_bytes]
+end
+
+def scenario_requirements
+  scenario_data[:required] || {}
+end
+
+def ram
+  facts[:memory][:system][:total_bytes]
+end
+
+def cores
+  facts[:processors][:count]
+end
+
+def check
+  requirements = scenario_requirements
+
+  if requirements.nil? || requirements.empty?
+    logger.debug 'No requirements for this scenario'
+    return
+  end
+
+  if facts.empty?
+    raise Exception, 'Failed to load facts'
+  end
+
+  if (min_ram = requirements['memory']) && ram < (min_ram * 0.9)
+    raise Exception, "This system has #{as_size(ram)} of total memory. Please ensure at least #{as_size(min_ram)} of total memory before running the installer."
+  end
+
+  if (min_cores = requirements['cores']) && cores < min_cores
+    raise Exception, "This system has #{cores} CPU cores. Please ensure at least #{min_cores} cores before running the installer."
+  end
+
+  if directories = requirements['directories']
+    (directories['minimum'] || {}).each do |directory, min_available|
+      available = available_space(directory)
+
+      if available < min_available
+        raise Exception, "Please ensure directory #{directory} has at least #{as_size(min_available)} available."
+      end
+    end
+  end
+end
+
+if app_value(:disable_system_checks)
+  logger.warn 'Skipping system checks.'
+else
+  begin
+    check
+  rescue Exception => e
+    logger.error e.message
+    exit(1)
+  end
+end
--- a/spec/hooks_pre_requirements_spec.rb
+++ b/spec/hooks_pre_requirements_spec.rb
@@ -1,0 +1,111 @@
+require 'spec_helper'
+require 'yaml'
+require 'kafo'
+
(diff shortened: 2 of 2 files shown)


hooks/pre/05-check_scenario_requirements.rb
```diff
@@ -0,0 +1,84 @@
...
+  a_list = File.expand_path(a).split('/')
+  b_list = File.expand_path(b).split('/')
+
+  b_list[0..a_list.size-1] == a_list
+end
+
+def facts(key)
+  command = Kafo::PuppetCommand.search_puppet_path('facter')
+  raise Exception, 'Facter not found' unless command
+  JSON.load(`#{command} --json error #{key}`)[key]
```
evgeni: why not just `#{command} --json` and then use the result as a huge Hash for local lookups. Saves us from running `facter` multiple times, which can be slow.

hooks/pre/05-check_scenario_requirements.rb
```diff
@@ -0,0 +1,84 @@
...
+def check_directory(directory, min_available)
+  mountpoints = facts('mountpoints')
+  mountpoint = mountpoints.keys.select { |mp| mp == directory || is_descendant?(mp, directory) }.group_by(&:size).max.last.first
+
+  if mountpoints[mountpoint]['available_bytes'] < min_available
+    raise Exception, "Please ensure directory #{directory} has at least #{as_size(min_available)} available."
+  end
+end
+
+if false && app_value(:disable_system_checks)
```
evgeni: That `false` should go, no? :)

hooks/pre/05-check_scenario_requirements.rb
```diff
@@ -0,0 +1,84 @@
...
+  required = {}
+  #required = {
+  #  'memory' => 8 * 1024 * 1024 * 1024,
+  #  'cores' => 1,
+  #  'directories' => {
+  #    '/' => 10 * 1024 * 1024 * 1024,
+  #  },
+  #}
+
+  check_mem(required['memory']) if required['memory']
```
evgeni: I must admit I would prefer if the checks would not be own methods. Mostly for the fact that they (usually, `check_dir` is an exception) are just `if a < b`.

my idea was to have

```ruby
checks = [
  { :name => 'memory', :fact => :memorysize_mb, :value => 7_900 }
]
checks.each do |check|
  if facts[check[:fact]] < check[:value]
    raise Exception, "This system has #{facts[check[:fact]]} #{check[:name]}, but at least #{check[:value]} are required."
  end
end
```

hooks/pre/05-check_scenario_requirements.rb
```diff
@@ -0,0 +1,84 @@
...
+  JSON.load(`#{command} --json error #{key}`)[key]
+end
+
+def check
+  # TODO: read required values from the scenario
+  #scenario = kafo.class.scenario_manager.select_scenario
+
+  required = {}
+  #required = {
+  #  'memory' => 8 * 1024 * 1024 * 1024,
```
evgeni: FYI, you want to check for less than 8G, more something like 7.8 or 7.9G, as if people have 8G memory, some of it might be claimed by kdump/crashkernel.

evgeni (changes requested):

hooks/pre/05-check_scenario_requirements.rb
ekohl: I also see that I left in "error" from ```--log-level error``` which was unavailable for facter 2. I'm still looking into this part but I hope that facter optimizes if you request a specific fact rather than the full data making it faster.

hooks/pre/05-check_scenario_requirements.rb
ekohl: That's why below I do a ```required * 0.9```

hooks/pre/05-check_scenario_requirements.rb
ekohl: I like the general idea but I wonder if it breaks down for directories and you end up over-engineering.

hooks/pre/05-check_scenario_requirements.rb
ekohl: That's why it's WIP. This allows me to run it directly from a non-hook context easing testing.

hooks/pre/05-check_scenario_requirements.rb
```diff
@@ -0,0 +1,94 @@
...
+end
+
+def check_cores(min_cores)
+  if facts('processors.count') < min_cores
+    raise Exception, "This system has less than #{min_cores} cores. Please ensure at least #{min_cores} cores before running the installer."
+  end
+end
+
+def check_directory(directory, min_available)
+  mountpoints = facts('mountpoints')
```
ekohl: Looks like this is facter 3+

ekohl:
Currently this is requiring Facter 3. It turns out Facter 2 is the only gem version you can get and it doesn't return structured facts. Facter 3 (at least by default) doesn't show the unstructured facts. Later I'll continue working on this and see if I can build the Facter 2 support so we can at least warn about CPUs and memory. Looks like your generic solution is needed after all @evgeni :)

ekohl:
I've decided that rather than working around kafo I'd fix some things in it. https://github.com/theforeman/kafo/pull/218 and https://github.com/theforeman/kafo/pull/219

ehelms:
Is this superceded by the tuning work?

ekohl:
Partly. This was the basis for the current implementation with tuning, but this also adds disk space checks. You can also recognize the `available_space` implementation that I suggested for the PostgreSQL migration work.

I think disk space checks are a valid thing to add, but I'd be fine with capturing that in a separate issue. It'd also need to be more dynamic IMHO. If you add Pulp, it needs to check the disk space for that for example.

PR #243, opened 2017-09-12
Refs #20893 - pin puppetlabs-stdlib

mmoll:


ekohl (approved):
In this case I'd expect the bot to accept a Refs because IMHO cross project references are always allowed.

ekohl:
What's weird is that when I run this locally I get a conflict between puppetlabs-stdlib < 4.2.0 and puppetlabs-stdlib/4.20.0. I've seen librarian struggle with pinning dependencies before.

ekohl:
Never mind that, I had a Puppetfile.lock.

PR #245, opened 2017-09-20
Switch puppetlabs/mysql back to a release

ekohl:


mmoll:
I'm unsure If the jump 3 -> 4 should be done in the 1.16 release cycle. While seeing that the major bump was only for Puppet version compatibility reasons, is there any gain form it?

ekohl:
Given we only just released the RC I think this should still be safe and I'd prefer to be on a release for our releases.

mmoll:
OK, please update Puppetfile.lock then :)

mmoll:
this also updates theforeman-puppet abnd puppetlabs-apache, but that's OK 👍

ekohl:
I know and I think those are actually good things

mmoll:
merged, bedankt @ekohl!

PR #246, opened 2017-09-20
Update 1.15-stable modules & git ignore

ekohl:


mmoll:
merged, bedankt @ekohl!

PR #247, opened 2017-09-27
Update Puppetfile.lock

xprazak2:
(cherry picked from commit f07567ee06742155a079f076823c6a3bfbc38120)


ekohl:
This is not what we want to do and it breaks installations. You pulled in major version updates of various modules that drop Puppet 3 support and 1.15 has committed to supporting that version.

Rather than cherry-picking ```Puppetfile.lock``` between branches we use ```librarian-puppet update``` since it'll respect pinnings in ```Puppetfile```. Even then we are careful and I selected only a few updates. We should pin more here, for example puppetlabs-stdlib 4.20.0 and kafo are incompatible (4cc5fe6e9c80ee9201e5cd404d032c3c4734e155 should be cherry-picked).

ekohl:
Closing in favor of https://github.com/theforeman/foreman-installer/pull/249

PR #248, opened 2017-09-27
Revert "Fixes #20889 - Use the Apache MPM event worker"

ekohl:
This reverts commit 4da2dee800b604033eeeb0a509b1a53a0ae5ab03. It turns
out there's a bug in Apache <2.4.25 which makes it unreliable. Since EL7
and Ubuntu 16.04 both ship a version with this bug, it'd only benefit
Debian Stretch.


mmoll:
merged, bedankt @ekohl!

PR #249, opened 2017-10-04
Update modules for 1.15.5

ekohl:
These include fixes for EL 7.4 and a fix for openscap in combination with puppet 4.


mmoll:
merged, bedankt @ekohl!

PR #251, opened 2017-10-05
Fixes #21225 - Set passenger_min_instances globally to

ehelms:
By default, puppet-puppet sets passenger_min_instances to processor
count which in some cases can starve out the Foreman passenger process
when. This sets the min to 1 globally to ensure Foreman and puppet
can share equally.


chris1984 (approved):

ekohl (approved):
This doesn't hurt the foreman use case and even though we don't use passenger for puppet anymore I think this is ok to merge.

mmoll:
I think this is OK, but is there something that should be changed also in puppet-puppet generally?

ehelms:
@mmoll I debated that, but if puppet-puppet is thought about in isolation, setting min instances to processorcount makes some sense.

ehelms:
@mmoll since you have unresolved question I will wait to merge this

mmoll:
I'm fine with this, merged, thanks @ehelms!

PR #252, opened 2017-10-06
Do proper deep copies when testing config/migrations

ekohl:


mmoll:
merged, bedankt @ekohl!

PR #254, opened 2017-10-16
update Puppetfile.lock for 1.16 RC2

mmoll:


mmoll:
If wanted, we could also update some dependencies...

ekohl:
> If wanted, we could also update some dependencies...

Let's do that.

mmoll:
updated some dependencies.

ekohl (approved):
This does mean we need similar katello changes to allow extlib 3 but I'll fix that.

ekohl:
Thanks!

PR #260, opened 2018-01-25
Fixes #22250 - prevent migration of custom hammer credentials

mbacovsky:


hooks/pre/20-check-hammer-credentials.rb
```diff
@@ -0,0 +1,30 @@
...
+POSSIBLE_RECENT_CONFIG_PATHS = [
+  '/etc/hammer/cli_conig.yml',
+  '/etc/hammer/cli.modules.d/foreman.yml',
+  '/root/.hammer/cli_conig.yml'
+]
+
+def has_password_set?(path)
+  if File.exist?(path)
+    config = YAML.load_file(path)
+    return true if config[:foreman].is_a?(::Hash) && (config[:foreman][:username != 'admin'] || config[:foreman][:password])
```
ekohl: I think the ```[:username != 'admin']``` should be ```[:username] != 'admin'```

ekohl (changes requested):
In general the code makes sense to me. I can't think of the various edge cases to know if something might have been missed.

One inline comment.

mbacovsky:
@ekohl, thanks for looking at it. Updated.
Cases I focused on are when the credentials are manually altered i.e. username is not the default 'admin' or the password is set. In such cases I presume the user has hammer configured and is using it so I'm preventing the installer to touch it.

hooks/pre/20-check-hammer-credentials.rb
```diff
@@ -0,0 +1,30 @@
+require 'yaml'
+require 'fileutils'
+
+NEW_HAMMER_CONFIG_PATH = '/root/.hammer/cli.modules.d'
+NEW_HAMMER_CONFIG_FILE = 'foreman.yml'
+POSSIBLE_RECENT_CONFIG_PATHS = [
+  '/etc/hammer/cli_conig.yml',
```
ekohl: conig => config

hooks/pre/20-check-hammer-credentials.rb
```diff
@@ -0,0 +1,30 @@
...
+  '/etc/hammer/cli.modules.d/foreman.yml',
+  '/root/.hammer/cli_conig.yml'
```
ekohl: conig => config

hooks/pre/20-check-hammer-credentials.rb
```diff
@@ -0,0 +1,30 @@
...
+end
+
+new_config_file = File.join(NEW_HAMMER_CONFIG_PATH, NEW_HAMMER_CONFIG_FILE)
+unless File.exist?(new_config_file)
+  # if there is foreman password or non-admin user set in any of the legacy configs
+  # create empty hammer foreman config to prevent installer from creating new one
+  if POSSIBLE_RECENT_CONFIG_PATHS.any? { |path| has_password_set?(path) }
+    FileUtils.mkdir_p(NEW_HAMMER_CONFIG_PATH)
+    File.open(new_config_file, "w+") do |file|
+      file.write("---\n:foreman: {}\n")
```
ekohl: I'm not sure if using the YAML module to output this would be cleaner but I'm sure this works too. Installer wise it just needs to exist.

ekohl (changes requested):
I got that and it makes sense.

hooks/pre/20-check-hammer-credentials.rb
mbacovsky: I used the YAML in earlier version but it produced 
```yaml
--- 
  !ruby/sym foreman: {}
```
which I didn't like for user config. Didn't found out how to suppress the ruby/sym notation.
Can use the YAML if you prefer it though.

ekohl: No, this is much simpler then. Let's go with this.

ekohl (approved):

ekohl:
Thanks!

ekohl:
I realized I rebased + merged rather than squash + merge :(

PR #262, opened 2018-01-30
Add Foreman Proxy scenario

sean797:


ekohl:
Could you add it here: https://github.com/theforeman/foreman-installer/blob/e98ef89455d09b922963e9364a612fee463f151b/spec/migration_spec.rb#L9

sean797:
Ah thanks, didn't check for any tests 👍

config/foreman-proxy-answers.yaml
```diff
@@ -0,0 +1,31 @@
+# Format:
+# <classname>: false - don't include this class
+# <classname>: true - include and use the defaults
+# <classname>:
+#   <param>: <value> - include and override the default(s)
+#
+# Every support plugin/compute class is listed, so that it
```
ekohl: compute might be redundant

config/foreman-proxy-answers.yaml
```diff
@@ -0,0 +1,31 @@
...
+# shows up in the interactive menu
+#
+# See params.pp in each class for what options are available
+---
+foreman_proxy:
+  custom_repo: true
+  use_autosignfile: true
```
ekohl: This parameter was removed in theforeman-foreman_proxy 7.0.0 so can be dropped

ekohl (changes requested):

sean797:
Thanks. Updated

ekohl (approved):
I think this is correct. @mmoll @ehelms @stbenjam could any of you have a look as well?

mmoll:
I didn't test this, but it's looking correct. As Puppet (server) is enabled by default, should it maybe get named "puppet-proxy" or similar to distinguish this from i.e. DNS or DHCP proxies?

ekohl:
I started to test this and we probably need something more in the Rakefile (and packaging) to include this. Will report back when I know more. Currently preparing the integration of katello-installer and one preparation patch from that will make it easy to add this scenario. Will report back when I know more.

config/foreman-proxy.yaml
```diff
@@ -0,0 +1,15 @@
+---
+:name: Foreman Proxy
+:description: Install a stand-alone Foreman Proxy.
+:log_dir: './_build/'
+:log_name: foreman-proxy.log
+:log_level: :debug
+:answer_file: ./config/foreman-proxy-answers.yaml
+:installer_dir: .
+:module_dirs: ./_build/modules
+:parser_cache_path: ./_build/parser_cache/foreman.yaml
```
ekohl, line 10: This should be `foreman-proxy`

ekohl (changes requested):
In https://github.com/theforeman/foreman-installer/pull/281 I've prepared a PR to generalize the packaging side of it to allow multiple scenarios. That should ensure this will actually work.

config/foreman-proxy.yaml
sean797: `foreman-proxy-content` doesn't have its own one. I thought this would just use `foreman.yml` ?

https://github.com/Katello/katello-installer/blob/master/config/foreman-proxy-content.yaml#L14-L16

ehelms: As long as the set of modules in the answer file is a subset of those in the foreman scenario this can be re-used like this.

ehelms:
I agree with @mmoll that this is a baseline puppet proxy and should likely indicate that. The alternative is to enable nothing by default and keep the scenario naming (which to me makes sense).

config/foreman-proxy.yaml
ehelms: I'm OK with using `foreman.yml` given I don't expect this to be stand-alone. @ekohl thoughts?

config/foreman-proxy-answers.yaml
```diff
@@ -0,0 +1,30 @@
...
+#
+# Every support plugin class is listed, so that it
+# shows up in the interactive menu
+#
+# See params.pp in each class for what options are available
+---
+foreman_proxy:
+  custom_repo: true
+puppet:
+  server: false
```
ekohl, line 15: Perhaps this should just set puppet to false by default? Not sure we should be including the agent config by default.

config/foreman-proxy.yaml
ekohl: I must admit I don't know enough about the exact parser cache. In the Rakefile I built it so it generates a cache file per scenario. If you add it, then it should generate the right configs and fix the config file in the build process: https://github.com/theforeman/foreman-installer/blob/7eb7ca11fecb88d7f785e504903db4e3350fc03a/Rakefile#L59

ekohl (changes requested):
Could you add it to the tests as well https://github.com/theforeman/foreman-installer/blob/7eb7ca11fecb88d7f785e504903db4e3350fc03a/spec/migration_spec.rb#L9

ehelms:
I included this in a newer PR. Closing in favor of it.

PR #263, opened 2018-02-17
Copy the repo list of debian9repo file and paste in /etc/apt/sources.…

Saurabh-Thakre:
…list.d

it will help geeks to install the foreman in debian 9


mmoll:
* go to https://theforeman.org/manuals/1.16/quickstart_guide.html
* select Debian 9 from the dropdown
* see all the needed information ready for copypasting

I think that should suffice?

PR #265, opened 2018-03-01
Prepare 1.17-RC2

ekohl:


mmoll (approved):

mmoll:
This looks correct. We need to have a look into those new kafo-linting errors at some point, but I have no problems to also merge this right now.

ehelms (approved):

ekohl:
@mmoll puppet-strings 1.2.0 came out and broke ```--emit-json-stdout```. Probably breaks all our cache generation and building too. https://tickets.puppetlabs.com/browse/PDOC-224

I'm hoping to get it quickly resolved upstream but otherwise we need to mass pin the gem. Kafo doesn't actually depend on it so we can't do it there afaik.

mmoll:
merged, bedankt @ekohl!

PR #266, opened 2018-03-26
Add a rake task to pin modules to released versions

ekohl:


ehelms:
Is the workflow:

 1) Run rake task
 2) Get standard output of pinned modules
 3) Copy output
 4) Open Puppetfile
 5) Paste content and remove previous module references

ekohl:
Prettty much that. I couldn't find a good way to do it in place. Perhaps I could use a global var to determine the output and then write that to Puppetfile. You'd lose comments but maybe that's not a problem.

ekohl:
Updated:
* Update the Puppetfile
* Add a missing closing quote
* Use the custom forge URL

ehelms (approved):

PR #267, opened 2018-03-27
Fixes #23030 - Add SCL Mongo Hiera

chris1984:


ekohl (changes requested):
Currently it's not loaded because it's not in `config/foreman-hiera.conf`. I think this should live in ```config/foreman.hiera/RedHat.yaml``` because it's only applicable to Red Hat distros. If we supported Fedora then that needed an exception because it doesn't have SCLs. Even if we currently don't use mongo on any of these platforms, it's good to do it the right way.

ehelms:
I feel like this is it's own special change that should be reflected by it's own hiera file. And honestly, if possible, in katello-installer otherwise this provides an odd split of functionality. Could we add a new katello.yml file that the hiera config looks for in the hierarchy?

ehelms:
Does this have to be added to the hierarchy or will it just work cause it exists @ekohl ?

ekohl (changes requested):
It needs to be added to the hierarchy.

chris1984:
@ehelms updated the pr

ehelms (approved):
@ekohl final review on you

config/foreman.hiera/mongo-redhat-7.yaml
```diff
@@ -0,0 +1,11 @@
...
+mongodb::globals::server_package_name: 'rh-mongodb34-mongodb-server'
+mongodb::globals::client_package_name: 'rh-mongodb34-mongodb'
+mongodb::globals::service_name: 'rh-mongodb34-mongod'
+mongodb::server::config: '/etc/opt/rh/rh-mongodb34/mongod.conf'
+mongodb::server::pidfilepath: '/var/opt/rh/rh-mongodb34/run/mongodb/mongod.pid'
+mongodb::server::dbpath: '/var/lib/mongodb'
+mongodb::server::syslog: true
+mongodb::server::logpath: false
+mongodb::server::config_template: 'mongodb/mongodb.conf.2.6.erb'
+mongodb::server::storage_engine: 'wiredTiger'
```
ehelms: @chris1984 This will force a change on upgrades won't it? Thus breaking upgrades? Remind me how were mitigating that.

config/foreman-hiera.conf
```diff
@@ -6,6 +6,7 @@
   - kafo_answers
   - security
   - tuning
+  - mongo-redhat-7
```
ekohl: The redhat should be `%{::osfamily}` and the `7` should be `%{::operatingsystemmajrelease}` so it's only loaded for EL7. You may need to rename the file to `mongo-RedHat-7.yaml`

ekohl (changes requested):
The actual yaml file looks good, but I'd prefer to avoid setting variables on other distros - even if they are probably not used on those.

config/foreman-hiera.conf
ehelms: @chris1984 can you tackle this point?

PR #269, opened 2018-04-04
Refs #22513 - Allow overriding postgresql server settings via hiera

chris1984:


config/foreman.hiera/custom.yaml
```diff
@@ -24,3 +24,7 @@
 # apache::mod::prefork::serverlimit: 256
 # apache::mod::prefork::maxclients: 256
 # apache::mod::prefork::maxrequestsperchild: 4000
+# Here are some examples of how you tune the Postgres options if needed:
```
ekohl: I'd add an empty line above this rather than below it.

config/foreman.hiera/custom.yaml
```diff
@@ -24,3 +24,7 @@
...
+#
+# postgresql::server::max_connections: 200
```
ekohl: It should be a hash:
```yaml
postgresql::server::config_entries:
  max_connections: 200
  shared_buffers: 256MB
```

ekohl (changes requested):

chris1984:
@ekohl updated added an empty line above and below so it is not all jammed together and easy to read.

ekohl (approved):
This is pending a puppetlabs-postgresql release, but other than that I think this is correct.

ehelms:
@ekohl If possible, I'd like to get this in before we branch the installer, which would include possibly setting postgres to nightly (or at least the commit we are wanting to include). I think you said in the past that would be OK, but wondering when you are seeing the branch point of the installer is to see if we could go ahead and do this.

Puppetfile
```diff
@@ -2,7 +2,7 @@ forge 'https://forgeapi.puppetlabs.com'
 
 # Dependencies
 mod 'puppetlabs/mysql',         '>= 4.0.0'
-mod 'puppetlabs/postgresql',    '>= 4.8.0'
+mod 'puppetlabs/postgresql',    :git => 'https://github.com/puppetlabs/puppetlabs-postgresql'
```
ekohl: Could you add a comment linking to https://tickets.puppetlabs.com/browse/MODULES-6963 so we know why it's there?

ekohl (changes requested):
I think the commit message could be clearer. How about `Allow overriding postgresql server settings via hiera`

chris1984:
@ekohl updated

config/foreman.hiera/custom.yaml
```diff
@@ -24,3 +24,9 @@
 # apache::mod::prefork::serverlimit: 256
 # apache::mod::prefork::maxclients: 256
 # apache::mod::prefork::maxrequestsperchild: 4000
+#
+# Here are some examples of how you tune the Postgres options if needed:
```
ekohl: I think the official spelling is PostgreSQL. @evgeni has been persistent in pointing this out so let's keep it correct.

chris1984:
@ekohl updated

ekohl (approved):
@mmoll what do you think?

mmoll (changes requested):
I'd rather fork that repo and have a defined state and not tap a foreign master branch

ehelms:
@mmoll Would setting Puppetfile to point at the commit in master we minimally want suffice for the time being?

mmoll:
@ehelms After having a look at the changes over there: Yes, that sounds sensible. :+1:

chris1984:
@mmoll how does this look?

Puppetfile
```diff
@@ -2,8 +2,8 @@ forge 'https://forgeapi.puppetlabs.com'
 
 # Dependencies
 mod 'puppetlabs/mysql',         '>= 4.0.0'
-mod 'puppetlabs/postgresql',    '>= 4.8.0'
-mod 'puppetlabs/puppetdb'
+# https://tickets.puppetlabs.com/browse/MODULES-6963
+mod 'puppetlabs/postgresql',    :git => 'https://github.com/puppetlabs/puppetlabs-postgresql' , :ref => 'Parameters `roles`, `config_entires`, and `pg_hba_rules` to `postgresql::server` for hiera'
```
mmoll: please use the git commit id here

ekohl: I'd also prefer the current latest commit in master since there are docs changes as well as others.

ekohl (changes requested):

chris1984:
@ekohl @mmoll how does it look now?

mmoll (dismissed):

Puppetfile
```diff
@@ -2,8 +2,7 @@ forge 'https://forgeapi.puppetlabs.com'
 
 # Dependencies
 mod 'puppetlabs/mysql',         '>= 4.0.0'
-mod 'puppetlabs/postgresql',    '>= 4.8.0'
-mod 'puppetlabs/puppetdb'
```
ekohl: Any reason you're removing puppetdb?

mmoll (approved):

chris1984:
Updated

ekohl (approved):

mmoll:
merged, thanks @chris1984!

PR #271, opened 2018-04-26
Fixes #16932 - Provide default postgresql tuning

ekohl:
As of yet this is untested but I wanted a place to link to.

### DIFF
--- a/config/foreman.hiera/custom.yaml
+++ b/config/foreman.hiera/custom.yaml
@@ -28,5 +28,5 @@
 # Here are some examples of how you tune the PostgreSQL options if needed:
 #
 # postgresql::server::config_entries:
-#   max_connections: 200
-#   shared_buffers: 256MB
+#   max_connections: 600
+#   shared_buffers: 1024MB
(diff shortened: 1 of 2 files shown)


psuriset:
Thanks for fixing this in foreman. These values seem to be good.

chris1984 (approved):

ekohl:
Confirmed this works for me on nightly.

mmoll:
This doesn't work with e.g. PostgreSQL 9.6:
```
LOG:  unrecognized configuration parameter "checkpoint_segments" in file "/etc/postgresql/9.6/main/postgresql.conf" line 629
FATAL:  configuration file "/etc/postgresql/9.6/main/postgresql.conf" contains errors
```
This exposes the same problem we had when trying to tune the Jenkins slaves. While RHEL 7.x always brings 9.2, recent Debian/Ubuntus bring 9.5, 9.6 or even 10. Also, there's the possibility to use postgresql.org repos with custom-hiera.yml to use recent PostgreSQLs with older OSes...

ekohl:
Interesting. I can put the merging config in common, set everything that is shared here and set segments in OS-specific configs or leave it out.

mmoll:
I just tested it with 9.6 and without `checkpoint_segments` but keeping the other settings it seems to work.

ekohl:
Updated to not set `checkpoint_segments`.

mmoll:
merged, bedankt @ekohl!

PR #275, opened 2018-06-09
Fixes #23871 - Report the proper error message on an empty IP address

ik5:


chris1984 (approved):

checks/dns.rb
```diff
@@ -12,7 +12,7 @@ def error_exit(message, code=2)
 hostname = `hostname -f`.chomp
 forwards = Resolv.getaddresses(hostname)
 
-if forwards.empty?
+if forwards.empty? || forewords.first.empty?
```
adamruzicka: `s/ewo/wa/` I guess

chris1984: Yes this needs to be changed.

chris1984 (changes requested):

checks/dns.rb
ik5: opps, so much for quick fix.

ekohl:
I'm interested to know how you reproduced this. Was there a messed up DNS server? Which ruby version and which OS?

ik5:
@ekohl the dns record did not find the fqdn, but the local host pointed in host to the fqdn, resulting in rejecting the address of `127.0.0.1` and `::ff`, and having only one single element in the array that is an empty string.

it's like once in a blue moon, every 50 dragon years type of a thing :)

I don't really think it's an issue for foreman-installer, but rather much more helpful information, so it will be easier to understand it. It took me like 30 minutes to understand that the issue is not an ipv6 for example, and then went to the code to understand what's going on :)

chris1984 (approved):
LGTM

ekohl:
I'd still like to know what triggered this. Can you tell the OS, ruby version, output of:
* `hostname -f`
* `dig HOSTNAME`
* `dig aaaa HOSTNAME`
* `cat /etc/hosts`

ik5:
`hostname -f` returned in that case
`foreman.xxxxx.local` (I do not disclosure the actual main domain).
but it was set under `/etc/hosts`:

```
$ cat /etc/hosts

127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6

127.0.0.1 foreman.xxxxx.local foreman
```

The reason for `hostname -f` resulting in `foreman.xxxxx.local` is due to `hostnamectl set-hostname foreman.xxxxx.local`.

The machine have two legs, one is external IP (but managed as a vlan that is not accessible from outside), and the other internal IP.

So the installer see that there is a FQDN structured domain, and start it's checking to do the reverse lookup, but for some reason, ruby itself does not return `127.0.0.1` (I tested it using `irb`), but an array with a single record of empty string.

The weird part, is that the C library do return it as `127.0.0.1`.

Please note that they fixed their DNS issues, and I removed the record from hosts.

ekohl:
Ah, that could very well be an invalid `/etc/hosts` config. I'll see if I can reproduce it. Perhaps we could even do a better detection and come up with a concrete suggestion to the user.

ik5:
@ekohl is there a testing mechanism fot it that I can create that mimics this issue?

ekohl:
@ik5 don't think we have a lot of test capabilities.

I'm wondering if this check isn't way too complicated. Basically we need `hostname -f` to match the `facter fqdn` output because some code relies on `facter` where our shell scripts use `hostname -f`. Long term we should move to explicit configuration where we no longer use the hostname, just as a default.

PR #280, opened 2018-06-22
Use a helper class for Puppetfile parsing

ekohl:
This doesn't use global variables and doesn't load the Puppetfile into the global context which provides some isolation.


ekohl:
@stbenjam I found a way not to use global variables :)

stbenjam (approved):
:+1:

mmoll:
merged, bedankt @ekohl!

PR #281, opened 2018-06-25
Refactor Rakefile to easily handle multiple scenarios

ekohl:
This is a preparation PR for merging katello-installer into foreman-installer but it also helps with additional scenarios (like https://github.com/theforeman/foreman-installer/pull/262).

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -56,6 +56,7 @@ LOGDIR = ENV['LOGDIR'] || "#{LOCALSTATEDIR}/log"
 DATAROOTDIR = DATADIR = ENV['DATAROOTDIR'] || "#{PREFIX}/share"
 MANDIR = ENV['MANDIR'] || "#{DATAROOTDIR}/man"
 PKGDIR = ENV['PKGDIR'] || File.expand_path('pkg')
+SCENARIOS = ['foreman']
 
 exporter_dirs = ENV['PATH'].split(':').push('/usr/bin', ENV['KAFO_EXPORTER'])
 exporter = exporter_dirs.find { |dir| File.executable? "#{dir}/kafo-export-params" } or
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -66,16 +67,46 @@ directory PKGDIR
...
+    scenario_config_replacements.each do |setting, value|
+      sh 'sed -i "s#\(.*%s:\).*#\1 %s#" %s' % [setting, value, t.name]
+    end
+  end
+
+  file "#{BUILDDIR}/parser_cache/#{scenario}.yaml" => ["#{BUILDDIR}/modules", "#{BUILDDIR}/parser_cache"] do |t|
+    sh "#{exporter}/kafo-export-params -c config/#{scenario}.yaml -f parsercache --no-parser-cache -o #{BUILDDIR}/parser_cache/#{scenario}.yaml"
+  end
+
+  # TODO: filename per scenario?
```
ekohl: I'm not entirely sure how to handle this. It doesn't really make sense to have the options of one scenario be part of the man page. Should we have a separate man page per scenario besides the generic one?

Rakefile
```diff
@@ -66,16 +67,46 @@ directory PKGDIR
...
+    'log_dir' => "#{LOGDIR}/foreman-installer",
+    'module_dirs' => "#{DATADIR}/foreman-installer/modules",
+    'parser_cache_path' => "#{DATADIR}/foreman-installer/parser_cache/#{scenario}.yaml",
+  }
   if ENV['KAFO_MODULES_DIR']
-    sh 'sed -i "s#.*\(:kafo_modules_dir:\).*#\1 %s#" %s' % [ENV['KAFO_MODULES_DIR'], t.name]
+    scenario_config_replacements['kafo_modules_dir'] = ENV['KAFO_MODULES_DIR']
+  end
+
+  file "#{BUILDDIR}/#{scenario}.yaml" => "config/#{scenario}.yaml" do |t|
```
ehelms: Personally, not a big fan of single variable names, makes reading require more interpretive overhead.

Rakefile
```diff
@@ -66,16 +67,46 @@ directory PKGDIR
...
+    cp t.prerequisites[0], t.name
+    scenario_config_replacements.each do |setting, value|
+      sh 'sed -i "s#\(.*%s:\).*#\1 %s#" %s' % [setting, value, t.name]
```
ehelms: Generall comment, I know we try to make the installer in theory usable as is, but a templated config would be nicer than sed.

Rakefile
```diff
@@ -153,36 +163,40 @@ task :build => [
...
-  "#{BUILDDIR}/parser_cache/foreman.yaml",
-]
+] + [SCENARIOS.map do |scenario|
+  [
+    "#{BUILDDIR}/#{scenario}.yaml",
+    "#{BUILDDIR}/#{scenario}.migrations",
+    "#{BUILDDIR}/#{scenario}-migrations-applied",
+    "#{BUILDDIR}/parser_cache/#{scenario}.yaml",
+  ]
+end].flatten
```
ehelms: This is a confusin bit of code to try and read.

Rakefile
ekohl: I think we actually use the scenario config to run tests as well. We can actually treat it as a YAML file because that's what the installer does at runtime anyway. That does mean we lose the comments but IMHO that's fine.

Rakefile
ekohl: I should probably introduce some helper variables

Rakefile
ehelms: That's tough in part because not every scenario has a 1:1 mapping with parser cache as well in their current form. There is a lot of re-use. Given this is how things currently work perhaps we can leave this as is for now?

ehelms:
I tested this with https://github.com/theforeman/foreman-installer/pull/262 and it worked without issue. I did not fully test the result but manually checked through the tarball. I'd love to see this get the minor comments addressed and merged.

Rakefile
ekohl: I think we should have a static `options.asciidoc` with the generic options and then a manpage per scenario.

ehelms: That makes sense if its easy enough to do. I don't think Katello has this today so the current functionality would not be a regression.

Rakefile
ekohl: I have verified this - we rely on correct values while building the parser cache. Templating then just means more work because you need to store an intermediate file.

Rakefile
```diff
@@ -89,16 +120,8 @@ file "#{BUILDDIR}/foreman-hiera.conf" => 'config/foreman-hiera.conf' do |t|
...
 
-file "#{BUILDDIR}/parser_cache/foreman.yaml" => ["#{BUILDDIR}/modules", "#{BUILDDIR}/parser_cache"] do |t|
-  sh "#{exporter}/kafo-export-params -c config/foreman.yaml -f parsercache --no-parser-cache -o #{BUILDDIR}/parser_cache/foreman.yaml"
-end
-
-file "#{BUILDDIR}/options.asciidoc" => "#{BUILDDIR}/parser_cache/foreman.yaml" do |t|
-  sh "#{exporter}/kafo-export-params -c config/foreman.yaml -f asciidoc -o #{BUILDDIR}/options.asciidoc"
-end
-
 file "#{BUILDDIR}/foreman-installer.8.asciidoc" =>
```
ekohl, line 123: This still only takes the foreman options into account. We do generate the asciidoc for every scenario. A simple future extension would be to start generating additional man pages (`man/foreman-installer-SCENARIO.8` perhaps)?

ehelms (approved):

PR #287, opened 2018-08-07
Add foreman_snapshot_management plugin

mdellweg:


ekohl:
Given this plugin is in the katello repository and needs Katello installed I think that the katello scenario in `katello-installer` is a better fit.

mdellweg:
@ekohl i'm sorry, but I think you confused something: This plugin does not depend on Katello.
I guess you thought about `foreman_scc_manager`.

ekohl (approved):
You are right, I was thinking of the wrong plugin.

ekohl:
Thanks!

PR #292, opened 2018-08-30
[WIP] Bump theforeman_proxy to 7.2.4

ekohl:
This is needed to fix https://projects.theforeman.org/issues/24628. Note that such a release doesn't actually exist yet and this is now a placeholder so I don't forget.


ekohl:
After discussion with @xprazak2 we decided not to change it for 1.18.3 since it's been this way for 1.18.2 and we haven't seen bug reports.

PR #296, opened 2018-10-15
Refs #25170 - Add migration for new seed param naming.

chris1984:


config/foreman.migrations/20181015123542_migrate_seed_params.rb
```diff
@@ -0,0 +1,6 @@
+mod = answers['foreman']
+if mod.is_a?(Hash)
+    ['admin_username', 'admin_password', 'admin_first_name', 'admin_last_name', 'admin_email'].each do |var|
+      mod["initial_#{var}"] = mod[var]
```
ekohl: Looks like this needs a `unless mod[var].nil?`

ekohl (changes requested):
Note: you can test this easily by changing `config/foreman-answers.yaml` and add `admin_username: myadmin` under `foreman:`, run `bundle exec rake spec`. It should complain that there's a `initial_admin_username` answer present which wasn't in the answers. It's how I usually test migrations. I'm thinking about making that easier so you can provide some fixtures.

chris1984:
@ekohl this is what I get when I do that, does the output look correct for the manual test?

https://paste.fedoraproject.org/paste/fFVtzNcOKF4tdjUeWHl5iw

ekohl (changes requested):
This now needs to be copied into the katello migration as well.

ekohl (approved):

PR #297, opened 2018-10-17
Fix rake build within RPM build

ehelms:
RPM building relies on an older Ruby 2.0 version where the source
property does not exist on a rake file task.


ehelms:
@ekohl We missed that builds have been failing since September 27th when we merged that update to the Rakefile. Here is a scratch build that should pass with this change -- http://koji.katello.org/koji/taskinfo?taskID=142041

ekohl (approved):
Weird that they passed locally.

PR #300, opened 2018-10-25
Fixes #25304 - Correct handle kafo_modules_dir

ekohl:
In 9929fe2ed9009fdec146a6ee75f2f2d7f084f6f7 the building process was changed. A regression was introduced that no longer uncommented the kafo_modules_dir.

Source (config/foreman.yaml):

    # :kafo_modules_dir: /usr/lib/ruby/vendor_ruby/kafo/modules

With the command:

    KAFO_MODULES_DIR=/path/to/kafo/modules bundle exec rake

Before the regression it produced:

    :kafo_modules_dir: /path/to/kafo/modules

After the regression it didn't uncomment the setting:

    # :kafo_modules_dir: /path/to/kafo/modules

This patch restores the old behavior.


tbrisker (approved):
Makes sense, thanks @ekohl !

PR #304, opened 2018-12-04
Add a script to create migrations

ekohl:


chris1984 (approved):
Have not tested it, but looks pretty straightforward.

ekohl:
I should note that it's based on https://github.com/Katello/katello-installer/blob/master/bin/create-migration but I went with shell to keep it straight forward.

ehelms:
I find this less straight forward. Bash is never as clean looking as Ruby, especially when I compare the two. Is there anyway we can move the foreman migrations to the same format easily so there is just one way to do things?

ekohl:
I considered that, but we'll need to ensure we don't run migrations twice. I'll have a look what's possible

ares:
I'd also prefer ruby implementation, but :+1: on the direction!

ekohl:
Rewritten as Ruby now

ehelms (approved):

PR #308, opened 2018-12-20
Fixes #25740 - Ensure a user is a member when removing

ekohl:
When a user is not part of a group, the removal fails. Because this is
the desired end state, the upgrade should not fail on that.

### DIFF
--- a/katello/hooks/pre/30-upgrade.rb
+++ b/katello/hooks/pre/30-upgrade.rb
@@ -163,8 +163,19 @@ def upgrade_qpid_paths
   end
 end
 
+def remove_user_from_group(user, group)
+  begin
+    is_member = Etc.getgrnam(group).mem.include?(user)
+  rescue ArgumentError
+    is_member = false
+  end
+  if is_member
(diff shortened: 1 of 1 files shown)


mmoll:
merged bedankt @ekohl!

evgeni (commented):
(my stupid client can't submit inline comments w/o a global one)

katello/hooks/pre/30-upgrade.rb
```diff
@@ -163,8 +163,19 @@ def upgrade_qpid_paths
...
+    is_member = false
+  end
+  if is_member
+    Kafo::Helpers.execute("gpasswd -d '#{user}' '#{group}'")
+  end
+end
+
 def drop_apache_foreman_group
-  Kafo::Helpers.execute("gpasswd -d apache foreman")
+  remove_user_from_group(apache, foreman)
```
evgeni, line 178: Shouldn't these be strings?

ekohl: You're right.

PR #309, opened 2018-12-27
Fixes #25740 - Use strings for user and group

ekohl:


mmoll:
merged bedankt @ekohl!

PR #312, opened 2019-01-11
Fixes #25844 - 1.21 installer migrations

ekohl:
This is not yet done because I need to go over more modules.


config/foreman.migrations/20190111180118_delete_removed_settings.rb
```diff
@@ -0,0 +1,63 @@
+DELETED = {
+  'foreman': [
+    'custom_repo',
+    'dynflow_in_core',
+    'email_conf',
+    'email_config_method',
+    'email_source',
+    'puppet_home',
+    'puppet_ssldir',
+  ],
```
tbrisker, line 10: Should we also add `authentication`, `locations_enabled` and `organizations_enabled` here?

ekohl: They should be in the clear section since the parameters still exist

ekohl:
Turns out my migration was bad. It was using symbols instead of strings which didn't actually delete the answers. Updated now and retesting.

ekohl:
Katello upgrade succeeds. Merging this now and we can see how to deal with other issues that might show up.

PR #314, opened 2019-01-16
Pin Puppet modules for 1.21

ekohl:


tbrisker (approved):
LGTM

tbrisker (approved):
Looks like this works now, very odd bug

PR #315, opened 2019-01-17
Rename katello-certs directory

ekohl:
There is an odd puppet bug that triggers when a directory with the same name exists:

```
$ puppet module install --version 5.0.0 --target-dir /tmp/wd --module_repository https://forgeapi.puppetlabs.com --modulepath /tmp/wd --module_working_dir /tmp/wd --ignore-dependencies katello-certs
Notice: Preparing to install into /tmp/wd ...
Notice: Downloading from https://forgeapi.puppetlabs.com ...
Notice: Installing -- do not interrupt ...
/tmp/wd
└── katello-certs (v5.0.0)
$ mkdir katello-certs
$ puppet module install --version 5.0.0 --target-dir /tmp/wd --module_repository https://forgeapi.puppetlabs.com --modulepath /tmp/wd --module_working_dir /tmp/wd --ignore-dependencies katello-certs
Notice: Preparing to install into /tmp/wd ...
Error: Is a directory @ io_getpartial - katello-certs
Error: Try 'puppet help module install' for usage
```

This then fails. By renaming the directory we avoid this collision.


ekohl:
Doing some more verification in https://github.com/theforeman/foreman-installer/pull/314

PR #317, opened 2019-01-31
Fixes #25951 - always print the credentials after install

evgeni:
authentication is mandatory now and the param is undef


ekohl (approved):
Recently I thought about this, probably while in the shower so didn't write it down.

tbrisker:
1.21 - a6e338f

PR #322, opened 2019-02-26
Add a puppet version check

ekohl:
This means we execute the tests twice, but that's not a lot of overhead.  It does mean we can run our cheap tests before doing the expensive full build.

This currently fails and lead to https://community.theforeman.org/t/dropping-puppet-4-support-in-our-installer/13029

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,11 @@ script:
   - bundle exec $INSTDIR/sbin/foreman-installer --help --scenario foreman-proxy-content
   - bundle exec $INSTDIR/sbin/foreman-installer --help --scenario katello
   #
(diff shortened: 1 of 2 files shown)


evgeni:
just to understand your intent here, you define a list of puppet versions we want to support, and then you ensure that all our (3rd party) modules actually support these versions? correct?

ekohl:
Correct, that is the intent. The original goal was to verify Puppet 6 support but now it has shown me the issue with Puppet 4.

evgeni:
okay, then I like the idea! now get that p4 support out and ack ;)

ekohl:
Removed Puppet 4

ehelms (approved):

PR #323, opened 2019-03-06
Add a DHCP warning

ekohl:
It is not obvious to users that the DHCP config is not automatically managed by Foreman. This adds a warning to the file instructing the user.

The text was written by Lukáš Zapletal.

It depends on https://github.com/theforeman/puppet-dhcp/pull/149. This makes it easy to customize the message in downstream, for example to include specific help URLs.

### DIFF
--- a/config/foreman.hiera/common.yml
+++ b/config/foreman.hiera/common.yml
@@ -1,0 +1,13 @@
+---
+dhcp::config_comment: >
+  READ: This file was written the foreman-installer and not by the Foreman
+
+  application. Any updates to subnets in the Foreman database are not
+
+  automatically reflected in this configuration and vice versa. Configuration
+
+  updates like DNS servers or adding/removing subnets must be done both in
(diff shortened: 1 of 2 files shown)


config/foreman.hiera/common.yml
```diff
@@ -0,0 +1,13 @@
+---
+dhcp::config_comment: >
+  READ: This file was written the foreman-installer and not by the Foreman
+
```
ekohl, line 4: These odd lines are needed, otherwise YAML interprets it as a single line.

zjhuntin (approved):
Looks good to me. The warning comes through :+1:

mmoll:
merged, bedankt @ekohl!

lzap:
Nice, this way we can add more warnings. Any good ideas? This was pain in my ass honestly, maybe DNS is similar case but it does not appear that often.

PR #324, opened 2019-03-06
Fixes #26256 - Fix proxy cert generation hooks

ekohl:
When merging the installers, the hooks were removed from the certificate generation process because most hooks are irrelevant. Not all, so this caused a regression.

Rather than enabling all hooks again, this takes the more minimal route of only adding the relevant hooks.


ekohl:
I've moved the ticket. For some reason I can't easily build a scratch RPM to try this out right now.

ekohl:
@johnpmitsch Tested with a scratch build on http://koji.katello.org/koji/taskinfo?taskID=178293 and verified I got instructions again.

ekohl:
Amended the commit so it properly updated Redmine.

johnpmitsch (approved):
works! :+1:

PR #325, opened 2019-03-08
Fixes #25842 - Correct DNS comparisons

ekohl:
DNS labels are case insenstive so regular string comparisons don't always work. The Resolv::DNS::Name class implements the correct comparisons.


mmoll:
merged, bedankt @ekohl!

PR #327, opened 2019-03-12
Fixes #26323 - Remove max_tasks_per_child setting

chris1984:


chris1984:
@ekohl let me know if that looks right, it sounds right but only the 2nd time I am doing one of these.

ekohl (approved):
LGTM. Can be merged after the puppet module PRs have been merged.

PR #332, opened 2019-03-14
Improve source file generation

ekohl:
This guarantees the file is actually freshly generated and outputs the generated filename. This avoids the need to glob.


mmoll:
works as expected, merged, bedankt @ekohl!

PR #333, opened 2019-03-15
Refs #26330 - Fix syntax on user group migration

ekohl:
002af7bb53b652f86709e879e72270b385161afa introduced a bug in the migration by using 2.0 incompatible syntax.


mmoll:
merged, bedankt @ekohl!

PR #335, opened 2019-03-18
Refs #26282 - Correct hiera config file location

ekohl:


mmoll:
I thought that was done on purpose before, but if changes can be simply minimized back, :+1:

mmoll:
merged, bedankt @ekohl!

ekohl:
Since the file on disk after installation is `/usr/share/foreman-installer/config/foreman-hiera.yaml` this is needed. In the packaged scenario file it was already correctly written via `Rakefile` but we needed a migration for upgrading users.

PR #336, opened 2019-03-21
Fixes #26409 - Locking of packages after installation

mbacovsky:
This patch add a new option --[no-]lock-package-versions.
The option enables/disables version locking of installed packages.
Once used the choice is stored in the scenario "custom" hash and
is reused in the following installer runs.

If the packages are locked when the installer starts
unlock is performed before installer proceed with the installation.

### DIFF
--- a/katello/hooks/boot/09-version_locking.rb
+++ b/katello/hooks/boot/09-version_locking.rb
@@ -1,9 +1,10 @@
-return nil unless system('command -v foreman-maintain > /dev/null')
-app_option(
-  '--[no-]lock-package-versions',
-  :flag,
-  "Let installer lock versions of the installed packages to prevent\n" \
-    "unexpected breakages during system updates. The choice is remembered\n" \
-    "and used in next installer runs.",
-  :default => nil
-)
+if system('command -v foreman-maintain > /dev/null')
+  app_option(
+    '--[no-]lock-package-versions',
+    :flag,
+    "Let installer lock versions of the installed packages to prevent\n" \
+      "unexpected breakages during system updates. The choice is remembered\n" \
+      "and used in next installer runs.",
+    :default => nil
+  )
+end
--- a/katello/hooks/boot/09-version_locking.rb
+++ b/katello/hooks/boot/09-version_locking.rb
@@ -1,3 +1,4 @@
+return nil unless system('command -v foreman-maintain > /dev/null')
 app_option(
   '--[no-]lock-package-versions',
   :flag,
--- a/katello/hooks/boot/09-version_locking.rb
+++ b/katello/hooks/boot/09-version_locking.rb
@@ -1,0 +1,8 @@
+app_option(
+  '--[no-]lock-package-versions',
+  :flag,
+  "Let installer lock versions of the installed packages to prevent\n" \
+    "unexpected breakages during system updates. The choice is remembered\n" \
+    "and used in next installer runs.",
+  :default => nil
+)
--- a/katello/hooks/pre_commit/09-version_locking.rb
+++ b/katello/hooks/pre_commit/09-version_locking.rb
@@ -9,11 +9,11 @@ if lock_versions != custom_config_value
 end
 
 if lock_versions
-  unless system('command -v foreman-maintain > dev/null')
+  unless system('command -v foreman-maintain > /dev/null')
     Kafo::Helpers.log_and_say :error, 'Locking of package versions was requested but foreman-maintain is not installed'
     kafo.class.exit 1
   end
-  unless system('foreman-maintain packages -h > /dev/null')
+  unless system('foreman-maintain packages -h > /dev/null 2>&1')
     Kafo::Helpers.log_and_say :error, 'Locking of package versions was requested but foreman-maintain version installed does not support it'
     kafo.class.exit 1
   end
--- a/katello/hooks/pre_commit/09-version_locking.rb
+++ b/katello/hooks/pre_commit/09-version_locking.rb
@@ -8,14 +8,21 @@ if lock_versions != custom_config_value
   kafo.config.configure_application
 end
 
-# unlock packages if locked
-`command -v foreman-maintain`
-if $?.success?
-  `foreman-maintain packages is-locked --assumeyes`
-  if $?.exitstatus == 0
-    Kafo::Helpers.log_and_say :info, "Package versions are locked. Continuing with unlock."
-    Kafo::Helpers.execute('foreman-maintain packages unlock --assumeyes')
+if lock_versions
+  unless system('command -v foreman-maintain > dev/null')
+    Kafo::Helpers.log_and_say :error, 'Locking of package versions was requested but foreman-maintain is not installed'
+    kafo.class.exit 1
   end
+  unless system('foreman-maintain packages -h > /dev/null')
+    Kafo::Helpers.log_and_say :error, 'Locking of package versions was requested but foreman-maintain version installed does not support it'
+    kafo.class.exit 1
+  end
+end
+
+# unlock packages if locked
+if system('foreman-maintain packages is-locked --assumeyes > /dev/null')
+  Kafo::Helpers.log_and_say :info, "Package versions are locked. Continuing with unlock."
+  Kafo::Helpers.execute('foreman-maintain packages unlock --assumeyes', false)
 end
 nil
 
--- a/katello/hooks/pre_commit/09-version_locking.rb
+++ b/katello/hooks/pre_commit/09-version_locking.rb
@@ -12,8 +12,8 @@ end
 `command -v foreman-maintain`
 if $?.success?
   `foreman-maintain installation is-locked --assumeyes`
-  if $?.exitstatus == 1
-    Kafo::Helpers.log_and_say :info, "Package versionss are locked. Continuing with unlock."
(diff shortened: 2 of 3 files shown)


mbacovsky:
This PR requires version locking feature in foreman-maintain to be merged https://github.com/theforeman/foreman_maintain/pull/249

katello/hooks/pre_commit/09-version_locking.rb
```diff
@@ -0,0 +1,21 @@
...
+  store_custom_config(:lock_package_versions, lock_versions)
+  kafo.config.configure_application
+end
+
+# unlock packages if locked
+`command -v foreman-maintain`
+if $?.success?
+  `foreman-maintain installation is-locked --assumeyes`
+  if $?.exitstatus == 1
+    Kafo::Helpers.log_and_say :info, "Package versionss are locked. Continuing with unlock."
```
ekohl:
```suggestion
    Kafo::Helpers.log_and_say :info, "Package versions are locked. Continuing with unlock."
```

katello/hooks/pre_commit/09-version_locking.rb
```diff
@@ -0,0 +1,21 @@
...
+if lock_versions != custom_config_value
+  store_custom_config(:lock_package_versions, lock_versions)
+  kafo.config.configure_application
+end
+
+# unlock packages if locked
+`command -v foreman-maintain`
+if $?.success?
+  `foreman-maintain installation is-locked --assumeyes`
+  if $?.exitstatus == 1
```
ekohl: Exit status 1 surprises me here. I would expect that you could do `foreman-maintain installation is-locked --assumeyes && foreman-maintain installation unlock --assumeyes` on the shell to ensure it's unlocked. Similarly `foreman-maintain installation is-locked --assumeyes || foreman-maintain installation lock --assumeyes` to lock.

ekohl (commented):
Untested, but overall it looks pretty clean.

mbacovsky:
@ekohl thanks for the review. I fixed the typo and changed the exitcode in formean-maintain and updated here. The changes are in separate commit and needs to be squashed on merge.

mbacovsky:
Updated, the `foreman-maintain installation` was renamed to `foreman-maintain packages`.

katello/hooks/pre_commit/09-version_locking.rb
```diff
@@ -0,0 +1,21 @@
...
+lock_versions = cli_param.nil? ? !!custom_config_value : cli_param
+
+if lock_versions != custom_config_value
+  store_custom_config(:lock_package_versions, lock_versions)
+  kafo.config.configure_application
+end
+
+# unlock packages if locked
+`command -v foreman-maintain`
+if $?.success?
```
ekohl: If this failed and the user indicated they want to lock this should be fatal.

katello/hooks/pre_commit/09-version_locking.rb
```diff
@@ -0,0 +1,21 @@
...
+  `foreman-maintain packages is-locked --assumeyes`
+  if $?.exitstatus == 0
+    Kafo::Helpers.log_and_say :info, "Package versions are locked. Continuing with unlock."
+    Kafo::Helpers.execute('foreman-maintain packages unlock --assumeyes')
```
ekohl: I assume unlocking is idempotent so do we even need to check if it's locked? IIRC the execute helper also prints + logs the output so the log line might not be needed. This will mostly depend on the output of foreman-maintain and I'll leave this up to you.

katello/hooks/boot/09-version_locking.rb
```diff
@@ -0,0 +1,8 @@
+app_option(
```
ekohl: One thought to implement this in regular foreman is to check if foreman-maintain is present and only then add the option. Probably best to leave that out of this PR though since it introduces complexities.

ekohl (commented):
Edit: started to write something here and then applied to lines.

katello/hooks/pre_commit/09-version_locking.rb
mbacovsky: @ekohl the unlock will probably be expensive on Debian so I wanted to do it only when necessary. On yum based systems it is probably not the case and the check could be skipped. I'll keep the foreman-maintain output in log only.

katello/hooks/boot/09-version_locking.rb
mbacovsky: I made an attempt to handle missing foreman-maintain. Though I guess it is a corner case as this is katello-only and we depend on f-m there.

katello/hooks/pre_commit/09-version_locking.rb
mbacovsky: Added some more checks

mbacovsky:
@ekohl I've added some more error handling and tuned the output. Please check. It is in a new commit.

katello/hooks/pre_commit/09-version_locking.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+end
+
+# unlock packages if locked
+if system('foreman-maintain packages is-locked --assumeyes > /dev/null')
+  Kafo::Helpers.log_and_say :info, "Package versions are locked. Continuing with unlock."
+  Kafo::Helpers.execute('foreman-maintain packages unlock --assumeyes', false)
+end
+nil
+
```
ekohl: Nit: empty line

katello/hooks/pre_commit/09-version_locking.rb
```diff
@@ -0,0 +1,28 @@
...
+    kafo.class.exit 1
+  end
+  unless system('foreman-maintain packages -h > /dev/null 2>&1')
+    Kafo::Helpers.log_and_say :error, 'Locking of package versions was requested but foreman-maintain version installed does not support it'
+    kafo.class.exit 1
+  end
+end
+
+# unlock packages if locked
+if system('foreman-maintain packages is-locked --assumeyes > /dev/null')
```
ekohl: This still makes it a hard requirement. Should this be inside `if lock_versions`?

mbacovsky: @ekohl this is outside `lock_versions` intentionally for the case the packages were locked manually and auto-locking is not enabled in installer. What behavior of installer would you expect in such case?

It is not actually hard req as the condition is negative also when the command does not exist. We may add `2>&1` to hide the error out though.

ekohl: You're right, I totally missed that. Nice :)

mbacovsky: @ekohl I assume it is okay to let installer unlock manually locked packages, please let me know if it is not. I also suppressed the error out for clean output. Updated.

ekohl: IMHO as long as it's documented, that's fine. Users can still opt of out locking and handle it manually if I read the code correct. That's what we would advise users.

mbacovsky:
@ekohl now when the PR does not depend on the foreman-maintain what prevents it from being merged? Is there anything I can help with?

jameerpathan111:
@mbacovsky @ekohl 
I have tested this PR, it LGTM, and it works as expected.
I haven't found any major issues in it yet.

```
# ./bin/foreman-maintain packages is-locked
Foreman related packages are not locked
```
```
# yum update
=========================================================================================================================================================================================
 Package                                                 Arch         Version                        Repository                                                                     Size
=========================================================================================================================================================================================
Installing:
 satellite-common                                        noarch       6.5.0-11.el7sat                Sat6-CI_Red_Hat_Satellite_6_5_Composes_Satellite_6_5_RHEL7                     15 k
     replacing  rh-ruby24-ruby.x86_64 2.4.5-91.el7
...
...#skipping full list of packages being updated
Updating:
 ansiblerole-insights-client                             noarch       1.6-1.el7sat                   Sat6-CI_Red_Hat_Satellite_6_5_Composes_Satellite_6_5_RHEL7                     17 k
...
...#skipping full list of packages being updated
Transaction Summary
=========================================================================================================================================================================================
Install    1 Package  (+77 Dependent packages)
Upgrade  220 Packages

Total download size: 472 M
Is this ok [y/d/N]: n

```
```
# satellite-installer --lock-package-versions
Resetting puppet server version param...
Installing             Done                                               [100%] [......................................................................................................]
Package versions are being locked.
```
```
# yum update

=========================================================================================================================================================================================
 Package                              Arch                                    Version                                          Repository                                           Size
=========================================================================================================================================================================================
Updating:
 zsh                                  x86_64                                  5.0.2-31.el7                                     rhel-7-server-rpms                                  2.4 M

Transaction Summary
=========================================================================================================================================================================================
Upgrade  1 Package

Total download size: 2.4 M
Is this ok [y/d/N]: n

```
```
# ./bin/foreman-maintain packages is-locked
Foreman related packages are locked
```
```
# satellite-installer --no-lock-package-versions
Resetting puppet server version param...
Package versions are locked. Continuing with unlock.
Installing             Done                                               [100%] [......................................................................................................]
  Success!
```

ekohl (approved):
I don't see any more objections, but I'm holding off on merging till I've completed all the module releases. Will look over this before branching.

ekohl:
Looks like I forgot to squash on merge.

PR #338, opened 2019-04-02
Release 1.21.1

ekohl:


Puppetfile.lock
```diff
@@ -9,7 +9,7 @@ FORGE
       puppetlabs-concat (< 6.0.0, >= 1.0.0)
       puppetlabs-postgresql (< 6.0.0, >= 4.2.0)
       puppetlabs-stdlib (< 6.0.0, >= 4.2.0)
-    katello-certs (5.0.0)
+    katello-certs (5.1.1)
```
tbrisker: I think this is missing https://github.com/theforeman/puppet-certs/pull/246, or do you think that should not be in 1.21.1?

ekohl: I considered it, but didn't think it was needed. Puppet ships its own ruby that is not FIPS compliant so it's not actually an issue in most cases.

mmoll (approved):

Puppetfile.lock
ekohl: Looks like I did cherry pick it in 1.20.3 so cherry picking it into 1.21.1 makes sense.

ekohl:
Updated with certs 5.1.2

tbrisker (approved):

mmoll:
merged, bedankt @ekohl!

PR #339, opened 2019-04-04
Fixes #26538 - return true if the user is not member of the group

evgeni:


katello/hooks/pre/30-upgrade.rb
```diff
@@ -171,6 +171,8 @@ def remove_user_from_group(user, group)
   end
   if is_member
     Kafo::Helpers.execute("gpasswd -d '#{user}' '#{group}'")
+  else
+    return true
```
ekohl:
```suggestion
    true
```

ekohl (approved):

tbrisker:
1.21.2 - de6efed

PR #343, opened 2019-04-12
Use UTC timestamp in migration name

mdellweg:
@ekohl I guess we want them to be in the right order...

### DIFF
--- a/bin/create-migration
+++ b/bin/create-migration
@@ -24,7 +24,7 @@ else
   glue = '-'
 end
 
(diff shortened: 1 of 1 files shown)


ekohl (approved):

ekohl:
Did you find an issue with this or were you just reading the source? I suspect this hasn't really been a problem because we have very few migrations.

mdellweg:
I saw an unexpected time in the timestamp. And then i thought: 'This can easily be fixed.'

PR #347, opened 2019-04-22
Fixes #25564 - addedTargetOption

patilsuraj767:
This issue was caused because of the fix in the following RFE.

[RFE] katello-certs-check to distinguish between Satellite and Capsule
https://projects.theforeman.org/issues/22694

Code snippet -
```
HOSTNAME=$(hostname -f)
CERT_HOSTNAME=$(openssl x509 -noout -subject -in $CERT_FILE | sed -e 's/^subject.*CN=\([a-zA-Z0-9\.\-]*\).*$/\1/')

if [ $EXIT_CODE == "0" -a $CERT_HOSTNAME == $HOSTNAME ]; then
echo -e "${GREEN}Validation succeeded${RESET}\n"
    cat <<EOF

To install the Katello main server with the custom certificates, run:

    satellite-installer --scenario satellite\\
                      --certs-server-cert "$(readlink -f $CERT_FILE)"\\
                      --certs-server-key "$(readlink -f $KEY_FILE)"\\
                      --certs-server-ca-cert "$(readlink -f $CA_BUNDLE_FILE)"


```
Issue is, we are checking the hostname against the cert CN and in case of wildcard cert CN is *.example.com. So this condition fails in case of wildcard certs that are to be installed in the satellite server.

Need to modify the logic, better to give the user the ability to define target for which cert to be generated.

```
katello-certs-check -t foreman -c /customcerts/cert_crt.pem -k /customcerts/cert_key.pem -b /customcerts/CA_crt.pem
```

I have earlier opened the PR for the same but it was closed because of my some force commits and lack of git knowledge and I don't think I can reopen the same PR again so creating this new one.  We can refer the old PR for the discussion we had regarding to this. 
https://github.com/Katello/katello-installer/pull/718

### DIFF
--- a/bin/katello-certs-check
+++ b/bin/katello-certs-check
@@ -11,7 +11,7 @@ function usage () {
 Verifies, that custom SSL certificate files are usable
 as part of the Katello installation. When passing filenames use absolute paths.
 
-usage: $0 -t TARGET -c CERT_FILE -k KEY_FILE -b CA_BUNDLE_FILEusage: $0 -c CERT_FILE -k KEY_FILE -b CA_BUNDLE_FILE
+usage: $0 -t TARGET -c CERT_FILE -k KEY_FILE -b CA_BUNDLE_FILE
 HELP
 }
 
--- a/bin/katello-certs-check
+++ b/bin/katello-certs-check
@@ -18,7 +18,7 @@ HELP
 while getopts "t:c:k:b:" opt; do
     case $opt in
         t)
-            TARGET=`echo $OPTARG|tr '[:upper:]' '[:lower:]'`
+            TARGET=$(echo $OPTARG|tr '[:upper:]' '[:lower:]')
             ;;
         c)
             CERT_FILE="$(readlink -f $OPTARG)"
@@ -42,14 +42,8 @@ done
 
 EXIT_CODE=0
 
-if [ -z "$CERT_FILE" -o -z "$KEY_FILE" -o -z "$CA_BUNDLE_FILE" -o -z "$TARGET" ]; then
+if [ -z "$CERT_FILE" -o -z "$KEY_FILE" -o -z "$CA_BUNDLE_FILE" ]; then
     echo 'One of the required parameters is missing.' >&2
-    usage
-    exit 1
-fi
-
-if [[ $TARGET != "foreman" && $TARGET != "foreman-proxy" ]]; then
-    echo -e 'Target not specified or incorrect. Target value should be foreman or foreman-proxy \n' >&2
     usage
     exit 1
 fi
@@ -171,7 +165,7 @@ check-ca-bundle
 check-cert-san
 check-cert-usage-key-encipherment
 
-if [ $EXIT_CODE == "0" -a "$TARGET" == "foreman" ]; then
+if [[ $EXIT_CODE == "0" ]] && ([[ $TARGET == "foreman" ]] || [[ -z "$TARGET" ]]) ; then
     echo -e "${GREEN}Validation succeeded${RESET}\n"
     cat <<EOF
 
@@ -189,6 +183,8 @@ To update the certificates on a currently running Katello installation, run:
                       --certs-server-key "$(readlink -f $KEY_FILE)" \\
                       --certs-server-ca-cert "$(readlink -f $CA_BUNDLE_FILE)" \\
                       --certs-update-server --certs-update-server-ca
+
+To use them inside a NEW \$FOREMAN_PROXY, rerun this command with -t foreman-proxy
 EOF
 elif [[ $EXIT_CODE == "0" ]]; then
   echo -e "${GREEN}Validation succeeded${RESET}\n"
--- a/bin/katello-certs-check
+++ b/bin/katello-certs-check
@@ -11,14 +11,14 @@ function usage () {
 Verifies, that custom SSL certificate files are usable
 as part of the Katello installation. When passing filenames use absolute paths.
 
-usage: $0 -t TARGET -c CERT_FILE -k KEY_FILE -b CA_BUNDLE_FILE
+usage: $0 -t [foreman|foreman-proxy] -c CERT_FILE -k KEY_FILE -b CA_BUNDLE_FILE
 HELP
 }
 
 while getopts "t:c:k:b:" opt; do
     case $opt in
         t)
-            TARGET=$OPTARG
+            TARGET=`echo $OPTARG|tr '[:upper:]' '[:lower:]'`
             ;;
         c)
             CERT_FILE="$(readlink -f $OPTARG)"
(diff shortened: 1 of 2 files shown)


chris1984:
Working on testing this

bin/katello-certs-check
```diff
@@ -11,12 +11,15 @@ function usage () {
 Verifies, that custom SSL certificate files are usable
 as part of the Katello installation. When passing filenames use absolute paths.
 
-usage: $0 -c CERT_FILE -k KEY_FILE -b CA_BUNDLE_FILE
+usage: $0 -t TARGET -c CERT_FILE -k KEY_FILE -b CA_BUNDLE_FILE
```
chris1984: Is it possible to add a section to the help saying what options we have for the -target?

bin/katello-certs-check
```diff
@@ -39,14 +42,17 @@ done
...
+if [ -z "$CERT_FILE" -o -z "$KEY_FILE" -o -z "$CA_BUNDLE_FILE" -o -z "$TARGET" ]; then
     echo 'One of the required parameters is missing.' >&2
     usage
     exit 1
 fi
 
-HOSTNAME=$(hostname -f)
-CERT_HOSTNAME=$(openssl x509 -noout -subject -in $CERT_FILE | sed -e 's/^subject.*CN=\([a-zA-Z0-9\.\-]*\).*$/\1/')
+if [[ $TARGET != "foreman" && $TARGET != "foreman-proxy" ]]; then
+    echo -e 'Target not specified or incorrect. Target value should be foreman or foreman-proxy \n' >&2
```
chris1984: For the error, I think we should have Foreman and Foreman-proxy capitalized and then take the input from the user and make it lowercase or adjust for that, but that can be done in a separate pr and is not blocking this, more just a grammar nitpick.

spec/fixtures/katello-certs-check/missing-parameter.txt
```diff
@@ -2,4 +2,4 @@ One of the required parameters is missing.
 Verifies, that custom SSL certificate files are usable
 as part of the Katello installation. When passing filenames use absolute paths.
 
-usage: |COMMAND| -c CERT_FILE -k KEY_FILE -b CA_BUNDLE_FILE
+usage: |COMMAND| -t TARGET -c CERT_FILE -k KEY_FILE -b CA_BUNDLE_FILE
```
chris1984: This is not needed, this is just for Puppet related things, the error in the Bash script if a user misses a param is good enough.

chris1984 (changes requested):
Looks good @patilsuraj767 just a few comments and I would be happy to get it merged. Works good with my testing and when I break the SSL cert the checks still work correctly too.

bin/katello-certs-check
ekohl: I think this is sometimes noted as:
```suggestion
usage: $0 -t [foreman|foreman-proxy] -c CERT_FILE -k KEY_FILE -b CA_BUNDLE_FILE
```

spec/fixtures/katello-certs-check/missing-parameter.txt
ekohl: Why is this not needed? This is the output when the script is ran and we compare those. AFAIK this prevents the tests from failing.

ekohl (commented):
Just a sanity check: the target is foreman but then it says `To install the Katello main server`. Naming is hard so I don't have a good suggestion here. Also wondering how we'd run this downstream. Would it be satellite/capsule then?

Overall I still like the idea of explicitly telling what the expected outcome is, but one option is to just display the full message (since the target is only used for that) every time.

chris1984:
@ekohl sorry on the spec line, I read it wrong and didn't see it was already in there. @patilsuraj767 ignore that comment. Thanks for catching it

bin/katello-certs-check
```diff
@@ -39,14 +42,17 @@ done
...
+if [ -z "$CERT_FILE" -o -z "$KEY_FILE" -o -z "$CA_BUNDLE_FILE" -o -z "$TARGET" ]; then
     echo 'One of the required parameters is missing.' >&2
     usage
     exit 1
 fi
 
-HOSTNAME=$(hostname -f)
-CERT_HOSTNAME=$(openssl x509 -noout -subject -in $CERT_FILE | sed -e 's/^subject.*CN=\([a-zA-Z0-9\.\-]*\).*$/\1/')
+if [[ $TARGET != "foreman" && $TARGET != "foreman-proxy" ]]; then
+    echo -e 'Target not specified or incorrect. Target value should be Foreman or Foreman-proxy \n' >&2
```
ekohl: It's fine to suggest lower case, even if you accept upper case.
```suggestion
    echo -e 'Target not specified or incorrect. Target value should be foreman or foreman-proxy \n' >&2
```

chris1984 (approved):
@patilsuraj767 looks a lot better thanks for addressing the comments. @ekohl your thoughts? Did you want him to rebase the pr or have Github do it?

bin/katello-certs-check
```diff
@@ -165,7 +171,7 @@ check-ca-bundle
 check-cert-san
 check-cert-usage-key-encipherment
 
-if [[ $EXIT_CODE == "0" ]] && [[ $CERT_HOSTNAME == $HOSTNAME ]]; then
+if [ $EXIT_CODE == "0" -a "$TARGET" == "foreman" ]; then
```
ekohl: I'm more fan of the bash conditions since they also handle unset parameters properly so they're a bit safer.
```suggestion
if [[ $EXIT_CODE == "0" ]] && [[ $TARGET == "foreman" ]]; then
```

bin/katello-certs-check
```diff
@@ -11,12 +11,15 @@ function usage () {
...
-usage: $0 -c CERT_FILE -k KEY_FILE -b CA_BUNDLE_FILE
+usage: $0 -t [foreman|foreman-proxy] -c CERT_FILE -k KEY_FILE -b CA_BUNDLE_FILE
 HELP
 }
 
-while getopts "c:k:b:" opt; do
+while getopts "t:c:k:b:" opt; do
     case $opt in
+        t)
+            TARGET=`echo $OPTARG|tr '[:upper:]' '[:lower:]'`
```
ekohl: More a fan of `$()` since you can nest those and it's easy to see where it begins and ends.
```suggestion
            TARGET=$(echo $OPTARG|tr '[:upper:]' '[:lower:]')
```

ekohl (commented):
Would it make sense to default to `foreman`? Only the instructions differ. You could even add a line to the Foreman scenario help text (line 192) stating:

```
  To use them inside a NEW \$FOREMAN_PROXY, rerun this command with -t foreman-proxy
```

patilsuraj767:
@ekohl I have made the required changes, please review the PR again.

ekohl (approved):

ekohl:
Thanks!

PR #355, opened 2019-05-21
Fixes #26623 - copy update certs hooks to katello-certs

evgeni:


.travis.yml
```diff
@@ -25,6 +25,7 @@ script:
   - bundle exec $INSTDIR/sbin/foreman-installer --help --scenario foreman-proxy-content
   - bundle exec $INSTDIR/sbin/foreman-installer --help --scenario katello
   - bundle exec $INSTDIR/sbin/foreman-proxy-certs-generate --help
+  - bundle exec $INSTDIR/sbin/foreman-proxy-certs-generate --help |grep -q certs-update-server
```
ekohl, line 28: I like this

Rakefile
```diff
@@ -290,6 +290,10 @@ task :install => :build do
   if CERTS_SCENARIOS.any?
     mkdir_p "#{DATADIR}/foreman-installer/katello-certs/scenarios.d"
     cp_r 'katello_certs/hooks', "#{DATADIR}/foreman-installer/katello-certs"
+    # 20-certs_update.rb lives in katello, but is *also* required in katello-certs
```
ekohl: I'm leaning to copying the files in git so there's a better overview there. We already copied other hooks.

ekohl (changes requested):

Rakefile
evgeni: Did we? I didn't find any duplicate files (only names)

ekohl: I guess I only copied relevant sections then. My worry is that nobody looks at the Rakefile and will have issues tracking down where a file came from. A relative symlink would also work and be easier to track down in development.

evgeni: Oh yeah, a symlink would work

evgeni: updated to be symlinks

ekohl (approved):
Haven't tested it, but it should work.

PR #360, opened 2019-05-23
Only perform mongodb-syspaths install on EL7

ehelms:


ekohl (approved):
I wonder if there are some pre checks that should also be disabled, but in itself this one is good I think.

PR #363, opened 2019-05-28
Lock puppet-strings to less than 2.0 if on Ruby 2.0

ehelms:


Gemfile
```diff
@@ -3,7 +3,11 @@ source 'https://rubygems.org'
 gem 'kafo', '>= 3.0.0'
 gem 'librarian-puppet'
 gem 'puppet', ENV.key?('PUPPET_VERSION') ? "~> #{ENV['PUPPET_VERSION']}" : '~> 6.0'
-gem 'puppet-strings'
+if RUBY_VERSION > '2.0'
```
mmoll: I think `if RUBY_VERSION < '2.1'` should be used here.

mmoll (changes requested):

mmoll:
`semverse` also needs a pinning for `< 2.1` and another one for `< 2.2` in my clean installs to work.

ehelms:
@mmoll I can add semverse -- what versions of ruby correpond to each of those?

mmoll:
`< 2` for Ruby `< 2.1` and `< 3` for Ruby `< 2.2`

ehelms:
@mmoll Updated, check me I got that right

Gemfile
```diff
@@ -14,5 +18,11 @@ end
 group :development do
   # Needed to pin dependencies
   gem 'puppet_forge'
-  gem 'semverse'
+  if RUBY_VERSION < '2.1'
+    gem 'semverse', '< 2'
+  elsif
```
mmoll: `elsif RUBY_VERSION < '2.2'`

mmoll (changes requested):

Gemfile
ehelms: Oops, makes sense!

ehelms:
Updated

mmoll (approved):

mmoll:
Works as expected for mw now, I guess this should get cherry-picked all the way down to 1.20?

ehelms:
I wasn't sure the implication for release branches. Right now this would fix a failure on CI.

mmoll:
Let's unblock that run now :+1: 
If the problem is also seen in older branches, anybody should feel free to cherry-pick as needed.

mmoll:
merged, thanks @ehelms!

PR #365, opened 2019-06-03
Fixes #26588 - Remove apache user from foreman group pre-install

tbrisker:


ekohl (changes requested):
Katello hooks are (for now) separated from Foreman hooks. Merging those is still on the roadmap so for now you need to duplicate code.

tbrisker:
Copied to the katello folder as well

tbrisker:
and now updated names to describe what it's actually doing

ekohl (approved):

PR #366, opened 2019-06-17
[WIP] Fixes #27069 - Bump versions for initial EL8 support

ekohl:
Within RH there's an internal effort (until CentOS 8 is out when we can
build upstream) to investigate running on RHEL8. A few modules have
received patches to allow the installation. This requires some minor
module updates, but they've been used in nightly and don't change
anything on EL7 so should be safe.

Note that this does not work upstream since there's no upstream EL8
packages.


mmoll:
tests fail :man_shrugging:

ekohl:
Added a cherry picked commit to fix the tests. I'm still in the process of testing this so marking it as WIP.

mmoll:
seeems to be fine now :+1:

ekohl:
There may be some issues that should get the `Fixed in 1.22.1` marking, like https://projects.theforeman.org/issues/26839. Haven't had time to test this yet though.

tbrisker:
What's the status here?

ekohl:
Closing since we decided not to push for EL8 support in 1.22 but rather do it as part of 1.23. Even though CentOS8 isn't out yet, a lot has been verified on RHEL8. Depending on when CentOS8 comes out, we'll either officially add support in 1.23 or 1.24.

PR #369, opened 2019-07-01
Allow easy branding of the foreman-proxy-certs-genereate tool

adamruzicka:
For example for Satellite it can be branded by dropping in a single file overriding the name methods.

```ruby
# katello_certs/hooks/boot/10-branding-override.rb

class Kafo::Helpers
  class << self
    def main_instance_name
      'Satellite'
    end

    def proxy_name
      'Capsule'
    end

    def installer_package
      'satellite-capsule'
    end

    def scenario_name
      'capsule'
    end

    def installer_command
      'satellite-instaler'
    end
  end
end
```

### DIFF
--- a/katello_certs/hooks/boot/02-message-helpers.rb
+++ b/katello_certs/hooks/boot/02-message-helpers.rb
@@ -1,5 +1,25 @@
 class Kafo::Helpers
   class << self
+    def main_instance_name
+      'Katello'
+    end
+
+    def proxy_name
+      'Smart Proxy'
+    end
+
+    def installer_package
+      'foreman-installer-katello'
+    end
+
+    def scenario_name
+      'foreman-proxy-content'
+    end
+
+    def installer_command
+      'foreman-installer'
+    end
+
     def proxy_instructions_message(kafo)
       fqdn = if kafo.param('foreman_proxy_certs', 'parent_fqdn')
                kafo.param('foreman_proxy_certs', 'parent_fqdn').value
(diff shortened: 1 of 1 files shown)


katello_certs/hooks/boot/02-message-helpers.rb
```diff
@@ -1,5 +1,25 @@
...
+    def installer_package
+      'foreman-installer-katello'
+    end
+
+    def scenario_name
+      'foreman-proxy-content'
+    end
+
+    def installer_command
+      'foreman-instaler'
```
ekohl:
```suggestion
      'foreman-installer'
```

ekohl (changes requested):
Makes sense

katello_certs/hooks/boot/02-message-helpers.rb
adamruzicka: Fixed, thanks for spotting this one

katello_certs/hooks/boot/02-message-helpers.rb
```diff
@@ -21,21 +41,22 @@ def proxy_instructions_message(kafo)
 
   To finish the installation, follow these steps:
 
-  If you do not have the smartproxy registered to the Katello instance, then please do the following:
+  If you do not have the #{proxy_name} registered to the #{main_instance_name} instance, then please do the following:
 
   1. yum -y localinstall http://#{fqdn}/pub/katello-ca-consumer-latest.noarch.rpm
   2. subscription-manager register --org "<%= color('#{org}', :info) %>"
 
-  Once this is completed run the steps below to start the smartproxy installation:
```
adamruzicka, line 29: Any idea why it originally said `smartproxy`? Was that a typo? Was the file additionally processed somehow?

ekohl: It may be well possible nobody read it that closely or cared enough to create an issue.

ekohl (approved):

PR #376, opened 2019-07-29
Fixes #27462 - Capture the openssl error correctly

hao-yu:
Older openssl version will still return 0 exit code and "OK" message
for some errors, such as the openssl in RHEL 7. This cause the
katello-certs-check script not capturing certificate error correctly.
This patch fixed the issue.

### DIFF
--- a/bin/katello-certs-check
+++ b/bin/katello-certs-check
@@ -128,11 +128,13 @@ function check-priv-key () {
 function check-ca-bundle () {
     printf "Checking CA bundle against the certificate file: "
     CHECK=$(openssl verify -CAfile $CA_BUNDLE_FILE -purpose sslserver -verbose $CERT_FILE 2>&1)
-    if [[ $? == "0" ]]; then
-        success
+
+    ERROR_PATTERN="error [0-9]+ at"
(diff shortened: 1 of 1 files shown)


ekohl:
This file is getting very complex, but we have a basic test for this file. Could you add a test case that covers this? https://github.com/theforeman/foreman-installer/blob/develop/spec/katello_certs_check_spec.rb is hopefully self documenting but I'll gladly help if you could provide some reproducer.

ekohl (changes requested):
Letting the bot pick up the review

ekohl:
@hao-yu have you looked at my comments? It'd be nice to have a proper reproducer for this.

hao-yu:
@ekohl Sorry, about the late reply. I will try to reproduce again and see if this issue is still valid in the latest Satellite.

ehelms:
Hi @hao-yu -- is this an issue you are planning to revisit?

hao-yu:
Sorry, for the delay. I will close this issue as I don't think I have time to write a test for it which might take me some time. Additionally I haven't seen anyone hitting this issue lately

PR #377, opened 2019-07-31
Update the pinning script

ekohl:
This is the version as it lived in katello-installer. It maintains the alignment making the diff easier to read.

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -1,4 +1,4 @@
-forge 'https://forgeapi.puppetlabs.com'
+forge 'https://forgeapi.puppetlabs.com/'
 
 # Dependencies
 mod 'puppetlabs/mysql',              '>= 4.0.0'
--- a/Rakefile
+++ b/Rakefile
@@ -12,28 +12,44 @@ begin
   require 'semverse'
 
   class FakePuppetfile
-    attr_reader :new_content
-
     def initialize
       @new_content = []
     end
 
     def forge(url)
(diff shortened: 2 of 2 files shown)


Puppetfile
```diff
@@ -1,4 +1,4 @@
-forge 'https://forgeapi.puppetlabs.com'
+forge 'https://forgeapi.puppetlabs.com/'
```
ekohl, line 1: This is what the script produces. Not sure if this is the best way or if we should update the script.

Rakefile
```diff
@@ -12,28 +12,44 @@ begin
...
+      else
+        @new_content << ['mod', name, options]
+      end
+    end
+
+    def content
+      max_length = @new_content.select { |type, value| type == 'mod' }.map { |type, value| value.length }.max
+
+      @new_content.each do |type, value, options|
+        if type == 'forge'
```
evgeni, line 39: is this code branch really worth it, or should be just have `"forge '#{url}'"` as a `header` variable and print that before looping over `content`?

evgeni (approved):
ack with a small question :)

Rakefile
ekohl: I implemented this in https://github.com/Katello/katello-installer/commit/0a42ab6b285613a964dd9071f70b89b5630a4691 and now just copied it. Technically you're right that it could be simplified.

evgeni: But no reason to block the PR on it :)

PR #379, opened 2019-08-07
Fixes #27540 - handle local pgsql start

upadhyeammit:
The upgrade tries to start local pgsql, mongodb services. To handle this one,
- The foreman-maintain is smart to check if service is local or external, so if system has foreman-maintain I am calling it to start service.
- When foreman-maintain is not available, I tried to handle two conditions,
  - pgsql and mongo are on external system, in this case I have checked if service file is there, if yes then only start/stop service.
  - pgsql and mongo are on local system, start or stop using the systemctl.

### DIFF
--- a/katello/hooks/boot/01-helpers.rb
+++ b/katello/hooks/boot/01-helpers.rb
@@ -44,6 +44,38 @@ class Kafo::Helpers
         $?.success?
       end
     end
+
+    def start_service(name)
+      return execute("foreman-maintain service start --only #{name}") if foreman_maintain?
+
+      if service_available?(name)
+        log_and_say :info, 'Expecting #{name} is managed by Foreman'
+        execute("systemctl start #{name}")
+      else
+        log_and_say :info, 'Expecting #{name} is not managed by Foreman, not trying to start'
+        true
+      end
+    end
+
+    def stop_service(name)
+      return execute("foreman-maintain service stop --only #{name}") if foreman_maintain?
+
+      if service_available?(name)
+        log_and_say :info, 'Expecting #{name} is managed by Foreman'
+        execute("systemctl stop #{name}")
+      else
+        log_and_say :info, 'Expecting #{name} is not managed by Foreman, not trying to stop'
+        true
+      end
+    end
+
+    def service_available?(name)
+      execute("which #{name}")
+    end
+
+    def foreman_maintain?
+      service_available?('foreman-maintain')
+    end
   end
 end
 
--- a/katello/hooks/boot/01-helpers.rb
+++ b/katello/hooks/boot/01-helpers.rb
@@ -44,38 +44,6 @@ class Kafo::Helpers
         $?.success?
       end
     end
-
-    def start_service(name)
-      return execute("foreman-maintain service start --only #{name}") if foreman_maintain?
-
-      if service_available?(name)
-        log_and_say :info, 'Expecting #{name} is managed by Foreman'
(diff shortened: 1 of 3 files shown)


katello/hooks/boot/01-helpers.rb
```diff
@@ -44,6 +44,38 @@ def execute_command(command, do_say, do_log)
...
+        log_and_say :info, 'Expecting #{name} is not managed by Foreman, not trying to stop'
+        true
+      end
+    end
+
+    def service_available?(name)
+      execute("which #{name}")
+    end
+
+    def foreman_maintain?
```
upadhyeammit: This method will be called multiple times, I am not sure where and how I can put it so it gets called only once for installer run. Any suggestions are welcome.

katello/hooks/pre/10-reset_feature.rb
```diff
@@ -79,11 +79,9 @@ def empty_mongo
   if remote_host?(mongo_config[:host])
     empty_remote_mongo(mongo_config)
   else
+    Kafo::Helpers.start_service('rh-mongodb34-mongod')
```
upadhyeammit: I have changed this to make it look unified change because this one is already having if guard and it only runs when service is running on foreman itself.

upadhyeammit:
If I see then pgsql and mongo can be external services so I have not changed other service calls, like for tomcat, pulp. If its wise to change it for all services then will add changes accordingly.
Thank You!

ekohl (changes requested):
What I don't like about this is the reliance on foreman-maintain. I'm aware that we already rely on it in some parts and I'm also unhappy about that.

One reason for this is that in the mongodb migration it should fail if it's unavailable because it can only work on a local machine.

I would much rather see that we look at the actual answers as the installer should. That means the line becomes:

```ruby
if param_value('foreman', 'db_manage') || param_value('katello', 'candlepin_manage_db')
  upgrade_step :start_postgresql, :run_always => true
end
```

It might also be good to add `--exclude postgresql` to the `stop_services` method.

upadhyeammit:
@ekohl I reverted changes. I did not skip pgsql stop because foreman-maintain handles it, if pgsql is external it wont stop it.
Thank you !

katello/hooks/pre/30-upgrade.rb
```diff
@@ -53,7 +53,7 @@ def mongo_mmapv1_check
       logger.info 'No changed needed, Mongo storage engine will installed/kept with WiredTiger'
     else
       # Stop Mongo 2.x
-      Kafo::Helpers.execute('systemctl stop mongod')
+      Kafo::Helpers.stop_service('mongod')
```
ekohl: Leftover here that can be reverted.

katello/hooks/pre/30-upgrade.rb
```diff
@@ -207,7 +207,9 @@ def fail_and_exit(message)
   upgrade_step :stop_services, :run_always => true
 
   if katello
```
ekohl: This now becomes redundant and you can drop this.

Edit: to be a bit less brief: I mean this exact line. What's inside the if statement should be sufficient

ekohl (changes requested):

upadhyeammit:
@ekohl removed the extra if condition, miss from my side! Thank You!

ekohl (approved):

tbrisker:
1.23 - b6a7147
1.22 - 0916d55

PR #386, opened 2019-09-04
Fixes #27784,#27785 - Remove redundant defaults from Katello + clean old migrations

ekohl:


ehelms (approved):
I'd like a core Katello developer to ACK this change. @jlsherrill can you take a look?

jlsherrill:
everything seems in order to me, that said i'm not super familiar with these files and their history

PR #387, opened 2019-09-06
Fixes #27775 - Introduce tuning for installation size

ekohl:
This introduces a --tuning parameter to Katello/Content proxies to optimize for a small/medium/large installation. To achieve this, it needs Kafo 4 which allows setting custom facts which are usable in the Hiera config.

### DIFF
--- a/config/foreman-proxy-content.migrations/190904144224-set_tuning_fact.rb
+++ b/config/foreman-proxy-content.migrations/190904144224-set_tuning_fact.rb
@@ -1,0 +1,2 @@
+scenario[:facts] = {} unless scenario.key?(:facts)
+scenario[:facts]['tuning'] ||= 'default'
--- a/config/foreman.hiera/tuning/sizes/extra-extra-large.yaml
+++ b/config/foreman.hiera/tuning/sizes/extra-extra-large.yaml
@@ -1,0 +1,17 @@
+---
+apache::mod::passenger::passenger_max_pool_size: 240
+apache::mod::passenger::passenger_max_request_queue_size: 1000
+
+apache::mod::prefork::serverlimit: 1024
+apache::mod::prefork::maxclients: 1024
+apache::mod::prefork::maxrequestsperchild: 4000
+
+candlepin::java_opts: "-Xms1024m -Xmx8192m"
+
+postgresql::server::config_entries:
+  max_connections: 1000
+  shared_buffers: 32GB
+  work_mem: 8MB
+  checkpoint_segments: 32
+  effective_cache_size: 32GB
+  autovacuum_vacuum_cost_limit: 2000
--- a/config/katello.migrations/190904144224-set_tuning_fact.rb
+++ b/config/katello.migrations/190904144224-set_tuning_fact.rb
@@ -1,0 +1,2 @@
+scenario[:facts] = {} unless scenario.key?(:facts)
+scenario[:facts]['tuning'] ||= 'default'
--- a/katello/hooks/boot/13-hiera.rb
+++ b/katello/hooks/boot/13-hiera.rb
@@ -1,0 +1,11 @@
+# TODO automatically get the tuning sizes - standard doesn't exist and only
+# loads base. The rest maps to config/foreman.hiera/tuning/sizes/$size.yaml
+TUNING_SIZES = ['default'] + ['medium', 'large', 'extra-large', 'extra-extra-large']
+TUNING_FACT = 'tuning'
+
+app_option(
+  '--tuning',
+  'INSTALLATION_SIZE',
+  "Tune for an installation size. Choices: #{TUNING_SIZES.join(', ')}",
+  :default => get_custom_fact(TUNING_FACT)
+)
--- a/config/foreman-hiera.yaml
+++ b/config/foreman-hiera.yaml
@@ -13,5 +13,6 @@ hierarchy:
       - "family/%{facts.os.family}-%{facts.os.release.major}.yaml"
       - "family/%{facts.os.family}.yaml"
       - "security.yaml"
-      - "tuning.yaml"
+      - "tuning/sizes/%{facts.kafo.scenario.custom.tuning}.yaml"
+      - "tuning/common.yaml"
       - "common.yaml"
--- a/config/foreman-proxy-content.yaml
+++ b/config/foreman-proxy-content.yaml
@@ -11,6 +11,8 @@
 :module_dirs: ./_build/modules
 :parser_cache_path: ./_build/parser_cache/foreman-proxy-content.yaml
 :hiera_config: ./config/foreman-hiera.yaml
+:facts:
+  tuning: 'default'
 
 :order:
   - certs
--- a/config/foreman.hiera/tuning/common.yaml
+++ b/config/foreman.hiera/tuning/common.yaml
@@ -11,6 +11,9 @@ apache::mod::passenger::passenger_max_requests: 10000
 apache::mod::passenger::passenger_min_instances: 1
 apache::mod::passenger::passenger_max_preloader_idle_time: 0
 
+qpid::open_file_limit: 65536
+qpid::router::open_file_limit: 150100
+
 postgresql::server::config_entries:
   checkpoint_completion_target: 0.9
   max_connections: 500
--- a/config/foreman.hiera/tuning/sizes/extra-large.yaml
+++ b/config/foreman.hiera/tuning/sizes/extra-large.yaml
@@ -1,0 +1,17 @@
+---
+apache::mod::passenger::passenger_max_pool_size: 120
+apache::mod::passenger::passenger_max_request_queue_size: 1000
+
+apache::mod::prefork::serverlimit: 1024
+apache::mod::prefork::maxclients: 1024
+apache::mod::prefork::maxrequestsperchild: 4000
+
+candlepin::java_opts: "-Xms1024m -Xmx8192m"
+
+postgresql::server::config_entries:
+  max_connections: 1000
+  shared_buffers: 16GB
+  work_mem: 8MB
+  checkpoint_segments: 32
+  effective_cache_size: 32GB
+  autovacuum_vacuum_cost_limit: 2000
--- a/config/foreman.hiera/tuning/sizes/large.yaml
+++ b/config/foreman.hiera/tuning/sizes/large.yaml
@@ -1,0 +1,14 @@
+---
+apache::mod::passenger::passenger_max_pool_size: 60
+apache::mod::passenger::passenger_max_request_queue_size: 1000
+
+apache::mod::prefork::serverlimit: 1024
+apache::mod::prefork::maxclients: 1024
+apache::mod::prefork::maxrequestsperchild: 4000
+
+postgresql::server::config_entries:
+  max_connections: 1000
+  shared_buffers: 8GB
+  checkpoint_segments: 32
+  effective_cache_size: 16GB
+  autovacuum_vacuum_cost_limit: 2000
--- a/config/foreman.hiera/tuning/sizes/medium.yaml
+++ b/config/foreman.hiera/tuning/sizes/medium.yaml
@@ -1,0 +1,14 @@
+---
+apache::mod::passenger::passenger_max_pool_size: 30
+apache::mod::passenger::passenger_max_request_queue_size: 1000
+
+apache::mod::prefork::serverlimit: 1024
+apache::mod::prefork::maxclients: 1024
+apache::mod::prefork::maxrequestsperchild: 4000
+
+postgresql::server::config_entries:
+  max_connections: 1000
+  shared_buffers: 4GB
+  checkpoint_segments: 32
+  effective_cache_size: 16GB
+  autovacuum_vacuum_cost_limit: 2000
--- a/config/katello.yaml
+++ b/config/katello.yaml
@@ -11,6 +11,8 @@
 :module_dirs: ./_build/modules
 :parser_cache_path: ./_build/parser_cache/katello.yaml
 :hiera_config: ./config/foreman-hiera.yaml
+:facts:
+  tuning: 'default'
 
 :order:
   - certs
--- a/katello/hooks/pre_commit/13-hiera.rb
+++ b/katello/hooks/pre_commit/13-hiera.rb
@@ -1,0 +1,17 @@
+TUNING_SIZES = ['default'] + ['medium', 'large', 'extra-large', 'extra-extra-large']
+TUNING_FACT = 'tuning'
+
+current = get_custom_fact(TUNING_FACT)
+new = app_value(:tuning)
+
+if current != new
+  unless TUNING_SIZES.include?(new)
+    say "<%= color('Invalid tuning profile', :bad) %>"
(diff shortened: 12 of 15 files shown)


ekohl:
Needs https://github.com/theforeman/kafo/pull/241 so still a draft.

ekohl:
On a fresh nightly with then this patch + kafo 4 applied:

```
[root@centos7-katello-nightly foreman-installer]# foreman-installer --help | grep tuning
    --tuning INSTALLATION_SIZE    Tune for an installation size. Choices: small, medium, large (default: "medium")
```

Then applying it:

```
[root@centos7-katello-nightly foreman-installer]# foreman-installer --tuning large --noop --verbose --disable-system-checks
# ....
[ WARN 2019-09-06T16:03:53 verbose]  /Stage[main]/Candlepin::Config/File[/etc/tomcat/tomcat.conf]/content: 
[ WARN 2019-09-06T16:03:53 verbose] --- /etc/tomcat/tomcat.conf	2019-09-06 14:21:23.722259817 +0000
[ WARN 2019-09-06T16:03:53 verbose] +++ /tmp/puppet-file20190906-18723-1fvsorg	2019-09-06 16:03:53.700071508 +0000
[ WARN 2019-09-06T16:03:53 verbose] @@ -30,7 +30,7 @@
[ WARN 2019-09-06T16:03:53 verbose]  CATALINA_TMPDIR="/var/cache/tomcat/temp"
[ WARN 2019-09-06T16:03:53 verbose]  
[ WARN 2019-09-06T16:03:53 verbose]  # You can pass some parameters to java here if you wish to
[ WARN 2019-09-06T16:03:53 verbose] -JAVA_OPTS="-Xms1024m -Xmx4096m"
[ WARN 2019-09-06T16:03:53 verbose] +JAVA_OPTS="-Xms1024m -Xmx8192m"
[ WARN 2019-09-06T16:03:53 verbose]  
[ WARN 2019-09-06T16:03:53 verbose]  # You can change your tomcat locale here
[ WARN 2019-09-06T16:03:53 verbose]  
```

ehelms:
The tuning sizes are hardcoded into the installer code right? What is the workflow for adding a new or custom tuning profile?

ekohl:
That's a good point and also one I'm still thinking about. Currently I'm leaning to globbing the paths so it can be dropped in. It may also make sense to add the scenario name in the path (`tuning/%{facts.kafo.scenario.name}/%{facts.kafo.scenario.custom.tuning}.yaml` as an additional layer) for scenarios to override values. A Katello large might be different than a Foreman Proxy Content large. Kafo 4 allows that.

katello/hooks/boot/13-hiera.rb
```diff
@@ -0,0 +1,15 @@
+TUNING_SIZES = ['small', 'medium', 'large']
```
ehelms: You would set this value based on the profiles found in the directory?

ekohl: Correct. And probably also always add `default` as an option.

ekohl:
Locally I tested this by doing a nightly install and then running:
```
yum -y install https://ekohl.nl/{foreman-installer{,-katello}-1.24.0-0.2.develop.20190906155222git13a7ce4,rubygem-kafo-4.0.0-1}.el7.noarch.rpm
```

ehelms:
Would tuning always have a value? That is, would a basic installation enable the small profile out of th e box?

ekohl:
Currently this is done in the migration via `scenario[:facts]['tuning'] ||= 'medium'`

ekohl:
At the end of last week I made sure Kafo 4 was in nightly and nightly was green. This PR now has literal copies of the recommendations from Satellite.

There are some things that need attention:
* Naming. The ones I chose now differ from the Redmine issue. I made a slight modification because think `xl` vs `xxl` is easy to confuse. Looking for feedback here
* Some options. They have TODOs in the YAML files. It's `apache::mod::passenger::passenger_max_requests` where the recommended value is 1/10th of what we ship by default. There's also `log_min_duration_statement` for postgres which is half of the default. I'm leaning to leaving both out as I trust think the current defaults are ok. Also looking for feedback here
* Input validation. When a user passes in an unknown tuning size, it silently accepts it.
* Testing. I haven't tested the latest iteration with the released kafo 4 in nightly.

Not a blocker, but ideally it would autodetect the tuning sizes from existing files.

ehelms:
```
Input validation. When a user passes in an unknown tuning size, it silently accepts it.
```

I don't expect this to be something that happens often, but if its an easy fix I think giving some indication to the user they gave an non-existent profile would be nice.

ekohl:
https://github.com/theforeman/foreman-installer/pull/387/files#diff-5e9660452f40a63fa483698b97426409R10-R14 doesn't appear to work. This is probably easy to work around by creating a `pre_validations` hook that looks at the value. That can be added later on after the basic functionality is verified :)

config/foreman-proxy-content.migrations/190904144224-set_tuning_fact.rb
```diff
@@ -0,0 +1 @@
+scenario[:facts]['tuning'] ||= 'small'
```
ehelms: So we default to small, which is interesting since we'd almost expect that to be the 'defaults' in the installer rather than a profile. It makes me wonder from a naming stand point if we should be calling `small` something like `standard` or `base`.

ekohl: I like standard.

sthirugn: I agree. `small` name will cause confusion as it doesn't really mean a small install. I think we should just call it `default` which is easier to understand. Thoughts?

katello/hooks/boot/13-hiera.rb
```diff
@@ -0,0 +1,16 @@
+# TODO automatically get the tuning sizes
+TUNING_SIZES = ['small', 'medium', 'large', 'extra-large', 'extra-extra-large']
```
sthirugn: Adding to my earlier comment in this PR, how about the following? This will make it in line with the [tuning profiles upstream](https://github.com/RedHatSatellite/satellite-support/tree/master/tuning-profiles).

`
TUNING_SIZES = ['default', 'medium', 'large', 'extra-large', '2-extra-large']
`

config/foreman.hiera/tuning/sizes/extra-extra-large.yaml
```diff
@@ -0,0 +1,22 @@
+---
+apache::mod::passenger::passenger_max_pool_size: 240
+apache::mod::passenger::passenger_max_request_queue_size: 1000
+# TODO this is 1/10th of the base
```
sthirugn: Good point, do you know how you got 10000 for this value for the base? That seems high.  If you can tell me how you go it, I will work on creating issues to get this fixed.

sthirugn (changes requested):

katello/hooks/boot/13-hiera.rb
ehelms: I lean towards `extra-extra-large` as that follows on `extra-large`.

config/foreman-proxy-content.migrations/190904144224-set_tuning_fact.rb
ehelms: The slightly ironic thing is, I would expect default to be 0 tuning, and then some sort of standard tuning profile except that I would then ask as a user why its not the default. So I can live with either name :)

ekohl: Default makes sense as well. Probably slightly more. At least both default and standard are better options than small.

katello/hooks/boot/13-hiera.rb
sthirugn: I mentioned `2-extra-large` because it felt like `extra` was redundant in `extra-extra-large` and also [satellite-support](https://github.com/RedHatSatellite/satellite-support/blob/master/tuning-profiles/custom-hiera-2ex-large-256G.yaml) uses `2ex` convention so users may be already familiar with it.

I am ok with either though.

ehelms:
> https://github.com/theforeman/foreman-installer/pull/387/files#diff-5e9660452f40a63fa483698b97426409R10-R14 doesn't appear to work. This is probably easy to work around by creating a `pre_validations` hook that looks at the value. That can be added later on after the basic functionality is verified :)

ACK -- let's open a tracking issue for this.

config/katello.migrations/190904144224-set_tuning_fact.rb
```diff
@@ -0,0 +1 @@
+scenario[:facts]['tuning'] ||= 'small'
```
ehelms: Based on the test failures, this needs to check that `scenario[:facts]` exists prior to trying to set a property of it. Also, Ruby makes me sad sometimes with the mixing of symbols and strings.

ehelms:
Most issues you identified have been covered with either a "lets create a tracking issue" or an answer to work from. Once the tests are fixed, is there anything blocking this?

ekohl:
No, currently I'm going over the installer modules so RC1 can go out on Monday without me being present. I hope to finish that work tomorrow and have time to finish this.

config/foreman.hiera/tuning/sizes/extra-extra-large.yaml
```diff
@@ -0,0 +1,22 @@
...
+apache::mod::prefork::maxrequestsperchild: 4000
+
+candlepin::java_opts: "-Xms1024m -Xmx8192m"
+
+postgresql::server::config_entries:
+  max_connections: 1000
+  shared_buffers: 32GB
+  work_mem: 8MB
+  checkpoint_segments: 32
+  checkpoint_completion_target: 0.9
```
ekohl: I'll drop this line since it matches the base value.

config/foreman.hiera/tuning/sizes/extra-extra-large.yaml
```diff
@@ -0,0 +1,22 @@
...
+apache::mod::prefork::maxclients: 1024
+apache::mod::prefork::maxrequestsperchild: 4000
+
+candlepin::java_opts: "-Xms1024m -Xmx8192m"
+
+postgresql::server::config_entries:
+  max_connections: 1000
+  shared_buffers: 32GB
+  work_mem: 8MB
+  checkpoint_segments: 32
```
ekohl, line 15: This will be problematic on PostgreSQL 9.5 and newer. That has dropped the setting and will prevent the server from starting up. Since only EL7 ships such an ancient version, we didn't include this setting in base. Adding this here will prevent users from using any SCL.

config/foreman.hiera/tuning/sizes/extra-extra-large.yaml
ekohl: b570b53b8e7df5cb0fdff30029a7e0c1a4c60dff added it. Initially 100000 was suggested in the ticket but per https://projects.theforeman.org/issues/18821#note-2 I asked a trusted user who suggested 10000.

ekohl:
I've pushed an updated version:
* Rebased
* Fix tests
* Strip out duplicate/TODO parameters leaving those for a future update
* Attempt to do validation without `signal_usage_error` but explicit logger/error

I'll apply this on a system and see if it works.

config/foreman.hiera/tuning/sizes/extra-extra-large.yaml
ehelms: When we do a Postgres upgrade, we will need to ensure we drop this from the tuning then?

ekohl: Correct. We are also knowingly breaking user setups who have chosen to use SCLs, even if it's not officially supported.

@sthirugn does this really have a large performance impact?

ekohl: Looks like https://wiki.postgresql.org/wiki/Tuning_Your_PostgreSQL_Server does strongly suggest this is a good thing.

ekohl:
Locally verified:
* Correct validation
* Store the scenario to disk when the tuning fact changes value

katello/hooks/boot/13-hiera.rb
```diff
@@ -0,0 +1,11 @@
+# TODO automatically get the tuning sizes - standard doesn't exist and only
+# loads base. The rest maps to config/foreman.hiera/tuning/sizes/$size.yaml
```
ehelms, line 2: Can we capture this TODO in a Redmine issue? Also the text is already out of date :) Should be "default doesn't exist"

ehelms:
Is there a design reason behind having base and default be the same thing but named differently? That seems confusing to developers and users.

ekohl:
> Is there a design reason behind having base and default be the same thing but named differently? That seems confusing to developers and users.

Base is always loaded and all other tuning is applied on top of that. So default = base + null and medium = base + medium. That's why in `TUNING_SIZES` I kept it to `['default'] + [...]` so that `[...]` would become a `listdir('/path/to/sizes')` when properly implemented. Technically you could create an empty `default.yaml` in sizes which would be more explicit.

ehelms:
That makes sense at a technical level. At a higher, slightly more user or sys admin centric level why not name base and default the same thing? I am not seeing where we need different names for the same concept. But I may be missing a nuance.

ekohl:
I could name `base` something like `common` to indicate it's shared. Would that make more sense?

ehelms:
My issue is that base is default and default is base. So why have the difference between the two? Thats the design decision that I am missing. Do we expect default to include more than base at some point? If not, it's less confusing to me to just name them the same thing: either base or default.

ekohl:
The key point is that `base` is **always** loaded. It is also used when you select large. That's why it's on a different level. Default is essentially "don't apply any additional tuning". If I create a `default.yaml`, we'll still load `base.yaml` and apply the values not specified in `default.yaml`.

ehelms:
I'm sorry, but that is still confusing :) We are essentially saying base is the `default` set of values. And I as a user and developer have to unpack a naming nuance. I am not seeing a scenario where base is not the default. Where I would have something conceptually known as the default that is *not* the base. I can let this go, but I am already confused as an admin. This is how I am looking at it if it helps to at least clarify my nit.

Someone will come along and ask "What are the default tuning values?" And our answer will be "Look at the base tuning config". To me the answer should be "Look at the default tuning config". Or we keep the answer the same "Look at the base tuning config" and change the question to "What are the base tuning values?".

ekohl:
The last commit renames `base.yaml` to `common.yaml`. This matches the `common.yaml` from a level above which is also loaded for every setup. We can also move `tuning/common.yaml` (formerly `tuning/base.yaml`) into `common.yaml`.

ehelms (approved):

ekohl:
Squashed the commits and added the `---` header to `medium.yaml`. Other than that it's the same commit.

config/foreman.hiera/tuning/sizes/extra-extra-large.yaml
sthirugn: yes, this setting is recommended

PR #390, opened 2019-09-12
Fixes #27848 - Disable HTTP by default on Katello's proxy

ekohl:
The HTTP and templates features are useful when a host needs to be provisioned and can't reach the Foreman server directly. Since this is by definition the Foreman server, the client should be able to reach it.

The values are dropped since they match the module defaults. It doesn't implement a migration to not touch existing installations which might have started to rely on it.

### DIFF
--- a/config/katello-answers.yaml
+++ b/config/katello-answers.yaml
@@ -28,9 +28,7 @@ puppet:
   server_foreman_ssl_cert: /etc/pki/katello/puppet/puppet_client.crt
   server_foreman_ssl_key: /etc/pki/katello/puppet/puppet_client.key
 foreman_proxy:
-  http: true
   ssl_port: '9090'
-  templates: true
   ssl_ca: /etc/foreman-proxy/ssl_ca.pem
   ssl_cert: /etc/foreman-proxy/ssl_cert.pem
   ssl_key: /etc/foreman-proxy/ssl_key.pem
--- a/config/katello.migrations/171030135054-foreman-proxy-update.rb
+++ b/config/katello.migrations/171030135054-foreman-proxy-update.rb
@@ -1,4 +1,0 @@
-# foreman_proxy defaults for templates
-
-answers['foreman_proxy']['http'] = true if answers['foreman_proxy']
-answers['foreman_proxy']['templates'] = true if answers['foreman_proxy']


ekohl:
@lzap you requested this to be turned on in https://projects.theforeman.org/issues/19720 but I don't see why this is needed. Why would you need to proxy on the same installation?

lzap:
Provisioning template proxying was a feature asked by PM, they wanted to have it turned on by default so we are opinionated about it. That's the context.

Now, in that ticked I complained about that when you turn on template feature on a proxy and HTTP port is not set, the feature is rendered useless. However that's also true for other features so I don't think we need to change this behavior. It just needs to be documented properly.

Not sure what this patch is all about tho, isn't this a regression to users who haven't touched this setting (and the next installer run after upgrade would shut the port down for them)?

ekohl:
> Provisioning template proxying was a feature asked by PM, they wanted to have it turned on by default so we are opinionated about it. That's the context.

Do you have any source about this? I don't see any benefit of this setup and we got a BZ where the security people complained they needed to open an additional unencrypted port.

> Not sure what this patch is all about tho, isn't this a regression to users who haven't touched this setting (and the next installer run after upgrade would shut the port down for them)?

It drops the migration so users upgrading from very old answers will never switch it on, just like users on a fresh installation. However, it doesn't touch existing answers so if it's on, it will stay on.

ehelms:
@lzap Any feedback to help us here?

lzap:
> Do you have any source about this? I don't see any benefit of this setup and we got a BZ where the security people complained they needed to open an additional unencrypted port.

I am referring to priority feature from 6.1: "all communication between managed nodes and server must go through proxy". I am under impression that this hasn't changed. That's also why template proxy plugin was/is installed by default.

ekohl:
Do you have more context about that feature and the goal? If a host runs both Foreman and Foreman Proxy then you don't isolate anything, you just expose an additional service via an unencrypted connection. I suspect the actual goal of that was the ability to run hosts in a DMZ where there's only a Foreman Proxy. That's what the `foreman-proxy-content` scenario is for, not `katello`. That installs templates by default already and is untouched by this.

ekohl:
IMHO should either merge this or go all the way and drop Foreman's HTTP vhost altogether. https://community.theforeman.org/t/removing-unattended-url/16700

ehelms (approved):

PR #393, opened 2019-09-17
Fixes #27876 - Add check for passphrase on key

chris1984:
Results of testing:
```bash
[root@centos7-katello-nightly ~]# katello-certs-check -b /root/certs/ourCA.crt -k /root/certs/privkey.pem -c /root/certs/server.crt 
Checking server certificate encoding: 
[OK]

Checking expiration of certificate: 
[OK]

Checking expiration of CA bundle: 
[OK]

Checking if server certificate has CA:TRUE flag 
[OK]

Checking for private key passphrase: 
[FAIL]

The /root/certs/privkey.pem contains a passphrase, please recreate the csr and key without a passphrase.
```

Certs to test with are in the redmine issue


ekohl (changes requested):
Could you add a test for this? https://github.com/theforeman/foreman-installer/blob/develop/spec/katello_certs_check_spec.rb is still quite minimal but guarantees it continues to work.

chris1984:
@ekohl not sure if this is right, but added a test.

bin/katello-certs-check
```diff
@@ -114,6 +114,17 @@ function check-cert-ca-flag () {
     fi
 }
 
+function check-passphrase () {
+    printf "Checking for private key passphrase: "
+    CHECK=$(cat $KEY_FILE | grep ENCRYPTED)
+    if [[ $? == "0" ]]; then
+        error 2 "The $KEY_FILE contains a passphrase, please recreate the csr and key without a passphrase."
```
ekohl: Wouldn't it be easier to remove the passphrase from the key?
```
cp -a "$KEY_FILE" "$KEY_FILE".old
openssl rsa -in "$KEY_FILE".old -out "$KEY_FILE"
```

spec/katello_certs_check_spec.rb
```diff
@@ -15,4 +15,11 @@ def fixture(filename)
     expect(stdout).to eq ''
     expect(status.exitstatus).to eq 1
   end
+
+  it 'with password on key' do
+    stdout, stderr, status = Open3.capture3(command)
```
ekohl: You probably need to modify the command to add a filename and store the file in git or generate it on the fly.

ekohl (changes requested):

bin/katello-certs-check
chris1984: Yeah, I will do that, I see you can do that in the manpage here:

https://www.openssl.org/docs/man1.1.1/man1/rsa.html

I should have looked at that first. I will do that and change it to a warning letting the user know since we want to be transparent.

ekohl: Yes, instructing the user is probably a better approach because it's safer.

chris1984:
@ekohl added a test for a 'good run' and then added mine with the fixtures, I am confused why it is adding `/n` and erroring out on that during the spec tests. Can you provide some feedback when you get a chance please? I think i am really close

spec/fixtures/katello-certs-check/foreman.example.com.crt
```diff
@@ -0,0 +1,84 @@
+Certificate:
+    Data:
+        Version: 1 (0x0)
+        Serial Number: 1048577 (0x100001)
+    Signature Algorithm: sha256WithRSAEncryption
+        Issuer: O=My Company/emailAddress=dev@community.theforeman.org, L=My Town, ST=State or Providence, C=US, CN=foremanca.example.com
```
ekohl: Probably better to use an `example.com` email address as well.

spec/fixtures/katello-certs-check/passphrase-onkey.txt
```diff
@@ -0,0 +1 @@
+The /home/travis/build/theforeman/foreman-installer/spec/fixtures/katello-certs-check/key_pass.key contains a passphrase, please recreate the key without a passphrase by doing: cp -a /home/travis/build/theforeman/foreman-installer/spec/fixtures/katello-certs-check/key_pass.key /home/travis/build/theforeman/foreman-installer/spec/fixtures/katello-certs-check/key_pass.key.old openssl rsa -in /home/travis/build/theforeman/foreman-installer/spec/fixtures/katello-certs-check/key_pass.key.old -out /home/travis/build/theforeman/foreman-installer/spec/fixtures/katello-certs-check/key_pass.key
```
ekohl: These absolute paths will mean that a developer can never have passing builds. Probably best to copy the pattern of replacing `|COMMAND|` but then for the `directory`.

spec/fixtures/katello-certs-check/passphrase-onkey.txt
```diff
@@ -0,0 +1,2 @@
+The /home/travis/build/theforeman/foreman-installer/spec/fixtures/katello-certs-check/key_pass.key contains a passphrase, please recreate the key without a passphrase by doing:
+cp -a /home/travis/build/theforeman/foreman-installer/spec/fixtures/katello-certs-check/key_pass.key /home/travis/build/theforeman/foreman-installer/spec/fixtures/katello-certs-check/key_pass.key.old openssl rsa -in /home/travis/build/theforeman/foreman-installer/spec/fixtures/katello-certs-check/key_pass.key.old -out /home/travis/build/theforeman/foreman-installer/spec/fixtures/katello-certs-check/key_pass.key
```
ekohl: This probably needs indenting similar to the normal command output.

chris1984:
@ekohl updated with green tests. I could not put variables in the fixture file so just asserted with dir variables like you talked about.

bin/katello-certs-check
```diff
@@ -114,6 +114,18 @@ function check-cert-ca-flag () {
     fi
 }
 
+function check-passphrase () {
+    printf "Checking for private key passphrase: "
+    CHECK=$(cat $KEY_FILE | grep ENCRYPTED)
+    if [[ $? == "0" ]]; then
+        error 2 "The $KEY_FILE contains a passphrase, please recreate the key without a passphrase by doing:
+        cp -a $KEY_FILE $KEY_FILE.old openssl rsa -in $KEY_FILE.old -out $KEY_FILE"
```
ekohl: Probably sufficient to just move the file. It's also two separate commands.
```suggestion
        mv $KEY_FILE $KEY_FILE.old
        openssl rsa -in $KEY_FILE.old -out $KEY_FILE"
```

bin/katello-certs-check
```diff
@@ -114,6 +114,18 @@ function check-cert-ca-flag () {
     fi
 }
 
+function check-passphrase () {
+    printf "Checking for private key passphrase: "
+    CHECK=$(cat $KEY_FILE | grep ENCRYPTED)
+    if [[ $? == "0" ]]; then
+        error 2 "The $KEY_FILE contains a passphrase, please recreate the key without a passphrase by doing:
```
ekohl:
```suggestion
        error 2 "The $KEY_FILE contains a passphrase. Remove key's passphrase by doing:
```

spec/katello_certs_check_spec.rb
```diff
@@ -15,4 +22,18 @@ def fixture(filename)
...
+    command_with_certs = "#{command} -b #{ca} -k #{key} -c #{cert}"
+    stdout, stderr, status = Open3.capture3(command_with_certs)
+    expect(stderr).to eq ''
+    expect(status.exitstatus).to eq 0
+  end
+
+  it 'with password on key' do
+    command_with_certs = "#{command} -b #{ca} -k #{badkey} -c #{cert}"
+    stdout, stderr, status = Open3.capture3(command_with_certs)
+    expect(stderr).to eq "The #{badkey} contains a passphrase, please recreate the key without a passphrase by doing: cp -a #{badkey} #{badkey}.old openssl rsa -in #{badkey}.old -out #{badkey}\n"
```
ekohl: I'd expect a newline after `:`

ekohl (changes requested):
This looks very close.

chris1984:
@ekohl updated and tests green still. I had to add `echo -e` on the fail function since it was not adding new lines on the actual script, instead just doing `\n` in the text.

chris1984:
Here is the update output:

```bash
[15:26:59] {root@satellite1} ~
$ ./katello-certs-check -b cacert.crt -k key_pass.key -c foreman.example.com.crt
Checking server certificate encoding: 
[OK]

Checking expiration of certificate: 
[OK]

Checking expiration of CA bundle: 
[OK]

Checking if server certificate has CA:TRUE flag 
[OK]

Checking for private key passphrase: 
[FAIL]

The /root/key_pass.key contains a passphrase, please recreate the key without a passphrase by doing: 
cp -a /root/key_pass.key /root/key_pass.key.old 
openssl rsa -in /root/key_pass.key.old -out /root/key_pass.key
```

ekohl (approved):

PR #394, opened 2019-09-18
Add a script to show module versions

ekohl:
This is a helper script to easily read the module versions from a lockfile. This makes it much easier what is bundled in a certain installer branch. The branch is a guess and might be incorrect. For example, when it's still being released from master.

### DIFF
--- a/bin/module-versions
+++ b/bin/module-versions
@@ -1,0 +1,27 @@
+#!/usr/bin/env ruby
+
+require 'librarian/puppet'
+
+def manifests(lockfile)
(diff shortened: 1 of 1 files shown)


ehelms:
Will this need an RPM/Deb spec update to omit this ?

ekohl:
My understanding is that the Rakefile specifically copies what actually should be there. In the same way that we don't include `bin/create-migration` even though it's nowhere in the ignores.

ehelms (approved):

PR #395, opened 2019-09-18
Fixes #27884 - No hooks after packages are locked

mbacovsky:
Hooks may want to install packages prevented by version locking. Suggest with the "99' that locking should be last hook in the 'post' group


mbacovsky:
@chris1984, could you please confirm this move is safe?

chris1984:
@mbacovsky it looks safe, I am assuming you want it to run as the very last step?

ekohl (approved):

ekohl:
LGTM, but will let @chris1984 and @mbacovsky agree before merging.

mbacovsky:
@chris1984 correct, we need to lock packages as a last step so it does not prevent other steps (e.g mongo-syspath) from installing things.

chris1984 (approved):
Makes sense, ACK from Me. @ekohl if you can merge please.

PR #397, opened 2019-09-23
Refs #27847 - Set parameter mapping

ekohl:
In puppet-certs master the default values are loaded via params.pp to workaround kafo limitations. This configures kafo to actually find it.  Otherwise it can't load the defaults and the values are nil.

### DIFF
--- a/katello_certs/config/foreman-proxy-certs.yaml
+++ b/katello_certs/config/foreman-proxy-certs.yaml
@@ -20,6 +20,9 @@
   :foreman_proxy_certs:
     :manifest_name: foreman_proxy_content
     :dir_name: certs
+    :params_name: foreman_proxy_content::params
+    # https://projects.theforeman.org/issues/27909
+    :params_path: certs/manifests/foreman_proxy_content/params.pp
 :order:
   - certs
   - foreman_proxy_certs


ekohl:
It can't generate the defaults but not sure why.
```
$ bundle exec $INSTDIR/sbin/foreman-proxy-certs-generate --help

Could not get default values, check log file at /tmp/tmp.oueXiC0vj1/var/log/foreman-installer/foreman-proxy-certs-generate.log for more information

The command "bundle exec $INSTDIR/sbin/foreman-proxy-certs-generate --help" exited with 25.

10.67s$ bundle exec $INSTDIR/sbin/foreman-proxy-certs-generate --help |grep -q certs-update-server

The command "bundle exec $INSTDIR/sbin/foreman-proxy-certs-generate --help |grep -q certs-update-server" exited with 1.
```

ekohl:
```
Error: Evaluation Error: Operator '[]' is not applicable to an Undef Value. (file: /tmp/tmp.sAudVOhpMm/share/foreman-installer/modules/certs/manifests/foreman_proxy_content/params.pp, line: 9, column: 18)
```

The problem is `$facts['networking']['fqdn']` which is the modern fact, but that requires facter 3. In bundler we can only have facter 2 which means `$facts['networking']` is `undef`. `undef['fqdn']` doesn't work.

ekohl:
:green_heart: due to https://github.com/theforeman/puppet-certs/pull/260 being merged.

PR #398, opened 2019-10-17
Refs #27775 - Update to Kafo 4

ekohl:


evgeni (approved):

evgeni:
this should be followed by a packaging update of the installer to update the requires.

PR #399, opened 2019-10-17
Fixes #28084 - Store non-default answers separate

ekohl:
By storing the non-default answers in a different layer, resetting works as expected. This relies on Kafo 4.

Currently a draft since the migrations do not like this.

Includes https://github.com/theforeman/foreman-installer/pull/398

### DIFF
--- a/config/foreman-answers.yaml
+++ b/config/foreman-answers.yaml
@@ -11,18 +11,22 @@
 ---
 foreman:
   configure_epel_repo: false
-foreman_proxy: {}
-puppet:
-  server: true
 foreman::cli: true
 foreman::cli::ansible: false
 foreman::cli::azure: false
 foreman::cli::discovery: false
+foreman::cli::kubevirt: false
 foreman::cli::openscap: false
 foreman::cli::remote_execution: false
 foreman::cli::tasks: false
 foreman::cli::templates: false
-foreman::cli::kubevirt: false
+foreman::compute::ec2: false
+foreman::compute::gce: false
+foreman::compute::libvirt: false
+foreman::compute::openstack: false
+foreman::compute::ovirt: false
+foreman::compute::rackspace: false
+foreman::compute::vmware: false
 foreman::plugin::ansible: false
 foreman::plugin::azure: false
 foreman::plugin::bootdisk: false
@@ -32,29 +36,23 @@ foreman::plugin::dhcp_browser: false
 foreman::plugin::digitalocean: false
 foreman::plugin::discovery: false
 foreman::plugin::expire_hosts: false
-foreman::plugin::kubevirt: false
+foreman::plugin::hooks: false
 foreman::plugin::host_extra_validator: false
+foreman::plugin::kubevirt: false
 foreman::plugin::memcache: false
 foreman::plugin::monitoring: false
 foreman::plugin::omaha: false
 foreman::plugin::openscap: false
 foreman::plugin::ovirt_provision: false
-foreman::plugin::tasks: false
-foreman::plugin::hooks: false
 foreman::plugin::puppetdb: false
 foreman::plugin::remote_execution: false
 foreman::plugin::remote_execution::cockpit: false
 foreman::plugin::salt: false
 foreman::plugin::setup: false
 foreman::plugin::snapshot_management: false
+foreman::plugin::tasks: false
 foreman::plugin::templates: false
-foreman::compute::ec2: false
-foreman::compute::gce: false
-foreman::compute::libvirt: false
-foreman::compute::openstack: false
-foreman::compute::ovirt: false
-foreman::compute::rackspace: false
-foreman::compute::vmware: false
+foreman_proxy: {}
 foreman_proxy::plugin::ansible: false
 foreman_proxy::plugin::chef: false
 foreman_proxy::plugin::dhcp::infoblox: false
@@ -69,3 +67,5 @@ foreman_proxy::plugin::openscap: false
 foreman_proxy::plugin::pulp: false
 foreman_proxy::plugin::remote_execution::ssh: false
 foreman_proxy::plugin::salt: false
+puppet:
+  server: true
--- a/config/foreman-answers.yaml
+++ b/config/foreman-answers.yaml
@@ -9,8 +9,7 @@
 #
 # See params.pp in each class for what options are available
 ---
-foreman:
-  configure_epel_repo: false
+foreman: true
 foreman::cli: true
 foreman::cli::ansible: false
 foreman::cli::azure: false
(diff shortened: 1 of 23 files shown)


ekohl:
Needs a rebase now.

config/foreman-answers.yaml
```diff
@@ -1,27 +1,20 @@
-# Format:
```
ekohl: Kafo always adds this header so probably good to keep it in and also add it to katello.

ehelms:
So with this change, the answers file should never change as a user tweaks their installation? Further, the implication is that a user should never touch this files by hand ever?

ekohl:
No, the idea is that you can disable everything, then enable it again and you'll get the proper scenario defaults again. Without this, if you disable a module and then enable it again you will get the Puppet module defaults again. That also means we can start with a disabled module where the scenario defaults differ from the Puppet module defaults.

ekohl:
This should now pass. It's seen some rebases so be careful in reviewing that I didn't miss any parameters. It's also fairly large so I split off https://github.com/theforeman/foreman-installer/pull/475.

config/foreman-hiera.yaml
```diff
@@ -8,6 +8,8 @@ hierarchy:
     path: "custom.yaml"
   - name: "Kafo Answers"
     path: "%{facts.kafo.scenario.answer_file}"
+  - name: "Scenario defaults"
+    path: "scenario/%{facts.kafo.scenario.id}.yaml"
```
ehelms, line 12: One downside to this approach is that the scenario must provide this file here or add it to the installed location for this repository. If the scenario is stored in it's own repository and delivered in its own RPM that is a bit uglier.

What are your thoughts on letting this be driven through the scenario's configuration file via  variable that denotes what file to look in or through a convention that introduces a new file e.g. `katello-defaults.yaml`

ekohl: Thinking out loud now.

Rather than another file, we could also store it in the scenario itself. That lets the package deal with it. The downside is that you now need migrations where with a standalone file you can just drop the file in place.

Another is using the custom fact storage to point to some other file. I now realize I didn't document this in Kafo's README.md, but https://github.com/theforeman/kafo/commit/414bc6d2c21ecf1424c08589848a3f1668377da3 should be straight forward enough. Then in foreman-hiera.yaml you can add a layer `%{facts.kafo.scenario.custom.overrides}` or similar.

However, this indirection does add a lot of complexity where dropping the single file in `/usr/share/foreman-installer/config/scenario/my_own_scenario.yaml` is much more straight forward. That means it's easier to follow and debug which makes it still my preference.

On a side note: longer term I'd prefer for all scenarios to be installed to `/usr/share/foreman-installer/scenarios.d`, including addons. For that I'd like to write a small helper script that takes a scenario file, an answers file, migrations directory and defaults file to place it in the right layout. This helper should be provided by foreman-installer.

ehelms:
I was thinking about what the answers file is reduced to with this change and it starts to get at an issue that I've had with the scenarios but not quite been able to express.

The 'answers' file now becomes quite 'dumb' per scenario, for example the katello answers (https://github.com/theforeman/foreman-installer/pull/399/files#diff-737238c186c4e959d92f46872b5adae5) are just a list of available modules. Strikes me that there could just be a master list of all available modules, and the scenario configuration would include:

 * turning them on
 * custom configuration for each
 * disabling a module's availability

Today, the latter is implied by not including the module in the answers file. This leads to duplication and omitting some modules in some scenarios or forgetting to add them in both. I'm not advocating for this PR being the change in this area but it does clarify this a bit. Scenario's should just be pre-defined set of defaults that create a "spin" of an installation. This would make them much simpler and possibly more "stackable" on top of a default rather than "merging" per say.

This also leads me back to another thing that has nagged at me as this sort of helps enable it. I always look at it as we should have two major installation types:

 * Foreman server
 * External Foreman Proxy

Scenarios then should just be some combination of defaults applied on top of one of those types to achieve your pre-defined "spin".

I got a bit off topic of this specific PR, it does shed light on this idea.

ekohl:
I think I may be getting what you're hinting at, but at this point it's very impractical IMHO because the answers file should have all the answers expanded. If you want to add in parts, then the combinations quickly explode and becomes hard to deal with. In the past we've always shifted this burden to the end user since we couldn't come up with a way to do it properly.

The problem is that if you install a Foreman server. Now let's say you have a scenario Remote Execution. If such a scenario needs to modify a parameter in a base module/scenario, it becomes very tricky.

A more concrete example. Let's say you want to install Infoblox. We might create a scenario Infoblox DHCP and Infoblox DNS as another since not everyone runs both.

In the current design, that's essentially `--enable-foreman-proxy-plugin-dhcp-infoblox`. However, you also need to set `--foreman-proxy-dhcp true --foreman-proxy-dhcp-provider infoblox` (and until 2.0 also `--foreman-proxy-dhcp-managed false`).

Doing this via an addon hiera file feels very complex because the layering gets very complicated. If the installer already ran, then you probably need to add a layer above the answers file to force it, but the installer has no code to hide the parameter which makes the UX of the installer unpredictable. Users expect that if an option is present, it performs a function.

It is probably much better to implement this via hooks and this is something I've been thinking about. I've been thinking of an easy way to "subscribe" to class enablement/disablement. Something like:
```ruby
on_class_enabled 'foreman_proxy::plugin::dhcp::infoblox' do
  require_class 'foreman_proxy'
  set_answer 'foreman_proxy', 'dhcp', true
  set_answer 'foreman_proxy', 'dhcp_provider', 'infoblox'
end
```

The helpers:
```ruby
def require_class(name)
  return if answers[name]
  raise "Class #{name} is required"
end

def set_answer(cls, parameter, value)
  answers[name] = {} unless answers[name].is_a?(Hash)
  answers[name][parameter] = value
end
```

Disabling classes is another complex matter. In the smart proxy case this is doable in some cases but you can instruct users in other cases. For example that disabling a plugin doesn't actually remove it or removal is not supported. With these class listener hooks we have much more flexibility to achieve this.

Another thing I've been thinking about is completely getting rid of the default answers file and only relying on migrations. Right now I think that's a step too far, but if we start to treat the answers file as a database it makes sense. On the other hand, I never liked the Rails/ActiveRecord approach of not having an authoritative source of how it should look like. This would solve the problem that we end up with `.rpmsave` files in `/etc/foreman-installer` though.

ekohl:
I've discovered that on a fundamental level this doesn't work with the current Kafo since it will still read the defaults from params and stores that in the answers, rather than looking at a lower Hiera level. This means it's fundamentally impossible to ship non-default answers for modules that are disabled. A fix in Kafo is needed.

PR #400, opened 2019-10-22
Refs #28067 - defaults for dynflow sidekiq redis

ezr-ondrej:


config/foreman.hiera/family/RedHat.yaml
```diff
@@ -1,3 +1,5 @@
 ---
 # Disable additional Phusion Passenger repo on EL, preferring EPEL/Foreman's
 apache::mod::passenger::manage_repo: false
+redis::params::default_install: false
```
ekohl: I think we do want the default install. We count on that in the pulpcore module
```suggestion
```

config/foreman.hiera/family/RedHat.yaml
```diff
@@ -1,3 +1,5 @@
...
+redis::params::manage_repo: false
```
ekohl: This should be set on the redis class itself
```suggestion
redis::manage_repo: false
```

ekohl (changes requested):

ehelms:
This PR and https://github.com/theforeman/foreman-installer/pull/434/files are attempting to do the same thing. However, I think based on comments #434 is more correct at this point.

ezr-ondrej:
Closing in favor of #434

PR #402, opened 2019-10-29
Expand the create-migration script

ekohl:
This allows creating multiple migrations at the same time following the correct directory conventions. This makes it easy to create the same migration for multiple scenarios at the same time. For this the argument order needs to be swapped. It also accepts the code on STDIN to make it
a single operation.

### DIFF
--- a/bin/create-migration
+++ b/bin/create-migration
@@ -1,31 +1,38 @@
 #!/usr/bin/env ruby
 
-if ARGV.length != 2
-  $stderr.puts "Usage #{$0} DIRECTORY MIRGATION_NAME"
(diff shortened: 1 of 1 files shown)


ehelms (approved):

ekohl:
https://github.com/theforeman/foreman-installer/blob/develop/config/foreman.migrations/20190423154755_remove-docker-abrt-plugins.rb and https://github.com/theforeman/foreman-installer/blob/develop/config/foreman.migrations/20190903120000_remove-cockpit-plugin.rb are both examples why this is useful. Note I'm to blame for at least one ;)

mmoll:
merged, bedankt @ekohl!

PR #405, opened 2019-11-01
Add azure_rm plugin

apuntamb:


tbrisker (commented):
This is failing because of missing `azure_rm` plugin in https://github.com/theforeman/puppet-foreman/tree/master/manifests/plugin - considering azure plugin is no longer supported, should we replace it with azure_rm?

apuntamb:
> This is failing because of missing `azure_rm` plugin in https://github.com/theforeman/puppet-foreman/tree/master/manifests/plugin - considering azure plugin is no longer supported, should we replace it with azure_rm?

Ahh Okk, but since our documentation has azure (older one) too, should we create a file there with azure_rm.pp? Or you think edit the existing one? Also, is there a script to create a new file under plugins directory?

ekohl:
I think replacing the plugin we install in `puppet-foreman` is fine. It is the proper replacement.

apuntamb:
> I think replacing the plugin we install in `puppet-foreman` is fine. It is the proper replacement.

How does this look: [puppet-foreman/767](https://github.com/theforeman/puppet-foreman/pull/767)?

apuntamb:
@ekohl @tbrisker Since we are using the older azure installer option here: [puppet-foreman/767](https://github.com/theforeman/puppet-foreman/pull/767), should we close this PR?

ekohl:
Yes, we can close this.

PR #406, opened 2019-11-19
Drop Katello upgrade steps for Mongo and Qpid certs

ehelms:
There have been 10 releases since the qpidd upgrade step was introduced and 8 since we upgraded Mongo. Therefore, I feel fairly confident we can drop these from our current upgrade steps.

@jlsherrill @chris1984


chris1984 (approved):
Yeah most people should be off Mongo 2.x now, ACK for the Mongo side of things

chris1984:
I wrote the qpid piece too, so I think that is good to go as well, looking at this again.

katello/hooks/post/09-mongo_syspath.rb
```diff
@@ -1,17 +0,0 @@
-def install_syspaths
-  if facts[:os][:release][:major] == '7'
-    `rpm -qa | grep rh-mongodb34-syspaths`
-    if $?.success?
-      logger.debug 'rh-mongodb34-syspaths already installed, skipping.'
-    else
-      logger.info 'rh-mongodb34-syspaths not present, installing.'
-      `yum install -y -q rh-mongodb34-syspaths`
```
ekohl, line 8: I think we still need to ensure this is installed. IIRC that can be done by setting the `mongodb::globals::server_package_name` to that in `config/foreman.hiera/family/RedHat-7.yaml`

ekohl (changes requested):
I'm inclined to also trim down `upgrade_step` to remove tracking of all steps since it now always calls it with `:run_always => true`. We can also do that in a later commit. If we need it in the future, I actually want to use the mechanisms the installer has built in rather than this custom solution.

ehelms:
Right now, I am planning to add a step for Postgres upgrade but I thought cleaning up first would be nice.

ekohl (approved):
Let's have a look at how the postgres one turns out. However, I'd like to use the native mechanism (`store_custom_config` and `get_custom_config`). See https://github.com/theforeman/foreman-installer/blob/5676453ec3226dff221cc1589a7118042fcdb843/katello/hooks/pre_commit/09-version_locking.rb on how to do that.

Of course that's all for the next PR and I'm now happy with this cleanup :)

ekohl:
Of course just now I realize that this could have used a Redmine issue to track. Please do one for the postgres upgrade :)

PR #407, opened 2019-11-20
Upgrade Postgres to 10 on Red Hat systems

ehelms:
Just a starting place for the upgrade workflow and approach. The upgrade this way *must* happen after the migrate Candlepin and Foreman steps which is a bit ugly. Changing this requires me to move these into being handled by the puppet modules / installation steps.

### DIFF
--- a/katello/hooks/boot/01-helpers.rb
+++ b/katello/hooks/boot/01-helpers.rb
@@ -16,6 +16,11 @@ class Kafo::Helpers
       Kafo::KafoConfigure.logger.send(level, message) if do_log
     end
 
+    def fail_and_exit(message)
+      log_and_say :error, message
+      exit 1
+    end
+
     def read_cache_data(param)
       YAML.load_file("/opt/puppetlabs/puppet/cache/foreman_cache_data/#{param}")
     end
@@ -47,13 +52,15 @@ class Kafo::Helpers
   end
 end
 
-# FIXME: remove when #23332 is released
 module HookContextExtension
+  # FIXME: remove when #23332 is released
   def param_value(mod, name)
     param(mod, name).value if param(mod, name)
   end
+
+  def local_postgresql?
+    param_value('foreman', 'db_manage') || param_value('katello', 'candlepin_manage_db')
+  end
 end
 
-unless Kafo::HookContext.method_defined?(:param_value)
-  Kafo::HookContext.send(:include, HookContextExtension)
-end
+Kafo::HookContext.send(:include, HookContextExtension)
--- a/katello/hooks/pre/30-upgrade.rb
+++ b/katello/hooks/pre/30-upgrade.rb
@@ -37,6 +37,15 @@ def migrate_foreman
   Kafo::Helpers.execute('foreman-rake db:migrate')
 end
 
+def postgresql_10_upgrade
+  stop_services
+  Kafo::Helpers.execute('yum -y install rh-postgresql10-postgresql-server')
+  Kafo::Helpers.execute('scl enable rh-postgresql10 -- postgresql-setup --upgrade')
+  Kafo::Helpers.execute('yum -y remove postgresql postgresql-server')
+  Kafo::Helpers.execute('rm -f /etc/systemd/system/postgresql.service')
+  Kafo::Helpers.execute('yum -y install rh-postgresql10-syspaths')
+end
+
 def upgrade_step(step, options = {})
   noop = app_value(:noop) ? ' (noop)' : ''
   long_running = options[:long_running] ? ' (this may take a while) ' : ''
@@ -46,7 +55,7 @@ def upgrade_step(step, options = {})
     Kafo::Helpers.log_and_say :info, "Upgrade Step: #{step}#{long_running}#{noop}..."
     unless app_value(:noop)
       status = send(step)
-      fail_and_exit "Upgrade step #{step} failed. Check logs for more information." unless status
+      Kafo::Helpers.fail_and_exit "Upgrade step #{step} failed. Check logs for more information." unless status
       touch_step(step)
     end
   end
@@ -65,11 +74,6 @@ def step_path(step)
   File.join(STEP_DIRECTORY, step.to_s)
 end
 
-def fail_and_exit(message)
-  Kafo::Helpers.log_and_say :error, message
-  kafo.class.exit 1
-end
-
 if app_value(:upgrade)
   Kafo::Helpers.log_and_say :info, 'Upgrading, to monitor the progress on all related services, please do:'
   Kafo::Helpers.log_and_say :info, '  foreman-tail | tee upgrade-$(date +%Y-%m-%d-%H%M).log'
@@ -80,7 +84,7 @@ if app_value(:upgrade)
 
   upgrade_step :stop_services, :run_always => true
 
-  if param_value('foreman', 'db_manage') || param_value('katello', 'candlepin_manage_db')
+  if local_postgresql?
     upgrade_step :start_postgresql, :run_always => true
   end
 
@@ -91,6 +95,10 @@ if app_value(:upgrade)
   if katello
     upgrade_step :migrate_candlepin, :run_always => true
     upgrade_step :migrate_foreman, :run_always => true
+  end
+
+  if local_postgresql? && facts[:os][:release][:major] == '7'
+    upgrade_step :postgresql_10_upgrade
   end
 
   Kafo::Helpers.log_and_say :info, 'Upgrade Step: Running installer...'
--- a/config/foreman-hiera.yaml
+++ b/config/foreman-hiera.yaml
@@ -10,6 +10,7 @@ hierarchy:
     path: "%{facts.kafo.scenario.answer_file}"
   - name: "Built in"
     paths:
+      - "scenario/%{facts.kafo.scenario.id}/family/%{facts.os.family}-%{facts.os.release.major}.yaml"
       - "family/%{facts.os.family}-%{facts.os.release.major}.yaml"
       - "family/%{facts.os.family}.yaml"
       - "security.yaml"
--- a/config/foreman.hiera/scenario/foreman-proxy-content/RedHat-7.yaml
+++ b/config/foreman.hiera/scenario/foreman-proxy-content/RedHat-7.yaml
@@ -1,0 +1,9 @@
+---
+postgresql::globals::version: '10'
+postgresql::globals::client_package_name: rh-postgresql10-postgresql-syspaths
+postgresql::globals::server_package_name: rh-postgresql10-postgresql-server-syspaths
+postgresql::globals::contrib_package_name: rh-postgresql10-postgresql-contrib-syspaths
+postgresql::globals::service_name: postgresql
+postgresql::globals::datadir: /var/opt/rh/rh-postgresql10/lib/pgsql/data
+postgresql::globals::confdir: /var/opt/rh/rh-postgresql10/lib/pgsql/data
+postgresql::globals::bindir: /usr/bin
--- a/config/foreman.hiera/scenario/katello/RedHat-7.yaml
+++ b/config/foreman.hiera/scenario/katello/RedHat-7.yaml
@@ -1,0 +1,9 @@
+---
+postgresql::globals::version: '10'
+postgresql::globals::client_package_name: rh-postgresql10-postgresql-syspaths
+postgresql::globals::server_package_name: rh-postgresql10-postgresql-server-syspaths
+postgresql::globals::contrib_package_name: rh-postgresql10-postgresql-contrib-syspaths
+postgresql::globals::service_name: postgresql
+postgresql::globals::datadir: /var/opt/rh/rh-postgresql10/lib/pgsql/data
+postgresql::globals::confdir: /var/opt/rh/rh-postgresql10/lib/pgsql/data
+postgresql::globals::bindir: /usr/bin
--- a/katello/hooks/pre_validations/30-mongo_storage_engine.rb
+++ b/katello/hooks/pre_validations/30-mongo_storage_engine.rb
@@ -1,8 +1,7 @@
 MONGO_DIR = '/var/lib/mongodb/'
 
 def fail_and_exit(message)
-  Kafo::Helpers.log_and_say :error, message
-  kafo.class.exit 1
+  Kafo::Helpers.fail_and_exit message
 end
 
 def disk_space
--- a/katello/hooks/pre_validations/32-upgrade.rb
+++ b/katello/hooks/pre_validations/32-upgrade.rb
@@ -1,0 +1,28 @@
+def check_postgresql_storage
+  #Ensure there is at least 1x /var/lib/pgsql free disk space available on the main filesystem
+  current_postgres_dir = '/var/lib/pgsql'
+  new_postgres_dir = '/var/opt/rh/rh-postgresql10/lib/pgsql'
+
+  begin
+    postgres_size = `du -b -s #{current_postgres_dir}`.split[0].to_i
+
+    if available_space(new_postgres_dir) < postgres_size
+      Kafo::Helpers.fail_and_exit "The postgres upgrade requires at least #{(postgres_size / 1024) / 1024} MB of storage."
+    end
+  rescue StandardError
+    Kafo::Helpers.fail_and_exit 'Failed to verify available disk space'
+  end
+end
+
+def available_space(directory = nil)
+  directory = '/' if directory.nil?
+  mountpoints = facts[:mountpoints]
+  until (mountpoint = mountpoints[directory.to_sym])
+    directory = File.dirname(directory)
+  end
+  mountpoint[:available_bytes]
+end
+
+if app_value(:upgrade)
+  check_postgresql_storage if local_postgresql?
+end


ekohl (commented):
If we put it in hiera this way, we also need to upgrade it on vanilla Foreman.

We also need to check if the database is local which means checking the `manage_database` parameter on the foreman class and katello's candlepin. For foreman you also need to make sure it's only ran on EL 7.

katello/hooks/pre/30-upgrade.rb
```diff
@@ -91,6 +102,7 @@ def fail_and_exit(message)
   if katello
     upgrade_step :migrate_candlepin, :run_always => true
     upgrade_step :migrate_foreman, :run_always => true
+    upgrade_step :postgresql_10_upgrade
```
ekohl, line 101: Can you move this to line 95 so we perform the DB migration before other migrations?

ehelms: See https://github.com/theforeman/foreman-installer/pull/407#issue-343739127

config/foreman.hiera/scenario/katello.yaml
```diff
@@ -0,0 +1,9 @@
+---
```
ehelms: Is this the correct way to handle this file? This worked when I tested it so I assume yes :)

katello/hooks/pre/30-upgrade.rb
ekohl: Why is that needed? Regardless, it needs the same if conditions (and `if katello` isn't needed anymore then). Otherwise you break remote database support.

config/foreman.hiera/scenario/katello.yaml
ekohl: In itself it's correct. I am slightly tempted to make it a bit more complex and go for `scenario/%{facts.kafo.scenario.id}/family/%{facts.os.family}-%{facts.os.release.major}.yaml` so we don't attempt to do this on EL8. Maybe that's just premature optimization though.

katello/hooks/pre/30-upgrade.rb
```diff
@@ -37,6 +37,17 @@ def migrate_foreman
   Kafo::Helpers.execute('foreman-rake db:migrate')
 end
 
+def postgresql_10_upgrade
+  #User should perform a backup
+  #Ensure there is at least 1x /var/lib/pgsql free disk space available on the main filesystem
+  Kafo::Helpers.execute('yum -y install rh-postgresql10-postgresql-server')
+  Kafo::Helpers.execute('foreman-maintain service stop')
+  Kafo::Helpers.execute('scl enable rh-postgresql10 -- postgresql-setup --upgrade')
+  Kafo::Helpers.execute('yum -y remove postgresql*')
```
ekohl: This also removes `postgresql-libs` which is a dependency of `tfm-rubygem-postgresql`, right? Probably also the `python3-psycopg2` package.

katello/hooks/pre/30-upgrade.rb
```diff
@@ -37,6 +37,29 @@ def migrate_foreman
...
 
+def postgresql_10_upgrade
+  #Ensure there is at least 1x /var/lib/pgsql free disk space available on the main filesystem
+  POSTGRES_DIR = '/var/lib/pgsql'
+
+  begin
+    available_space = `df -k --total --exclude-type=tmpfs --output=avail`.split("\n")[-1].to_i
+    postgres_size = File.directory?(POSTGRES_DIR) ? `du -k -s #{POSTGRES_DIR}`.split[0].to_i : 0
+    if available_space < postgres_size
+      $stderr.puts "The postgres upgrade requires at least #{postgres_size} of storage."
```
ekohl: This is in KB
```suggestion
      $stderr.puts "The postgres upgrade requires at least #{postgres_size} KB of storage."
```

katello/hooks/pre/30-upgrade.rb
```diff
@@ -37,6 +37,29 @@ def migrate_foreman
...
+      kafo.class.exit 1
+    end
+  rescue StandardError
+    $stderr.puts 'Failed to verify available disk space'
```
ekohl: I'm wondering if this should be a fatal error as well.

katello/hooks/pre/30-upgrade.rb
```diff
@@ -37,6 +37,29 @@ def migrate_foreman
...
+  end
+
+  Kafo::Helpers.execute('yum -y install rh-postgresql10-postgresql-server')
+  Kafo::Helpers.execute('foreman-maintain service stop')
+  Kafo::Helpers.execute('scl enable rh-postgresql10 -- postgresql-setup --upgrade')
+  Kafo::Helpers.execute('yum -y remove postgresql postgresql-server')
+  Kafo::Helpers.execute('rm -f /etc/systemd/system/postgresql.service')
```
ekohl, line 45: It surprises me that `yum remove postgresql-server` doesn't clean this up.

ehelms: Cause it gets overwritten by the module:

```
[root@centos7-katello-nightly-pg10 vagrant]# rpm -qlf /etc/systemd/system/postgresql.service 
file /etc/systemd/system/postgresql.service is not owned by any package
```

ekohl: Oh that's ugly. That should be using drop in files.

katello/hooks/pre/30-upgrade.rb
```diff
@@ -37,6 +37,30 @@ def migrate_foreman
...
+def postgresql_10_upgrade
+  #Ensure there is at least 1x /var/lib/pgsql free disk space available on the main filesystem
+  POSTGRES_DIR = '/var/lib/pgsql'
+
+  begin
+    available_space = `df -k --total --exclude-type=tmpfs --output=avail`.split("\n")[-1].to_i
+    postgres_size = File.directory?(POSTGRES_DIR) ? `du -k -s #{POSTGRES_DIR}`.split[0].to_i : 0
+    if available_space < postgres_size
+      $stderr.puts "The postgres upgrade requires at least #{postgres_size} KB of storage."
+      kafo.class.exit 1
```
ekohl: Since https://github.com/theforeman/kafo/commit/64810c08a0fe08d5ef76b20144ce709ee52e6f9f I think this can just be `exit` (and below as well):

```suggestion
      exit 1
```

ekohl (commented):
You should ensure `foreman-proxy-content.yaml` has the same content as `katello.yaml` once we add pulpcore. Can be a symlink. Currently we define it in the module but I'd like to also define it via hiera.

katello/hooks/pre/30-upgrade.rb
```diff
@@ -37,6 +37,28 @@ def migrate_foreman
...
+    postgres_size = File.directory?(POSTGRES_DIR) ? `du -k -s #{POSTGRES_DIR}`.split[0].to_i : 0
+    if available_space < postgres_size
+      fail_and_exit 'The postgres upgrade requires at least #{postgres_size} KB of storage.'
+    end
+  rescue StandardError
+    fail_and_exit 'Failed to verify available disk space'
+  end
+
+  Kafo::Helpers.execute('yum -y install rh-postgresql10-postgresql-server')
+  Kafo::Helpers.execute('foreman-maintain service stop')
```
ekohl: Wondering whether this should reuse `stop_services`

katello/hooks/pre/30-upgrade.rb
```diff
@@ -93,5 +115,9 @@ def fail_and_exit(message)
     upgrade_step :migrate_foreman, :run_always => true
   end
 
+  if param_value('foreman', 'db_manage') || param_value('katello', 'candlepin_manage_db')
```
ekohl: Probably good to add a variable `has_local_postgresql` and reuse that.

ehelms:
I went ahead and updated to do `scenario/<name>/family/<os-name>-<version>.yaml` since EL8 isn't that far off and this prevents having to re-do this part later.

There are improvements to the overall upgrade workflow we've mentioned but I would like to not block the upgrade effort on them. My preferred workflow would be:

 * merge this when it feels robust to get nightly onto new postgres and upgrade workflow
 * tackle other upgrade issues for Foreman 2.0
 * use fixing upgrade issues to handle forced postgres upgrade for Foreman scenario

The use of the `syspaths` RPMs should mean we don't have to touch foreman-maintain.

katello/hooks/pre/30-upgrade.rb
```diff
@@ -37,6 +37,28 @@ def migrate_foreman
   Kafo::Helpers.execute('foreman-rake db:migrate')
 end
 
+def postgresql_10_upgrade
+  #Ensure there is at least 1x /var/lib/pgsql free disk space available on the main filesystem
+  POSTGRES_DIR = '/var/lib/pgsql'
+
+  begin
+    available_space = `df -k --total --exclude-type=tmpfs --output=avail`.split("\n")[-1].to_i
```
ekohl: I think this should get the disk space available on `/var/opt/rh/rh-postgresql10/lib/pgsql/data` (or the closest path that does exist). Total disk space isn't quite correct.
```ruby
new_postgres_dir = '/var/opt/rh/rh-postgresql10/lib/pgsql/data'
until File.directory?(new_postgres_dir)
  new_postgres_dir = File.dirname(new_postgres_dir)
end
available_space = `df -k --output=avail "#{new_postgres_dir}"`.split("\n").last.to_i
```

In `katello/hooks/pre_validations/30-mongo_storage_engine.rb` there's similar code so it may make sense to create a helper for this at some point :)

katello/hooks/pre/30-upgrade.rb
```diff
@@ -37,6 +37,28 @@ def migrate_foreman
...
+  rescue StandardError
+    fail_and_exit 'Failed to verify available disk space'
+  end
+
+  Kafo::Helpers.execute('yum -y install rh-postgresql10-postgresql-server')
+  Kafo::Helpers.execute('foreman-maintain service stop')
+  Kafo::Helpers.execute('scl enable rh-postgresql10 -- postgresql-setup --upgrade')
+  Kafo::Helpers.execute('yum -y remove postgresql postgresql-server')
+  Kafo::Helpers.execute('rm -f /etc/systemd/system/postgresql.service')
+  Kafo::Helpers.execute('yum -y install rh-postgresql10-syspaths')
```
ekohl, line 46: Since this is a pre hook, the installer runs after it. That ensures all `-syspaths` are installed so is this really needed? Does `rh-postgresql10-syspath` even exist as a package?

ekohl (commented):
Overall this looks very good. I'd like to also upgrade vanilla Foreman before 2.0 hits, but I'd be happy to experiment with upgrading just katello first given the immediate need there.

katello/hooks/pre/30-upgrade.rb
ekohl: To make it really EL-8 ready you can add `facts[:os][:release][:major] == '7'` as a condition.

katello/hooks/pre/30-upgrade.rb
ehelms: We want syspaths for the services and the client bits so I think using the overall package is ok.

```
rh-postgresql10-syspaths.x86_64 : System-wide wrappers for the rh-postgresql10
```

katello/hooks/pre/30-upgrade.rb
```diff
@@ -37,6 +37,43 @@ def migrate_foreman
...
+  stop_services
+  Kafo::Helpers.execute('yum -y install rh-postgresql10-postgresql-server')
+  Kafo::Helpers.execute('scl enable rh-postgresql10 -- postgresql-setup --upgrade')
+  Kafo::Helpers.execute('yum -y remove postgresql postgresql-server')
+  Kafo::Helpers.execute('rm -f /etc/systemd/system/postgresql.service')
+  Kafo::Helpers.execute('yum -y install rh-postgresql10-syspaths')
+end
+
+def available_space(directory = nil)
+  `df -k --total --exclude-type=tmpfs --output=avail #{directory}`.split("\n")[-1].to_i
```
ekohl: `--exclude-type=tmpfs` is technically not really effective because you ask it for a specific directory. Same for `--total`.

katello/hooks/pre/30-upgrade.rb
```diff
@@ -37,6 +37,43 @@ def migrate_foreman
...
 
+def postgresql_10_upgrade
+  return unless facts[:os][:release][:major] == '7'
+
+  #Ensure there is at least 1x /var/lib/pgsql free disk space available on the main filesystem
+  current_postgres_dir = '/var/lib/pgsql'
+  new_postgres_dir = '/var/opt/rh/rh-postgresql10/lib/pgsql'
+
+  begin
+    unless File.exist?(new_postgres_dir)
```
ekohl: Any reason you're not using my code? My version detected the separate partition `/var` for example where yours doesn't.

katello/hooks/pre/30-upgrade.rb
ekohl: Alternative implementation that doesn't require shelling out:
```ruby
def available_bytes(directory)
  mountpoints = facts[:mountpoints]
  until (mountpoint = mountpoints[directory])
    directory = File.dirname(directory)
  end
  mountpoint[:available_bytes]
end
```

katello/hooks/pre/30-upgrade.rb
ehelms: I was not clear from looking at your code `how` it detected the separate partition without erroring when it did not exist _and_ then checked the base filesystem.

ekohl: It started with the directory and went up until the directory actually exists. You can give `df` any directory and it'll find the correct mountpoint as long as it exists. If you keep going up, you'll end up with `/` and that always exists:
```
[1] pry(main)> File.dirname('/doesnotexist')
=> "/"
```

ehelms: Not on my systems, which is part of why I didn't go that route:

```
irb(main):001:0> File.dirname('/var/foo/var/this/I/am/not/real')
=> "/var/foo/var/this/I/am/not"
```

ekohl: That's why there was the `until` loop :)
```
[1] pry(main)> directory = '/var/foo/var/this/I/am/not/real'
=> "/var/foo/var/this/I/am/not/real"
[2] pry(main)> until File.directory?(directory)
[2] pry(main)*   directory = File.dirname(directory)
[2] pry(main)* end  
=> nil
[3] pry(main)> puts directory
/var
=> nil
```

katello/hooks/pre/30-upgrade.rb
ehelms: Using this requires passing `available_bytes(:/)` if the directory does not exist? Just understanding how this works for the default case:

e.g
```
{:/=>{:available=>"34.04 GiB", :available_bytes=>36553474048, :capacity=>"13.26%", :device=>"/dev/vda1", :filesystem=>"ext4", :options=>["rw", "seclabel", "relatime", "data=ordered"], :size=>"39.25 GiB", :size_bytes=>42140401664, :used=>"5.20 GiB", :used_bytes=>5586927616}, :"/dev"=>{:available=>"3.81 GiB", :available_bytes=>4092948480, :capacity=>"0%", :device=>"devtmpfs", :filesystem=>"devtmpfs", :options=>["rw", "seclabel", "nosuid", "size=3997020k", "nr_inodes=999255", "mode=755"], :size=>"3.81 GiB", :size_bytes=>4092948480, :used=>"0 bytes", :used_bytes=>0}, :"/dev/hugepages"=>{:available=>"0 bytes", :available_bytes=>0, :capacity=>"100%", :device=>"hugetlbfs", :filesystem=>"hugetlbfs", :options=>["rw", "seclabel", "relatime"], :size=>"0 bytes", :size_bytes=>0, :used=>"0 bytes", :used_bytes=>0}, :"/dev/mqueue"=>{:available=>"0 bytes", :available_bytes=>0, :capacity=>"100%", :device=>"mqueue", :filesystem=>"mqueue", :options=>["rw", "seclabel", "relatime"], :size=>"0 bytes", :size_bytes=>0, :used=>"0 bytes", :used_bytes=>0}, :"/dev/pts"=>{:available=>"0 bytes", :available_bytes=>0, :capacity=>"100%", :device=>"devpts", :filesystem=>"devpts", :options=>["rw", "seclabel", "nosuid", "noexec", "relatime", "gid=5", "mode=620", "ptmxmode=000"], :size=>"0 bytes", :size_bytes=>0, :used=>"0 bytes", :used_bytes=>0}, :"/dev/shm"=>{:available=>"3.82 GiB", :available_bytes=>4100939776, :capacity=>"0.00%", :device=>"tmpfs", :filesystem=>"tmpfs", :options=>["rw", "seclabel", "nosuid", "nodev"], :size=>"3.82 GiB", :size_bytes=>4100960256, :used=>"20.00 KiB", :used_bytes=>20480}, :"/run"=>{:available=>"3.81 GiB", :available_bytes=>4091801600, :capacity=>"0.22%", :device=>"tmpfs", :filesystem=>"tmpfs", :options=>["rw", "seclabel", "nosuid", "nodev", "mode=755"], :size=>"3.82 GiB", :size_bytes=>4100960256, :used=>"8.73 MiB", :used_bytes=>9158656}, :"/run/user/1000"=>{:available=>"782.20 MiB", :available_bytes=>820195328, :capacity=>"0%", :device=>"tmpfs", :filesystem=>"tmpfs", :options=>["rw", "seclabel", "nosuid", "nodev", "relatime", "size=800972k", "mode=700", "uid=1000", "gid=1000"], :size=>"782.20 MiB", :size_bytes=>820195328, :used=>"0 bytes", :used_bytes=>0}, :"/sys/fs/cgroup"=>{:available=>"3.82 GiB", :available_bytes=>4100960256, :capacity=>"0%", :device=>"tmpfs", :filesystem=>"tmpfs", :options=>["ro", "seclabel", "nosuid", "nodev", "noexec", "mode=755"], :size=>"3.82 GiB", :size_bytes=>4100960256, :used=>"0 bytes", :used_bytes=>0}, :"/var/lib/nfs/rpc_pipefs"=>{:available=>"0 bytes", :available_bytes=>0, :capacity=>"100%", :device=>"sunrpc", :filesystem=>"rpc_pipefs", :options=>["rw", "relatime"], :size=>"0 bytes", :size_bytes=>0, :used=>"0 bytes", :used_bytes=>0}}
```

ekohl: It may need a `to_sym`. The idea is that you go "up" until it actually is found. By doing so, you're guaranteed to find the correct mountpoint that `mkdir -p` would also use. A minimal reproducer with debugging:
```ruby
require 'json'

# Taken from lib/kafo/base_context.rb
def symbolize(data)
  case data
  when Hash
    Hash[data.map { |key, value| [key.to_sym, symbolize(value)] }]
  when Array
    data.map { |v| symbolize(v) }
  else
    data
  end
end

FACTS = symbolize(JSON.parse(`facter --json`))

# Actual code

def available_bytes(directory)
  mountpoints = FACTS[:mountpoints]
  puts "Attempting #{directory}"
  until (mountpoint = mountpoints[directory.to_sym])
    raise Exception, "Could not find #{directory}" if directory == '/'
    directory = File.dirname(directory)
    puts "Attempting #{directory}"
  end
  puts "Found #{directory}"
  mountpoint[:available_bytes]
end

puts "Mountpoints: #{FACTS[:mountpoints].keys.inspect}"
puts available_bytes('/tmp/does/not/exist')
```

Output:
```
$ ruby disk_space.rb 
Mountpoints: [:/, :"/boot", :"/boot/efi", :"/dev", :"/dev/hugepages", :"/dev/mqueue", :"/dev/pts", :"/dev/shm", :"/run", :"/run/user/1000", :"/run/user/1000/gvfs", :"/tmp", :"/var/lib/nfs/rpc_pipefs"]
Attempting /tmp/does/not/exist
Attempting /tmp/does/not
Attempting /tmp/does
Attempting /tmp
Found /tmp
8328683520
```

Edit: the raise is redundant if everything works as it should, but prevents an infinite loop. I had this when I forgot the `.to_sym`.

katello/hooks/pre/30-upgrade.rb
```diff
@@ -37,6 +37,48 @@ def migrate_foreman
...
+  stop_services
+  Kafo::Helpers.execute('yum -y install rh-postgresql10-postgresql-server')
+  Kafo::Helpers.execute('scl enable rh-postgresql10 -- postgresql-setup --upgrade')
+  Kafo::Helpers.execute('yum -y remove postgresql postgresql-server')
+  Kafo::Helpers.execute('rm -f /etc/systemd/system/postgresql.service')
+  Kafo::Helpers.execute('yum -y install rh-postgresql10-syspaths')
+end
+
+def available_space(directory = nil)
+  directory = directory.nil? ? :/ : directory.to_sym
```
ehelms: I tackled the default and `to_sym` this way -- seemed to work for me testing locally.

katello/hooks/pre/30-upgrade.rb
```diff
@@ -37,6 +37,48 @@ def migrate_foreman
...
+  mountpoints = facts[:mountpoints]
+  until (mountpoint = mountpoints[directory])
```
ekohl: The `.to_sym` needs to be here because `directory` is modified in the loop. Otherwise it won't find mountpoints like `/var`.

ehelms: Right, good catch

katello/hooks/pre/30-upgrade.rb
```diff
@@ -37,6 +37,48 @@ def migrate_foreman
...
 
+def postgresql_10_upgrade
+  return unless facts[:os][:release][:major] == '7'
+
+  #Ensure there is at least 1x /var/lib/pgsql free disk space available on the main filesystem
+  current_postgres_dir = '/var/lib/pgsql'
+  new_postgres_dir = '/var/opt/rh/rh-postgresql10/lib/pgsql'
+
+  begin
+    until File.directory?(new_postgres_dir)
```
ekohl: The until loop is also in `available_space` so doesn't need to be duplicated here.

katello/hooks/pre/30-upgrade.rb
```diff
@@ -37,6 +37,48 @@ def migrate_foreman
...
+      new_postgres_dir = File.dirname(new_postgres_dir)
+    end
+
+    postgres_size = `du -k -s #{current_postgres_dir}`.split[0].to_i
```
ekohl: available space is in bytes and this in KB.

katello/hooks/pre/30-upgrade.rb
```diff
@@ -37,6 +37,48 @@ def migrate_foreman
   Kafo::Helpers.execute('foreman-rake db:migrate')
 end
 
+def postgresql_10_upgrade
+  return unless facts[:os][:release][:major] == '7'
+
+  #Ensure there is at least 1x /var/lib/pgsql free disk space available on the main filesystem
```
ekohl: I think we should move the disk space check to `katello/hooks/pre_validations` so we don't even start an upgrade unless there's sufficient disk space. Failing an upgrade leaves a feeling of an inconsistent state if it's not really.

ekohl (commented):
I feel like kafo should provide a proper mechanism to provide generic helpers. Things like `has_local_postgresql?` and `available_space` are useful in various places. Going to think about that.

katello/hooks/pre/30-upgrade.rb
ehelms: Editing in two places for testing, I caught htis on my live system just not in git!

ekohl: I often have the same issue :( Technically it's nowadays possible to run straight from git but I never really do this.

ehelms:
I moved to pre_validation which caused a bit of a re-factor of how I handled a few common methods. I'm trying to re-factor as little as possible with this change but I have seen some places improvements could be made.

katello/hooks/pre/30-upgrade.rb
```diff
@@ -37,6 +37,17 @@ def migrate_foreman
   Kafo::Helpers.execute('foreman-rake db:migrate')
 end
 
+def postgresql_10_upgrade
+  return unless facts[:os][:release][:major] == '7'
```
ekohl: I think this should be a guard around `upgrade_step :postgresql_10_upgrade`. The difference is that otherwise on EL8 you'll have a file that says it did the postgresql 10 upgrade which it didn't really.

katello/hooks/boot/01-helpers.rb
```diff
@@ -16,6 +16,11 @@ def log_and_say(level, message, do_say = true, do_log = true)
       Kafo::KafoConfigure.logger.send(level, message) if do_log
     end
 
+    def fail_and_exit(message)
```
ekohl, line 19: If you move this to `HookContextExtension` you don't need to prefix it in the actual hooks.

katello/hooks/boot/01-helpers.rb
```diff
@@ -52,6 +57,10 @@ module HookContextExtension
   def param_value(mod, name)
```
ekohl, line 57: Can you move the `FIXME` from line 55 to just above this function? Otherwise someone might remove `local_postgresql?` as well. The `unless Kafo::HookContext.method_defined?(:param_value)` at the end of the file is also a risk.

Perhaps it's cleaner to add a separate module for things that shouldn't be upstreamed to Kafo itself.

ekohl (commented):
Yes, the mongo hook can benefit a lot from these extensions.

katello/hooks/boot/01-helpers.rb
ehelms: The function it calls, log_and_say is already in the helper so I wanted to co-locate them for the short term. Based on https://github.com/theforeman/foreman-installer/pull/407#discussion_r349763345 and my own observations, dropping the Helper for functions that are added via an extension as a separate re-factor would be a good clean up. And to provide some parity with Foreman hooks.

I would love to tackle that after this. Trying not to overload this particular change.

ekohl: Fair enough. You can consider adding them in both places but this is already getting large.

ehelms:
This is ready for another look.

katello/hooks/boot/01-helpers.rb
```diff
@@ -47,11 +52,15 @@ def execute_command(command, do_say, do_log)
...
     param(mod, name).value if param(mod, name)
   end
+
+  def local_postgresql?
+    param_value('foreman', 'db_manage') || param_value('katello', 'candlepin_manage_db')
+  end
 end
 
 unless Kafo::HookContext.method_defined?(:param_value)
```
ekohl: Perhaps this `unless` part should be around `def param_value` so we unconditionally include the `HookContextExtension`. Or just straight up remove the conditional part.

katello/hooks/boot/01-helpers.rb
```diff
@@ -47,11 +52,15 @@ def execute_command(command, do_say, do_log)
...
-# FIXME: remove when #23332 is released
 module HookContextExtension
+  # FIXME: remove when #23332 is released
   def param_value(mod, name)
     param(mod, name).value if param(mod, name)
   end
+
+  def local_postgresql?
+    param_value('foreman', 'db_manage') || param_value('katello', 'candlepin_manage_db')
+  end
```
ekohl, line 63: How about adding the `fail_and_exit` this way:

```suggestion
  end
  
  def fail_and_exit(message)
    Kafo::Helpers.fail_and_exit(message)
  end
```

I'm also tempted to extract the helpers cleanup to a separate commit.

ekohl (commented):
My biggest concern is the `unless` part, other than that I think this is pretty much ready to merge.

Off topic: I also want to take the `HookContextExtension` further. It was introduced by Martin but should probably be a more official way to customize the hook context. Nuking Kafo::Helpers and doing it right is a good way to merge foreman and katello hooks.

katello/hooks/boot/01-helpers.rb
ehelms: Agreed, I dropped this.

katello/hooks/boot/01-helpers.rb
ehelms: I think the behavior here preserves the way it was working the best and co-locates the log methods. In general, this will need clean up but I really want to do it *after* this postgres work is merged. I can file a Redmine issue to track it if that helps.

ekohl: Probably a personal preference, but I'd have this PR as 2 commits: one that builds the infrastructure with `fail_and_exit` and the `HookContextExtension` and then one that only does the actual migration. Not a real difference in the end. Just a bit of personal preference in how I develop :)

ekohl (approved):
I didn't have a chance to test this yet, but it looks good.

PR #409, opened 2019-11-22
Fixes #28331 - Base memory checks on tuning profile

ekohl:
Each tuning profile has different system requirements so a generic 8 GB memory check is no longer sufficient. This patch always validates the tuning profile to be valid and uses it to check memory and CPU cores. It also uses facter rather than trying to determine it manually which makes the code easier to read.

By always validating the tuning, it becomes important that users are migrated when a tuning profile is removed.

Note that I don't have the specific requirements at hand and haven't tested this yet.


katello/hooks/pre_commit/13-hiera.rb
```diff
@@ -1,16 +1,43 @@
...
+  'default' => [1, 8],
+  'medium' => [1, 10],
+  'large' => [1, 20],
+  'extra-large' => [1, 128],
+  'extra-extra-large' => [1, 256],
+}
 TUNING_FACT = 'tuning'
 
 current = get_custom_fact(TUNING_FACT)
 new = app_value(:tuning)
```
evgeni: `new` is a keyword, can we use something else?

katello/hooks/pre_commit/13-hiera.rb
```diff
@@ -1,16 +1,43 @@
-TUNING_SIZES = ['default'] + ['medium', 'large', 'extra-large', 'extra-extra-large']
+# A mapping of tuning profile to the required CPU cores and memory in GB
+# TODO: check actual required values
+TUNING_SIZES = {
+  'default' => [1, 8],
```
evgeni: can this be a Hash instead of an Array? two values might still be possible to remember, but if we ever want to add more this will be messy.

also, how about modern style Hash syntax with `:` instead of `=>`?

katello/hooks/pre_commit/13-hiera.rb
ekohl: Already existed, but good point.

katello/hooks/pre_commit/13-hiera.rb
ekohl: > can this be a Hash instead of an Array? two values might still be possible to remember, but if we ever want to add more this will be messy.

I considered that, but went with the simple route. If we add more (per scenario, disk space) then we can refactor it later

> also, how about modern style Hash syntax with `:` instead of `=>`?

We still need to support Ruby 2.0 and the key is a string anyway.

ekohl:
Updated:
* Added a commit to rename `new`
* Rephrased the error message slightly
* Added a missing `[:system]` for the memory fact
* Only check for 90% of the memory instead of 95%. On my forklift VM with 6 GB it wouldn't pass with a 95% check but would with 90% (with a 6 GB check instead of default 8).

katello/hooks/pre_commit/13-hiera.rb
ehelms: I echo @evgeni , please make this a hash.

katello/hooks/pre_commit/13-hiera.rb
```diff
@@ -1,17 +1,44 @@
...
+    say "<%= color('Insufficient memory for tuning size', :bad) %>"
+    say "Tuning profile '#{new_tuning}' requires at least #{required_memory} GB of memory"
+    exit 102
   end
 
-  store_custom_fact(TUNING_FACT, new)
+  if facts[:processors][:count] < required_cores
+    say "<%= color('Insufficient CPU cores for tuning size', :bad) %>"
+    say "Tuning profile '#{new_tuning}' requires at least #{required_cores} CPU cores"
+    exit 103
```
ehelms: What are the various exit numbers meanings?

ekohl: Well, 103 now means you have insufficient CPU cores :) Perhaps we should add a table in README similar to https://github.com/theforeman/kafo#exit-code

katello/hooks/pre_commit/13-hiera.rb
ekohl: I'll do so when we have a table of the actual required values so I'm not changing the same piece of code twice.

ehelms: I don't entirely follow what the actual required values has to do with it? I am thinking:

```
TUNING_SIZES = {
  'default' => {'cpu_cores' => 1, 'memory' => 8},
  'medium' => {'cpu_cores => 1, 'memory' => 10},
```

This is very readable for me without having to go read the code to figure out how these values are being used.

katello/hooks/pre_commit/13-hiera.rb
ehelms: That would be a good start. Mapping them to constants could also be a good tactic:

```
INSUFFICIENT_CPU_CORES = 103

exit INSUFFICIENT_CPU_CORES
```

katello/hooks/pre_commit/13-hiera.rb
ekohl: The values I used for the profiles were placeholders. Luckily @sthirugn documented them in https://github.com/theforeman/theforeman.org/pull/1504 which makes my life easier :) Now that I have the correct list, I'll update it.

The only thing is that the default profile has a very different value from what we actually require.

ekohl:
I'm tempted to store the tuning profiles and their sizes in the scenario so we can be more flexible. It would allow to have a different minimum for a content proxy than a katello. It'd increase the scope too much now, but I think it'd be clear. We'd also gain flexibility to have a different minimum in upstream vs downstream.

ekohl:
Updated:
* Updated the TUNING_SIZES definition. The profiles still use strings since that's the input. The dashes also break symbols so this is cleaner. The specific profile does use symbols.
* Updated the error message to include both memory and cores

```console
# foreman-installer --tuning large
Insufficient memory for tuning size
Tuning profile 'large' requires at least 64 GB of memory and 16 CPU cores
# foreman-installer --tuning medium # manually changed the values just to trigger a CPU count error
Insufficient CPU cores for tuning size
Tuning profile 'default' requires at least 6 GB of memory and 3 CPU cores
# foreman-installer --tuning default
Preparing installation Done
...
```

katello/hooks/pre_commit/13-hiera.rb
ehelms: THoughts on this idea? Even if stored at the top of the file. I know that if I com eback to this file I will likely look at 101, 102, 103 and start googling Ruby exit codes and rabbit hole.

ekohl: I'll start with just defining them in the same file. We can do the general exit code cleanup. Then we should also look at the other hooks which probably override Kafo exit codes somewhere.

ekohl: Updated

katello/hooks/pre_commit/13-hiera.rb
```diff
@@ -1,17 +1,48 @@
...
+unless app_value(:disable_system_checks)
+  required_cores = required[:cpu_cores]
+  required_memory = required[:memory]
+
+  # Check if it's actually 90% of the required. If a crash kernel is enabled
+  # then the reported total memory is lower than in reality.
+  if facts[:memory][:system][:total_bytes] < (required_memory * 1024 * 1024 * 1024 * 0.9)
+    say "<%= color('Insufficient memory for tuning size', :bad) %>"
+    say "Tuning profile '#{new_tuning}' requires at least #{required_memory} GB of memory and #{required_cores} CPU cores"
+    exit EXIT_INSUFFICIENT_CPU_CORES
```
ehelms: This should be memory :)

katello/hooks/pre_commit/13-hiera.rb
```diff
@@ -1,17 +1,48 @@
...
   end
 
-  store_custom_fact(TUNING_FACT, new)
+  if facts[:processors][:count] < required_cores
+    say "<%= color('Insufficient CPU cores for tuning size', :bad) %>"
+    say "Tuning profile '#{new_tuning}' requires at least #{required_memory} GB of memory and #{required_cores} CPU cores"
+    exit EXIT_INSUFFICIENT_MEMORY
```
ehelms: This should be CPU cores :)

katello/hooks/pre_commit/13-hiera.rb
ekohl: It has been a long day.

ekohl:
Swapped the exit codes

katello/hooks/pre_commit/13-hiera.rb
```diff
@@ -1,17 +1,48 @@
-TUNING_SIZES = ['default'] + ['medium', 'large', 'extra-large', 'extra-extra-large']
+# A mapping of tuning profile to the required CPU cores and memory in GB
+TUNING_SIZES = {
+  'default' => {cpu_cores: 1, memory: 8},
```
evgeni, line 3: didn't you say you wanted be ruby 2.0 compatible? :)

evgeni (commented):
looks good to me besides the "but ruby2.0" comment :)

katello/hooks/pre_commit/13-hiera.rb
ekohl: Turns out that's Ruby 1.9 that introduced this syntax. I tested it on EL7 and it worked as expected so I'm good with it :)

evgeni: :D

ehelms (approved):

sthirugn:
@ekohl Great work, thank you for this PR.

PR #411, opened 2019-11-25
Fix REX Cockpit port and Azure support

ekohl:
theforeman-foreman 13.1.0 brings AzureRM and Azure CLI support. Also a bugfix in the Cockport port.

Depends on https://github.com/theforeman/puppet-foreman/pull/775.


ekohl:
cc @tbrisker

evgeni (approved):
while I am unsure why this PR is called "fix" while it contains features, these are all features we want in 1.24, so ACK

tbrisker (approved):
Thanks @ekohl !

ekohl:
> while I am unsure why this PR is called "fix" while it contains features, these are all features we want in 1.24, so ACK

Next time I'll write "Azure support and fix REX Cockpit port"

PR #416, opened 2019-11-27
Fixes #28370 - set the default tuning profile to default

evgeni:
when there is no tuning profile applied yet, `get_custom_fact` will return
`nil` and that's not a sensible default for the tuning profile


evgeni:
meh, I can't read. there is a migration.

PR #417, opened 2019-11-29
Fixes #23727 - Rely on puppet-pulp to migrate the database

ekohl:
https://github.com/theforeman/puppet-pulp/blob/9ee5f9e9539754a2a270695a70911dd6417a4de8/manifests/database.pp#L11-L18 should already take care of this. By removing the marker file we guarantee it runs. It also ensures the correct services are running prior to migration.

### DIFF
--- a/katello/hooks/pre/30-upgrade.rb
+++ b/katello/hooks/pre/30-upgrade.rb
@@ -2,6 +2,9 @@ require 'fileutils'
 
 STEP_DIRECTORY = '/etc/foreman-installer/applied_hooks/pre/'
 
+# puppet-pulp uses this file to know whether it needs to run pulp-manage-db
+PULP2_MIGRATION_MARKER_FILE = '/var/lib/pulp/init.flag'
+
 def stop_services
   Kafo::Helpers.execute('foreman-maintain service stop')
 end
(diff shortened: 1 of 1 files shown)


katello/hooks/pre/30-upgrade.rb
```diff
@@ -87,8 +84,8 @@ def step_path(step)
     upgrade_step :start_postgresql, :run_always => true
   end
 
-  if katello || foreman_proxy_content
-    upgrade_step :migrate_pulp, :run_always => true
+  if File.exist?(PULP2_MIGRATION_MARKER_FILE)
+    File.unlink(PULP2_MIGRATION_MARKER_FILE)
```
ehelms, line 88: Why unlink?

ekohl: https://puppet.com/docs/puppet/latest/type.html#exec-attribute-creates (currently direct links are broken) states:
> creates     => # A file to look for before running the command...

The exec runs `pulp-manage-db && touch /var/lib/pulp/init.flag`. That makes it idempotent. By removing the file, we trigger the exec and thus a database migration.

ehelms (approved):

PR #418, opened 2019-11-29
Rely on puppet-candlepin for cpdb execution

ekohl:
Draft because it includes https://github.com/theforeman/foreman-installer/pull/417 and relies on https://github.com/theforeman/puppet-candlepin/pull/139.

### DIFF
--- a/katello/hooks/pre/30-upgrade.rb
+++ b/katello/hooks/pre/30-upgrade.rb
@@ -2,6 +2,8 @@ require 'fileutils'
 
 STEP_DIRECTORY = '/etc/foreman-installer/applied_hooks/pre/'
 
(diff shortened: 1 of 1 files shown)


ekohl:
@ehelms with the Foreman DB migration work added we can remove the `--upgrade` flag on this and make it a regular hook. We should check if `postgresql-server` is installed prior to running it.

ehelms:
You can take this out of draft and I think it's good to go. I would not drop the upgrade hook here due to postgres upgrades.

ekohl:
Rebased. Still depends on https://github.com/theforeman/puppet-candlepin/pull/139 to be merged first so for now still a draft.

ehelms (approved):
I'm monitoring current nightly with some fixes that have been merged. I'll merge this once I think its in a good place.

PR #422, opened 2019-12-03
Fixes #28407 - Use rh-redis5 on EL7

ekohl:
Submitting as a draft since I haven't tested this yet.


ehelms (approved):
Seems reasonable, Redis is now in Katello nightly usage as well.

PR #423, opened 2019-12-03
Install Postgresql server RPMs to prevent selinux issue

ehelms:


katello/hooks/pre/32-install_postgresql.rb
```diff
@@ -0,0 +1,7 @@
+# Working around https://tickets.puppetlabs.com/browse/PUP-2169
+if local_postgresql? && facts[:os][:release][:major] == '7' && !app_value(:upgrade)
+  `rpm -qa | grep rh-postgresql10-postgresql-server-syspaths`
+  unless $?.success?
+    Kafo::Helpers.execute('yum -y install rh-postgresql10-postgresql-server-syspaths')
```
evgeni: I know this is not the `upgrade` path, so the system *should* be clean, but would it be sufficient to install `rh-postgresql10-postgresql-server` and let the installer install `rh-postgresql10-postgresql-server-syspaths` later?

katello/hooks/pre/32-install_postgresql.rb
```diff
@@ -0,0 +1,7 @@
+# Working around https://tickets.puppetlabs.com/browse/PUP-2169
+if local_postgresql? && facts[:os][:release][:major] == '7' && !app_value(:upgrade)
+  `rpm -qa | grep rh-postgresql10-postgresql-server-syspaths`
```
evgeni:
```suggestion
  `rpm -q rh-postgresql10-postgresql-server-syspaths`
```

no need to `grep`, `rpm` will exit non-zero if the package is not installed.

katello/hooks/pre/32-install_postgresql.rb
```diff
@@ -0,0 +1,7 @@
+# Working around https://tickets.puppetlabs.com/browse/PUP-2169
+if local_postgresql? && facts[:os][:release][:major] == '7' && !app_value(:upgrade)
+  `rpm -q rh-postgresql10-postgresql-server`
```
ekohl: I'd prefer the package name in a variable

ekohl (approved):
I'd prefer this hook is number 29 and drops the `!app_value(:upgrade)` part so it always installs it. Then you can drop it from `katello/hooks/pre/30-upgrade.rb`. Not strictly a blocker

ekohl (approved):

katello/hooks/pre/29-install_postgresql.rb
```diff
@@ -0,0 +1,10 @@
+# Working around https://tickets.puppetlabs.com/browse/PUP-2169
+if local_postgresql? && facts[:os][:release][:major] == '7' && !app_value(:upgrade)
+  package = 'rh-postgresql10-postgresql-server'
```
ehelms, line 3: Added

katello/hooks/pre/29-install_postgresql.rb
```diff
@@ -0,0 +1,10 @@
+# Working around https://tickets.puppetlabs.com/browse/PUP-2169
+if local_postgresql? && facts[:os][:release][:major] == '7' && !app_value(:upgrade)
```
ehelms, line 2: For now, I don't want this messing with the upgrade workflow given thats a very orchestrated change.

PR #424, opened 2019-12-05
Refs #28353 - Move fail_and_exit to HookContextExtension

ekohl:
By placing it in HookContextExtension, it can be used in every hook without redefining it.


evgeni (approved):
I like red diffs

ekohl:
> I like red diffs

Unlike test results, here red is good :)

PR #427, opened 2019-12-10
Set PostgreSQL 10 defaults on EL 7 for all scenarios

ehelms:
This effectively moves Foreman and Katello Devel scenarios to PostgreSQL 10 as well without having to write a bunch of work arounds.


ekohl (changes requested):
we'll need upgrade hooks in vanilla Foreman as well to upgrade those users as well. I'd prefer to first make sure Katello can upgrade safely before we copy those.

ehelms:
This PR (https://github.com/theforeman/foreman-installer/pull/436) will add the ground work for adding this upgrade flag.

ekohl:
https://github.com/theforeman/puppet-foreman/pull/781 will also simplify the logic needed.

ehelms:
Blocked on -- https://github.com/theforeman/foreman-installer/pull/440

ehelms:
Will now require migrating the upgrade hook to main Foreman -- https://github.com/theforeman/foreman-installer/pull/443

ekohl:
Looks like https://github.com/theforeman/foreman-installer/pull/444 now includes this so should we close this?

PR #428, opened 2019-12-11
Always set locale when upgrading postgres

ehelms:


katello/hooks/pre/30-upgrade.rb
```diff
@@ -35,8 +35,9 @@ def migrate_foreman
 
 def postgresql_10_upgrade
   stop_services
+  locale = `locale | grep LANG`.split('LANG=')[1].strip
```
ekohl: We enforce `en_US.utf8` in [puppet-foreman](https://github.com/theforeman/puppet-foreman/blob/36aa5389b7da41d4770fa61a1c10a78239695735/manifests/database/postgresql.pp#L25) and [puppet-candlepin](https://github.com/theforeman/puppet-candlepin/blob/2f08eeb92c96a0c2bca3e381780cc58530467459/manifests/database/postgresql.pp#L42). Perhaps safer to do the same here?

ehelms:
Errors seen:

```
lc_collate values for database "postgres" do not match:  old "en_GB.UTF-8", new "en_US.UTF-8"
```

evgeni:
as discussed on irc (but noting here for everyone else), I think it would be better to pull this data from the running database instead of the environment the current setup is running in.

the collate (and ctype) data can be pulled by something like `psql -t -c "select datcollate,datctype from pg_catalog.pg_database where datname='postgres';" -d postgres`

ehelms:
Any ideas on how to run that psql command from Ruby? The normal backticks or system command doesn't like having to switch users:

```
  locale_t = `su - postgres -c "psql -t -c \"select datcollate,datctype from pg_catalog.pg_database where datname='postgres';\" -d postgres"`.strip
```

ekohl:
`su` involves using pam while `runuser` avoids the pam session. That probably helps.

ehelms:
Oh actually, I wonder if its because of the `-c` in psql so its a shell within a shell and Ruby can't handle that.

ehelms:
For example:

```
[root@centos7-katello-3-13 vagrant]# irb
irb(main):001:0> `runuser postgres -c "psql -d postgres -t -c \"select datcollate,datctype from pg_catalog.pg_database where datname='postgres';\""`
could not change directory to "/home/vagrant"
ERROR:  syntax error at end of input
LINE 1: select
              ^
sh: : command not found
=> ""
```

ehelms:
I;m stuck on how to get this data without a crazy solution. Any appetite for solving this to unblock nightly with this more naive approach and filing an issue to find a more robust solution?

ekohl:
Is `%x{runuser postgres -c "psql -d postgres -t -c \"select datcollate,datctype from pg_catalog.pg_database where datname='postgres';\""}` better?

Edit: otherwise open3 is a more robust way since you can pass in arguments as an array rather than as a shell string.

ehelms:
The problem I realized isn't with switching users, the problem as I can ascertain is that the Ruby system command opens a shell and then psql opens a shell and so the shell within a shell does not work. This does work if I put the whole command in a bash scrpit and then execute the bash script instead.

I could:

 1) Write the command to a temporary file and then execute that
 2) Put a script into foreman-installer bin/ to return the locale values we need and call that

evgeni:
```ruby
(name, owner, enconding, collate, ctype, privileges) = %x{runuser postgres -c 'psql -lt | grep -E "^\s+postgres"'}.chomp.split('|').map(&:strip)
```

evgeni:
(`psql -l` lists all databases and their properties, which is similar to what you'd get by poking the tables I mentioned before)

katello/hooks/pre/30-upgrade.rb
evgeni: fwiw, the failing DB is `postgres`, not `foreman` or `candlepin`, and we don't create that.

katello/hooks/pre/30-upgrade.rb
```diff
@@ -34,9 +34,11 @@ def migrate_foreman
 end
 
 def postgresql_10_upgrade
+  start_postgresql
```
ekohl, line 37: Due to line 82 you should be able to count on PG already running. This avoids a duplicated log for the end user (since `Kafo::Helpers.execute` is verbose).

ehelms: I thought about that, but if something errors you can potentially end up in a scenario where this continually fails (as the line after this won't work). I wanted to err on the side of better safe than sorry.

ekohl: Since `systemctl` is quiet, should we call it quietly by passing in `do_say` as false?

Also note that execute returns a boolean whether it succeeded but doesn't actually halt execution. @evgeni noticed that elsewhere as well. We should handle that so it actually stops the upgrade attempt.

katello/hooks/pre/30-upgrade.rb
```diff
@@ -34,9 +34,11 @@ def migrate_foreman
...
+  (name, owner, enconding, collate, ctype, privileges) = %x{runuser postgres -c 'psql -lt | grep -E "^\s+postgres"'}.chomp.split('|').map(&:strip)
   stop_services
   Kafo::Helpers.execute('yum -y install rh-postgresql10-postgresql-server')
-  Kafo::Helpers.execute('scl enable rh-postgresql10 -- postgresql-setup --upgrade')
+  Kafo::Helpers.execute("scl enable rh-postgresql10 \"PGSETUP_INITDB_OPTIONS='--lc-collate=\"#{collate}\" --lc-ctype=\"#{ctype}\"' postgresql-setup --upgrade\"")
```
evgeni: couple of questions/comments, all of which are probably from the "super picky" kind:
* before this used `--` between `scl enable …` and the actual command, now it doesn't, what does that mean?
* should we also pass `--locale=%{locale}` for consistency?
* I think the quoting is a bit off. the string will evaluate to `scl enable rh-postgresql10 "PGSETUP_INITDB_OPTIONS='--lc-collate="#{collate}" --lc-ctype="#{ctype}"' postgresql-setup --upgrade"` so the section that starts at `PGSETUP_INITDB_OPTIONS` actually ends after `--lc-collate`
I wonder if the following is easier to parse
```suggestion
  Kafo::Helpers.execute(%Q{scl enable rh-postgresql10 "PGSETUP_INITDB_OPTIONS='--lc-collate=#{collate} --lc-ctype=#{ctype}' postgresql-setup --upgrade"})
```
(I don't think we need to quote collate/ctype)

ehelms: THanks for the suggestion, I updated it to a version that drops even more quotes and works.

ehelms:
```
yum -y install rh-postgresql10-postgresql-server finished successfully!
 * upgrading from 'postgresql.service' to 'rh-postgresql10-postgresql.service'
 * Upgrading database.
 * Upgraded OK.
WARNING: The configuration files were replaced by default configuration.
WARNING: The previous configuration and data are stored in folder
WARNING: /var/lib/pgsql/data.
 * See /var/lib/pgsql/upgrade_rh-postgresql10-postgresql.log for details.
scl enable rh-postgresql10 "PGSETUP_INITDB_OPTIONS='--lc-collate=en_US.UTF-8 --lc-ctype=en_US.UTF-8' postgresql-setup --upgrade" finished successfully!
```

katello/hooks/pre/30-upgrade.rb
ehelms: I'm not sure I get the problem with this logging what it's doing. I'd rather it have that as someone debugging an issue than not.

And yes, planning to address after this fix goes in: https://projects.theforeman.org/issues/28496

katello/hooks/pre/30-upgrade.rb
evgeni: answering myself: Note: if you use <command> consisting of multiple arguments, you either need to use quotes or the -- command separator. Everything that follows the separator will be considered a command or its argument.

evgeni (approved):
i am happy to adress the other issues via https://projects.theforeman.org/issues/28496

this gets the job done

katello/hooks/pre/30-upgrade.rb
ekohl: My concern is that the user will see `systemctl start postgresql` logged twice during an upgrade which may look odd. Not a critical issue and probably worth a larger refactoring to see how we want to deal with output. I hope we'll get to that when we merge clean up the hooks and merge them with the foreman scenario.

katello/hooks/pre/30-upgrade.rb
```diff
@@ -34,9 +34,11 @@ def migrate_foreman
 end
 
 def postgresql_10_upgrade
+  start_postgresql
+  (name, owner, enconding, collate, ctype, privileges) = %x{runuser postgres -c 'psql -lt | grep -E "^\s+postgres"'}.chomp.split('|').map(&:strip)
```
ekohl, line 38: Nit: usually a fan of long options in scripts (`--list --tuples-only` instead of `-lt`)

ekohl (approved):

PR #429, opened 2019-12-12
Fixes #28496: exit 1 on execute command

ehelms:
This makes a change for all commands calling execute that if they fail this will stop execution of the installer. I think this is a safe change given it would stop things like upgrade in its tracks if something fails.

### DIFF
--- a/katello/hooks/post/30-upgrade.rb
+++ b/katello/hooks/post/30-upgrade.rb
@@ -1,13 +1,11 @@
 require 'fileutils'
 
 def db_seed
-  status = execute('foreman-rake db:seed')
-  fail_and_exit "Upgrade failed while seeding the DB" unless status
+  execute('foreman-rake db:seed')
 end
 
 def upgrade_tasks
-  status = execute('foreman-rake upgrade:run')
-  fail_and_exit "Application Upgrade Failed" unless status
+  execute('foreman-rake upgrade:run')
 end
 
 if app_value(:upgrade)
--- a/katello/hooks/pre_validations/31-upgrade-puppet.rb
+++ b/katello/hooks/pre_validations/31-upgrade-puppet.rb
@@ -1,6 +1,6 @@
 unless app_value(:noop)
-  if File.exist?('/opt/puppetlabs/puppet/bin/ruby') && execute("rpm -q puppet-agent-oauth", false, false)
-    unless execute("/opt/puppetlabs/puppet/bin/ruby -e \"require 'oauth'\"", false, false)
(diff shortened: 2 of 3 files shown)


katello/hooks/post/30-upgrade.rb
```diff
@@ -1,13 +1,11 @@
 require 'fileutils'
 
 def db_seed
-  status = Kafo::Helpers.execute('foreman-rake db:seed')
-  fail_and_exit "Upgrade failed while seeding the DB" unless status
+  Kafo::Helpers.execute('foreman-rake db:seed')
```
evgeni: This drops the custom error message, which I think is OK, but would like to hear other peoples opinions.

katello/hooks/pre_validations/31-upgrade-puppet.rb
```diff
@@ -1,7 +1,7 @@
 unless app_value(:noop)
-  if File.exist?('/opt/puppetlabs/puppet/bin/ruby') && execute("rpm -q puppet-agent-oauth", false, false)
-    unless execute("/opt/puppetlabs/puppet/bin/ruby -e \"require 'oauth'\"", false, false)
-      execute("yum -y reinstall puppet-agent-oauth")
+  if File.exist?('/opt/puppetlabs/puppet/bin/ruby') && system("rpm -q puppet-agent-oauth")
+    unless system("/opt/puppetlabs/puppet/bin/ruby -e \"require 'oauth'\"")
+      Kafo::Helpers.execute("yum -y reinstall puppet-agent-oauth")
```
ekohl: Any reason you're changing this line? I thought we'd want to keep using the execute method.

ehelms: This line should just be the execute method now which I updated.

katello/hooks/pre_validations/31-upgrade-puppet.rb
```diff
@@ -1,6 +1,6 @@
 unless app_value(:noop)
-  if File.exist?('/opt/puppetlabs/puppet/bin/ruby') && execute("rpm -q puppet-agent-oauth", false, false)
-    unless execute("/opt/puppetlabs/puppet/bin/ruby -e \"require 'oauth'\"", false, false)
+  if File.exist?('/opt/puppetlabs/puppet/bin/ruby') && system("rpm -q puppet-agent-oauth")
+    unless system("/opt/puppetlabs/puppet/bin/ruby -e \"require 'oauth'\"")
```
ehelms, line 3: These two are moving to system since we want to know if it's successful but without logging or stopping the execution.

ekohl (approved):

PR #430, opened 2019-12-12
Refs #28353: Drop Kafo::Helpers and move to extensions

ehelms:


katello/hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -0,0 +1,58 @@
...
+    log_and_say :error, message
+    exit 1
+  end
+
+  def local_postgresql?
+    param_value('foreman', 'db_manage') || param_value('katello', 'candlepin_manage_db')
+  end
+
+  def log_and_say(level, message, do_say = true, do_log = true)
+    style = case level
```
ekohl, line 17: I was thinking about this and if we can actually add them to the color scheme that Kafo already uses.

katello/hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -0,0 +1,58 @@
...
+              'bad'
+            when :debug
+              'yellow'
+            else
+              level
+            end
+
+    # \ and ' characters could cause trouble in ERB, make sure to escape them
+    escaped_message = message.gsub('\\', '\\\\\\').gsub("'", %q{\\\'})
+    say "<%= color('#{escaped_message}', :#{style}) %>" if do_say
```
ekohl, line 28: I was looking at this but haven't tried yet if it works:
```suggestion
    say HighLine.color(message, style.to_sym) if do_say
```

ekohl (approved):
This is a good first pass. I think that almost all of the methods in HookContextExtension (except `local_postgresql?`) are good candidates to include in Kafo itself.

katello/hooks/boot/01-kafo-hook-extensions.rb
ehelms: Sounds like a good one to capture into a Redmine issue.

ehelms:
For moving into Kafo -- https://projects.theforeman.org/issues/28498

PR #431, opened 2019-12-12
Refs #28353: Rename boot hooks to option naming convention

ehelms:


ekohl:
I think the Ruby convention is underscores. Since they're ruby files, should we instead adopt that?

ehelms:
I suppose I can go either way. There are guidelines, but less clear for this:

https://guides.rubygems.org/name-your-gem/

So reading that I suppose sticking with:

<number>-<category>-name_of_the_thing.rb

e.g.

10-option-reset.rb
11-option-clear_pulp_content.rb

ehelms:
@ekohl thoughts on the naming?

PR #434, opened 2019-12-17
Fixes #28525: Install redis from SCL to prevent selinux issue

ehelms:


hooks/pre/30-install_redis_scl.rb
```diff
@@ -0,0 +1,10 @@
+# Working around https://tickets.puppetlabs.com/browse/PUP-2169
+if facts[:os][:release][:major] == '7'
```
ekohl: Please also check if this is the Red Hat family
```suggestion
if facts[:os][:family] == 'RedHat' && facts[:os][:release][:major] == '7'
```

hooks/pre/30-install_redis_scl.rb
```diff
@@ -0,0 +1,10 @@
...
+  package = 'rh-redis5-redis'
+
+  `rpm -q #{package}`
+
+  unless $?.success?
+    execute("yum -y install #{package}")
```
ekohl: I'm wondering if we should just leave this up to Puppet:
```suggestion
    execute("puppet resource package #{package} ensure=installed")
```

Then you can drop the whole `rpm -q` as well.

ekohl (changes requested):
You will need to check if a local Redis is actually needed. Can we introduce a hook context `local_redis?` for this?

ehelms:
I don't think I realized we were supporting a non-local Redis.

ekohl:
We probably don't (though the current Foreman PR has a parameter for it), but in the vanilla foreman case there is no separate proxy scenario. So you disable the foreman module which means no local redis is needed anymore. Technically right now the foreman-proxy-content scenario also doesn't require it.

ehelms:
I can add just for long term potential support. As far as I know, we have two scenarios for Redis:

 1) Pulp 3 on a server
 2) Dynflow on a server

And in both cases we are using the same Redis instance and have no need to externalize it.

ekohl:
You can also name it `redis?` or `has_redis?` but we need the check, at least for the Foreman scenario. This is to cover the `foreman-installer --scenario --no-enable-foreman` which is the way to install just a foreman-proxy. We don't have Redis in that case.

ehelms:
This PR going in first (https://github.com/theforeman/foreman-installer/pull/436) will make this solution cleaner. So I am gonna block on that before updating this one.

hooks/pre/30-install_redis_scl.rb
ehelms: Based on this, I wanted to introduce a single command to do this across the hooks:

https://github.com/theforeman/foreman-installer/pull/440

ehelms:
The main branch plus https://github.com/theforeman/foreman-installer/pull/444 have some nice helper functions that I will wait on before updating this

ehelms:
This now uses the `ensure_package` function, provides a single hook and includes a `local_redis?` and `el7?` check.

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -30,6 +30,14 @@ def local_postgresql?
     param_value('foreman', 'db_manage') || param_value('katello', 'candlepin_manage_db')
   end
 
+  def local_redis?
+    !param_value('foreman', 'jobs_sidekiq_redis_url')
```
ekohl: If the foreman module is disabled then this results in true.

ekohl (changes requested):

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -26,10 +26,22 @@ def fail_and_exit(message)
...
+  def foreman_server?
+     module_enabled?('foreman')
+  end
+
   def local_postgresql?
     param_value('foreman', 'db_manage') || param_value('katello', 'candlepin_manage_db')
   end
 
+  def local_redis?
+    !param_value('foreman', 'jobs_sidekiq_redis_url') && foreman_server?
```
ehelms: Added a check that its a foreman_server

ekohl (commented):
In this case I think `module_enabled?('foreman') && !param_value('foreman', 'jobs_sidekiq_redis_url')` was probably sufficient. You probably also want also do it when pulpcore is enabled.

ehelms:
The use of `foreman_server?` function is included because the same logic is being used in other PRs. So adding the helper here just prevents a re-factor. I find this cleaner than putting the same `module_enabled?('foreman')` all over our code like we do today.  Examples:

 * https://github.com/theforeman/foreman-installer/pull/444/files#diff-c5094b8de55721df96da41bf73960170R13
 * https://github.com/theforeman/foreman-installer/pull/444/files#diff-8bcd35edfa4d3d66697be4c5d5a5f5e6R26

Added pulpcore check

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -26,10 +26,22 @@ def fail_and_exit(message)
...
+  def foreman_server?
+    module_enabled?('foreman')
+  end
+
   def local_postgresql?
     param_value('foreman', 'db_manage') || param_value('katello', 'candlepin_manage_db')
   end
 
+  def local_redis?
+    (foreman_server? && !param_value('foreman', 'jobs_sidekiq_redis_url')) || module_enabled?('pulpcore')
```
ekohl: We don't enable pulpcore as a module in the installer.
```suggestion
    (foreman_server? && !param_value('foreman', 'jobs_sidekiq_redis_url')) || param_value('foreman_proxy::plugin::pulp', 'pulpcore_enabled')
```

ehelms: Ahh yea, I was on autopilot writing this. Thanks!

ekohl (approved):

PR #435, opened 2019-12-17
Add foreman proxy scenario

ehelms:
Includes https://github.com/theforeman/foreman-installer/pull/262 with updates

### DIFF
--- a/config/foreman-proxy-answers.yaml
+++ b/config/foreman-proxy-answers.yaml
@@ -11,9 +11,7 @@
 ---
 foreman_proxy:
   custom_repo: true
-puppet:
-  server: false
-foreman_proxy::plugin::abrt: false
+puppet: false
 foreman_proxy::plugin::ansible: false
 foreman_proxy::plugin::chef: false
 foreman_proxy::plugin::dhcp::infoblox: false
--- a/config/foreman-proxy-answers.yaml
+++ b/config/foreman-proxy-answers.yaml
@@ -1,0 +1,30 @@
+# Format:
+# <classname>: false - don't include this class
+# <classname>: true - include and use the defaults
+# <classname>:
+#   <param>: <value> - include and override the default(s)
+#
+# Every support plugin class is listed, so that it
+# shows up in the interactive menu
+#
+# See params.pp in each class for what options are available
+---
+foreman_proxy:
+  custom_repo: true
+puppet:
+  server: false
+foreman_proxy::plugin::abrt: false
+foreman_proxy::plugin::ansible: false
+foreman_proxy::plugin::chef: false
+foreman_proxy::plugin::dhcp::infoblox: false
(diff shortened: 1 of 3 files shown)


config/foreman-proxy-answers.yaml
```diff
@@ -0,0 +1,28 @@
...
+# <classname>:
+#   <param>: <value> - include and override the default(s)
+#
+# Every support plugin class is listed, so that it
+# shows up in the interactive menu
+#
+# See params.pp in each class for what options are available
+---
+foreman_proxy:
+  custom_repo: true
```
ekohl, line 13: This no longer exists and can be dropped.

ekohl (changes requested):
My issue with this is that we will need to update **all** manuals. They say nothing about passing in a `--scenario` argument. I'd like to hear @tbrisker's opinion about this as well. On the other hand, 2.0 is a good moment for that.

tbrisker:
I'm fine with adding a scenario, tbh I'm a bit surprised that while the installer is able to install just a proxy (and has been for a long while if i'm not mistaken), [the manual](https://theforeman.org/manuals/nightly/index.html#4.3.1SmartProxyInstallation) tells users to manually install it from packages.

ekohl:
That's from the advanced section. [Installation Scenarios](https://theforeman.org/manuals/nightly/index.html#3.2.3InstallationScenarios) describes how to use the installer for it. It basically comes down to passing in `--no-enable-foreman --no-enable-foreman-cli` and you have the same result as this. The benefit for us as maintainers is that we don't need to duplicate all the migrations. We also don't force uses to make an explicit choice whether something is a proxy or a foreman server: there's just components you can enable/disable.

ehelms:
I'm not sure I follow clearly what the argument against this is. As I see it we have two deployments in Foreman:

 1) Foreman Server w/ builtin proxy
 2) External proxy

We should make this obvious and easy for users to install one or the other.

ekohl:
> I'm not sure I follow clearly what the argument against this is

It is additional work for us and by merging this now we break a lot of documentation. First of all, this needs a Redmine issue so it's properly tracked in the release notes. Then we need to make sure our documentation is updated to reflect this before RC1. As long as we can make that happen, I'm good with it.

After that we should also consider how we handle it in forklift. For Katello we also generate `-proxy` boxes. Should we do the same for vanilla and further explode the number of boxes we have?

ehelms:
Ahh yea, I am good with making documentation updates. We technically are not breaking anything by including this. We are adding functionality that we can switch our documentation to as the preferred method for installation of a stand alone proxy but it does not invalidate the current method.

ekohl:
Well, before this change this worked:
```
yum -y install foreman-installer
foreman-installer
```

After it needs to be:
```
yum -y install foreman-installer
foreman-installer --scenario foreman
```

That means documentation changes.

PR #436, opened 2019-12-17
Fixes #28530: Combine post install hooks into a single one

ehelms:
This moves to a single post install hook to handle all scenarios.
This introduces the message helpers to Foreman hooks and starts
to provide the hook extensions Katello uses.

This starts the road to a single hooks directory that handles all scenarios properly. Further it introduces a few concepts in Katello's hooks to Foreman's to pave the way for more changes. I did not want to re-write it all in one go but rather baby step changes in. Thus, I figured the message helpers and post install hook being combined was a good first.


hooks/boot/02-message-helpers.rb
```diff
@@ -40,5 +40,15 @@ def new_install_message(kafo)
...
       say "      Initial credentials are <%= color('admin', :info) %> / <%= color('#{kafo.param('katello_devel', 'admin_password').value}', :info) %>"
     end
+
+    def failure_message
+      say "  <%= color('Something went wrong!', :bad) %> Check the log for ERROR-level output"
+    end
+
+    def log_message(kafo)
+      # This is always useful, success or fail
+      log = kafo.config.app[:log_dir] + '/' + kafo.config.app[:log_name]
```
ekohl: I think this also works:
```suggestion
      log = kafo.config.log_file
```

hooks/boot/02-message-helpers.rb
```diff
@@ -6,7 +6,7 @@ def success_message
 
     def server_success_message(kafo)
       success_message
-      say "  * <%= color('Katello', :info) %> is running at <%= color('#{kafo.param('foreman', 'foreman_url').value}', :info) %>"
+      say "  * <%= color('Foreman', :info) %> is running at <%= color('#{param('foreman','foreman_url').value}', :info) %>"
```
ekohl: Does `param()` here work? This is not in the hook context I believe.

ekohl (commented):
Kafo has the interesting thing that it always loads `#{installer_dir}/hooks` ([source](https://github.com/theforeman/kafo/blob/93aaa70eff832fc6ccfbcbbf672f5cd775007420/lib/kafo/hooking.rb#L27)). If we always load the basic hooks as well, then we can change the `installer_dir` to `./` and leave out `./hooks` in `hook_dirs`. I initially implemented it this way to keep the separation, but it also requires workarounds (like 85f84a150a1d31bb0fef4eadd2679384fc95589a). My goal was always to have `installer_dir: .` in `katello.yaml` and `foreman-proxy-content.yaml`.

Speaking of foreman-proxy-content, that probably needs the same treatment.

Other implications of this are:

* Introduces `--reset-foreman-db` to the Katello scenario (and fpc if that is also modified). Has a lot in common with the `--reset` option Katello now has. Probably a good chance for unification.
* Introduces `--detailed-exitcodes` (a good thing IMHO)
* `katello/hooks/pre/21-check-hammer-credentials.rb` is now a duplicate of `hooks/pre/20-check-hammer-credentials.rb`
* `katello/hooks/pre/25-remove_apache_from_foreman_group.rb` is now a duplicate of `hooks/pre/25-remove_apache_from_foreman_group.rb`

Overall I really like the intent of this change.

I've been thinking about (unit) testing hooks. https://github.com/ekohl/foreman-installer/commit/1d80a8fad212cd324853078d5ac2494f7fdd8b8d#diff-ffd29a2cef868d990ba1fe2337fe9b23 can serve as inspiration.

ehelms:
Updated and filed an issue to resolve --reset and --reset-foreman-db (https://projects.theforeman.org/issues/28533) as a follow up. The code change seemed big enough to warrant it's own PR.

ekohl:
> Updated and filed an issue to resolve --reset and --reset-foreman-db (https://projects.theforeman.org/issues/28533) as a follow up. The code change seemed big enough to warrant it's own PR.

Fair enough. Put some comments there.

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -0,0 +1,11 @@
+module HookContextExtension
```
ekohl, line 1: Using the same name as in the Katello hook feels fragile. Probably best to rename the one in Katello to `KatelloHookContextExtension`.

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -0,0 +1,11 @@
...
+  def success_file
+    File.join(File.dirname(Kafo::KafoConfigure.config_file), '.installed')
```
ekohl: I think this should also work and be cleaner in testing:

```suggestion
    File.join(File.dirname(kafo.config.config_file), '.installed')
```

Ideally we'd avoid having a `.installed` file. I'd actually consider storing it in the scenario itself. However, I won't block on that.

ekohl (changes requested):
Can you also update https://github.com/theforeman/foreman-installer/blob/4838c1a8644d02017894f61984b0796581ee7e46/config/foreman-proxy-content.yaml#L10

ehelms:
Thanks for the feedback, updated

hooks/boot/02-message-helpers.rb
```diff
@@ -40,5 +40,15 @@ def new_install_message(kafo)
...
       say "      Initial credentials are <%= color('admin', :info) %> / <%= color('#{kafo.param('katello_devel', 'admin_password').value}', :info) %>"
     end
+
+    def failure_message
+      say "  <%= color('Something went wrong!', :bad) %> Check the log for ERROR-level output"
+    end
+
+    def log_message(kafo)
+      # This is always useful, success or fail
+      log = kafo.config.log_file
```
ekohl: Nit: Not sure if the separate variable really adds that much.

hooks/post/10-post_install_message.rb
```diff
@@ -1,30 +1,28 @@
...
+if module_enabled?('katello')
+  Kafo::MessageHelpers.certs_generate_command_message
+end
+
+if module_enabled?('katello_devel')
+  Kafo::MessageHelpers.dev_server_success_message(@kafo)
+  Kafo::MessageHelpers.dev_new_install_message(@kafo) if new_install?
 end
 
-# Puppetmaster?
```
ekohl, line 23: Any reason to drop this?

ekohl (commented):
Just thinking out loud. If you have multiple extensions you include, could you have something like:

```ruby
module HookContextExtension
  def post_message
    # ...
  end
end
Kafo::HookContext.send(:include, HookContextExtension)
```
Now in a later hook if you want to override this:
```ruby
module CustomizedMessagesExtension
  def post_message
    # ...
  end
end
Kafo::HookContext.send(:include, CustomizedMessagesExtension)
```

I'm wondering if we can avoid the whole MessageHelper class. It's also rather ugly that it's in the Kafo namespace.

hooks/post/10-post_install_message.rb
ehelms: Do we announce any other service that is running? Seemed odd to me. We are installing Foreman, that should be all we point users to and they can go figure out where things are running if needed from the smart proxy attached.

config/foreman-proxy-content.migrations/191218160614-update-hook-dirs.rb
```diff
@@ -0,0 +1,2 @@
+scenario[:hook_dirs] = ['/usr/share/foreman-installer/katello/hooks'] unless scenario[:hook_dirs].include?('/usr/share/foreman-installer/katello/hooks')
+scenario[:installer_dir] = '/usr/share/foreman-installer/katello' unless scenario[:installer_dir] == '/usr/share/foreman-installer/katello'
```
ehelms: @ekohl I am blanking a bit on how to handle these migrations given we need it to migrate to update the full-path but the tests will fail because it compares the pre and post migrated states to current config file which has relative paths.

ekohl: How about:
```ruby
if File.basename(scenario[:installer_dir]) == 'katello'
  scenario[:hook_dirs] = [File.join(scenario[:installer_dir], 'hooks')]
  scenario[:installer_dir] = File.dirname(scenario[:installer_dir])
end
```

spec/migration_spec.rb
```diff
@@ -26,24 +26,4 @@
       end
     end
   end
-
-  %w(foreman-proxy-content katello).each do |scenario_name|
```
ehelms: Note I dropped this test. After exploring it a bit and some thinking, this no longer seemed like it made sense to verify since we are actively wanting to migrate and change the config now as we make further changes.

ekohl: IMHO it does make sense to have a point in time to migrate from. Where this point is can be debated, but the goal is to simulate a user who installs Foreman 1.x and then upgrades to the latest scenario. We've seen in the past that old scenario + migrations can create a different scenario that we expected and has helped to catch some bugs. If anything, I'd actually like to have more of these scenarios to test our migrations. It's a lot of Ruby code we don't have other tests for.

ehelms: Is this covered by adding https://github.com/theforeman/foreman-installer/pull/437 ?

hooks/boot/02-message-helpers.rb
```diff
@@ -6,7 +6,7 @@ def success_message
 
     def server_success_message(kafo)
       success_message
-      say "  * <%= color('Katello', :info) %> is running at <%= color('#{kafo.param('foreman', 'foreman_url').value}', :info) %>"
+      say "  * <%= color('Foreman', :info) %> is running at <%= color('#{kafo.param('foreman','foreman_url').value}', :info) %>"
```
ekohl: Rubocop will should at you for removing the space :)
```suggestion
      say "  * <%= color('Foreman', :info) %> is running at <%= color('#{kafo.param('foreman', 'foreman_url').value}', :info) %>"
```

spec/migration_spec.rb
ekohl: Yes, but I'd like to merge that before dropping it here to make sure we still have decent coverage. Or keep it in for now.

ehelms:
Updated and added test back with fixture updates

ekohl (approved):
Untested, but I think this looks correct. This is a good step towards a truly unified foreman-installer

hooks/boot/02-message-helpers.rb
```diff
@@ -6,7 +6,7 @@ def success_message
 
     def server_success_message(kafo)
       success_message
-      say "  * <%= color('Katello', :info) %> is running at <%= color('#{kafo.param('foreman', 'foreman_url').value}', :info) %>"
+      say "  * <%= color('Foreman', :info) %> is running at <%= color('#{kafo.param('foreman', 'foreman_url').value}', :info) %>"
```
evgeni, line 9: This changes the message from "Katello" to "Foreman" (while keeping "Katello" in the devel scenarios), which I think is fine, but might be unexpected by users? Should this be the scenario name instead?

ekohl: If it uses the scenario name, then it might also solve the downstream branding and remove the need for message helpers.

hooks/boot/01-kafo-hook-extensions.rb
ekohl: This was marked as resolved, but in `foreman-installer/katello/hooks/boot/01-kafo-hook-extensions.rb` it still uses the same name.

hooks/boot/02-message-helpers.rb
ehelms: The scenario name usage makes the assumption that thing running is always congruent with the name. I don't want to confuse users further with things like "Luna is running at" "Foreman Proxy with Content is running at". The software that's running in this case is Foreman (with plugins).

ekohl: Not for this PR, but we can think about storing that in the scenario. New fields can be added and I think I'd rather deal with that in scenarios than in ruby code.

PR #437, opened 2019-12-18
Add migration test from post-merge to current

ehelms:
I wanted to explore the idea of testing old config and answers to latest state to ensure nothing is missed. I purposefully only did katello to start and left it failing to show the kinds of things this finds. 

 1) Does this make sense?
 2) Are the things being caught useful?

Is there a more robust test perhaps that can be done with a fully built out answer and config file generated with all parameters? Perhaps both tests would be useful and this is the first step.

### DIFF
--- a/spec/fixtures/foreman-1.21/katello-answers.yaml
+++ b/spec/fixtures/foreman-1.21/katello-answers.yaml
@@ -1,0 +1,85 @@
+---
+certs:
+  generate: true
+  deploy: true
+  group: foreman
+katello:
+  package_names:
+  - katello
+  - tfm-rubygem-katello
+foreman:
+  initial_organization: Default Organization
+  initial_location: Default Location
+  configure_epel_repo: false
+  configure_scl_repo: false
+  max_keepalive_requests: 10000
+  ssl: true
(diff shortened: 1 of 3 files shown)


ekohl:
> Does this make sense?

Yes!

> Are the things being caught useful?

Not entirely since we expect changes. Probably better to have a file with the expected output so we can manually apply changes that we expect. Some things it finds are because of 3c58237084c5060a93472f486c49d5d463dd6d27. Having a migration for that doesn't really make sense because after running the installer, the value will be restored anyway.

> Is there a more robust test perhaps that can be done with a fully built out answer and config file generated with all parameters? Perhaps both tests would be useful and this is the first step.

I'd say a full answer file as a pipeline box post installation (so with all the answers stored) makes sense, but in addition to a minimal one.

ehelms:
The more I think about it, the more I don't know how to pull this off in a reliable way. Given an expanded answer's file is not easy to generate and is dependent upon the environment within which it is generated comparing two to ensure correctness seems like a hard problem. There will be parameters worth caring about and those we should ignore. But it's a long list of parameters.

PR #440, opened 2020-01-06
Fixes #28653: Introduce install_package command

ehelms:
This also combines all Kafo extension functions into a single hook.

### DIFF
--- a/hooks/boot/01-kafo-hook-extensions.rb
+++ b/hooks/boot/01-kafo-hook-extensions.rb
@@ -12,6 +12,15 @@ module HookContextExtension
     !File.exist?(success_file)
   end
 
+  def ensure_package(package, state = 'installed')
+    unless ['installed', 'absent', 'latest'].include?(state)
+      fail_and_exit('Incorrect package state supplied to ensure_package')
+    end
+
+    bin_path = Kafo::PuppetCommand.search_puppet_path('puppet')
+    execute("#{bin_path} resource package #{package} ensure=#{state}")
+  end
+
   def fail_and_exit(message)
     log_and_say :error, message
     exit 1
@@ -31,9 +40,7 @@ module HookContextExtension
               level
             end
 
-    # \ and ' characters could cause trouble in ERB, make sure to escape them
-    escaped_message = message.gsub('\\', '\\\\\\').gsub("'", %q{\\\'})
-    say "<%= color('#{escaped_message}', :#{style}) %>" if do_say
+    say HighLine.color(message, style.to_sym) if do_say
     Kafo::KafoConfigure.logger.send(level, message) if do_log
   end
 
--- a/katello/hooks/pre/29-install_postgresql.rb
+++ b/katello/hooks/pre/29-install_postgresql.rb
@@ -1,10 +1,4 @@
 # Working around https://tickets.puppetlabs.com/browse/PUP-2169
 if local_postgresql? && facts[:os][:release][:major] == '7' && !app_value(:upgrade)
-  package = 'rh-postgresql10-postgresql-server'
-
-  `rpm -q #{package}`
-
-  unless $?.success?
-    execute("yum -y install #{package}")
-  end
+  ensure_package('rh-postgresql10-postgresql-server', 'installed')
 end
--- a/katello/hooks/pre/30-upgrade.rb
+++ b/katello/hooks/pre/30-upgrade.rb
@@ -37,11 +37,12 @@ def postgresql_10_upgrade
   start_postgresql
   (name, owner, enconding, collate, ctype, privileges) = %x{runuser postgres -c 'psql -lt | grep -E "^\s+postgres"'}.chomp.split('|').map(&:strip)
   stop_services
-  execute('yum -y install rh-postgresql10-postgresql-server')
+  ensure_package('rh-postgresql10-postgresql-server', 'installed')
   execute(%Q{scl enable rh-postgresql10 "PGSETUP_INITDB_OPTIONS='--lc-collate=#{collate} --lc-ctype=#{ctype} --locale=#{collate}' postgresql-setup --upgrade"})
-  execute('yum -y remove postgresql postgresql-server')
+  ensure_package('postgresql', 'absent')
+  ensure_package('postgresql-server', 'absent')
   execute('rm -f /etc/systemd/system/postgresql.service')
-  execute('yum -y install rh-postgresql10-syspaths')
+  ensure_package('rh-postgresql10-syspaths', 'installed')
 end
 
 def upgrade_step(step, options = {})


hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -1,11 +1,70 @@
...
+              'bad'
+            when :debug
+              'yellow'
+            else
+              level
+            end
+
+    # \ and ' characters could cause trouble in ERB, make sure to escape them
+    escaped_message = message.gsub('\\', '\\\\\\').gsub("'", %q{\\\'})
+    say "<%= color('#{escaped_message}', :#{style}) %>" if do_say
```
ekohl: How about directly calling it so we don't need to escape:
```suggestion
    say HighLine.color(message, style.to_sym) if do_say
```

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -1,11 +1,70 @@
...
+    Kafo::KafoConfigure.logger.send(level, message) if do_log
+  end
+
+  def execute(commands, do_say = true, do_log = true)
+    commands = commands.is_a?(Array) ? commands : [commands]
```
ekohl: We hardly use arrays and in most cases I think it actually doesn't make sense.

ekohl (commented):
I don't see the actual definition of install_package.

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -1,11 +1,72 @@
...
     File.join(File.dirname(kafo.config.config_file), '.installed')
   end
 
   def new_install?
     !File.exist?(success_file)
   end
+
+  def install_package(package)
+    execute("puppet resource package #{package} ensure=installed")
+  end
```
ehelms, line 22: It's here now, not sure how I missed the key part of the PR!

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -1,11 +1,72 @@
 module HookContextExtension
+  # FIXME: remove when #23332 is released
```
ekohl: I'd like to keep this minimal. Didn't we have a separate PR for merging the hooks? Could you perhaps separate it into two commits at least?

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -1,11 +1,72 @@
...
+
   def success_file
     File.join(File.dirname(kafo.config.config_file), '.installed')
   end
 
   def new_install?
     !File.exist?(success_file)
   end
+
+  def install_package(package)
```
ekohl: Bikeshedding: would `ensure_package_installed` be more better? When unfamiliar with it, you may want to still check if it's installed to optimize, even though it's not needed.

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -1,11 +1,72 @@
...
+    execute("puppet resource package #{package} ensure=installed")
```
ekohl: Puppet may not be on `$PATH`. Calling `sudo foreman-installer` will not execute `/etc/profile.d/puppet-agent.sh`. For this Kafo has the ugly workaround that's [search_puppet_path](https://github.com/theforeman/kafo/blob/bbf9288473e0f1af9892c9cba0dc665be4ea51b3/lib/kafo/puppet_command.rb#L37-L42)

ehelms: So you are wanting me to copy that logic to here?

hooks/boot/01-kafo-hook-extensions.rb
ehelms: I can break out yet another PR for merging these extensions. Please try to suggest those types of changes as early as you can. This change blocks multiple other changes and is dragging on.

hooks/boot/01-kafo-hook-extensions.rb
ehelms: That can be a follow up change if yo uwant to file an issue.

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -1,11 +1,73 @@
...
   def success_file
     File.join(File.dirname(kafo.config.config_file), '.installed')
   end
 
   def new_install?
     !File.exist?(success_file)
   end
+
+  def ensure_package_installed(package)
+    bin_path = Kafo::PuppetCommand.search_puppet_path('puppet')
```
ehelms, line 20: I think I can call this this way from the hooks.

ekohl: That was going to be my suggestion.

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -1,11 +1,73 @@
...
+    execute("#{bin_path} resource package #{package} ensure=installed")
```
ekohl: Now that I read more, would it make sense to have the ensure state as a parameter? There's still a `yum -y remove` in the same code block below and `ensure=absent` would allow us to abstract that away as well. I can also imagine we may want a `ensure=latest`
```suggestion
    execute("#{bin_path} resource package #{package} ensure=#{ensure}")
```

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -1,11 +1,73 @@
...
+  end
+
+  def fail_and_exit(message)
+    log_and_say :error, message
+    exit 1
+  end
+
+  def local_postgresql?
+    param_value('foreman', 'db_manage') || param_value('katello', 'candlepin_manage_db')
```
ekohl: For the backlog: pulpcore can also require a local postgres.

hooks/boot/01-kafo-hook-extensions.rb
ekohl: I don't require a separate PR, but splitting changes into logical separate commits gives a cleaner git log. Personally I often use that. So one commit to move everything from Katello and another to add the additional method.

ehelms:
This now requires -- https://github.com/theforeman/foreman-installer/pull/442

ehelms:
Rebased with the hook combining now in. This does add the state parameter (with a validation check) to make the `ensure_package` function more usable.

evgeni (approved):
Executing Puppet is a tad slower than using rpm/yum directly ("installing" an installed package takes 6s, removing a removed one 2s with Puppet, wheras rpm detects that something is already installed in 0.04s ;)) but I think that's fine as long we're not installing/purging huge lists of packages here.

ehelms:
That's slooowww by comparison. But should only be used sparingly so lets give it a try.

evgeni:
> That's slooowww by comparison. But should only be used sparingly so lets give it a try.

Yeah, as long we're not calling that func 1000 times per installer run, we're OK.

ekohl:
A lot of that it Puppet initialization time. If we need that, it may actually be faster to write small puppet files for migrations instead.

PR #442, opened 2020-01-08
Fixes #28688: Combine hook extensions into a single file

ehelms:


ekohl (approved):
Since this is just moving things, I won't comment on further improvements and let's get this in now.

PR #443, opened 2020-01-08
Fixes #28694: Move upgrade flag and functions to main hooks

ehelms:


ekohl (commented):
I'm actually opposed to having a `--upgrade` flag at all. For historical reasons it may be needed in katello and we can't drop it immediately (and we've gotten rid of mos), but I'd like to avoid introducing it for the Foreman scenario. Can we check if rh-postgresql10-syspaths is installed instead to detect if it's already been upgraded? Even if we don't, the user has no choice in upgrading to postgresql 10 and requiring them to pass an option is just asking for issues.

ehelms:
how would you think about detecting an "upgrade step" as complete (using Postgresql upgrade as the example):  

1) use something about the action to tell you, e.g. checking that rh-postgresql10-syspaths is installed
2) adding a new scenario config value that is a list of strings that represent "upgrade" actions and if they are complete, e.g.

katello.yaml
```
installer_dir: .
upgrade_steps:
  - postgresql10
```

3) Other

PR #444, opened 2020-01-08
Fixes #28697: Add Postgres upgrade to Foreman scenario

ehelms:


ehelms:
This is designed to run Postgres upgrade in all situations where Foreman is present and its RHEL 7 regardless of upgrade parameter.

hooks/pre/30-upgrade-postgresql.rb
```diff
@@ -0,0 +1,18 @@
+def postgresql_10_upgrade
+  execute('foreman-maintain service start --only=postgresql')
```
evgeni: Is `f-m` required on a plain Foreman machine these days? I don't think it is.

hooks/pre_validations/30-upgrade-postgresql.rb
```diff
@@ -0,0 +1,32 @@
+def check_postgresql_storage
+  #Ensure there is at least 1x /var/lib/pgsql free disk space available on the main filesystem
+  current_postgres_dir = '/var/lib/pgsql'
+  new_postgres_dir = '/var/opt/rh/rh-postgresql10/lib/pgsql'
+
+  begin
+    postgres_size = `du -b -s #{current_postgres_dir}`.split[0].to_i
+
+    if available_space(new_postgres_dir) < postgres_size
+      fail_and_exit "The postgres upgrade requires at least #{(postgres_size / 1024) / 1024} MB of storage."
```
evgeni:
```suggestion
      fail_and_exit "The PostgreSQL upgrade requires at least #{(postgres_size / 1024) / 1024} MB of storage."
```

hooks/pre/30-upgrade-postgresql.rb
```diff
@@ -0,0 +1,18 @@
...
+  (name, owner, enconding, collate, ctype, privileges) = %x{runuser postgres -c 'psql -lt | grep -E "^\s+postgres"'}.chomp.split('|').map(&:strip)
+  execute('foreman-maintain service stop')
+  execute('yum -y install rh-postgresql10-postgresql-server')
+  execute(%Q{scl enable rh-postgresql10 "PGSETUP_INITDB_OPTIONS='--lc-collate=#{collate} --lc-ctype=#{ctype} --locale=#{collate}' postgresql-setup --upgrade"})
+  execute('yum -y remove postgresql postgresql-server')
+  execute('rm -f /etc/systemd/system/postgresql.service')
+  execute('yum -y install rh-postgresql10-syspaths')
+end
+
+if local_postgresql? && facts[:os][:release][:major] == '7' && module_enabled?('foreman')
```
evgeni: Should this check for os family being RedHat? Not that I envision us to run on Debian 7, but just to be thorough… :)

hooks/pre_validations/30-upgrade-postgresql.rb
```diff
@@ -0,0 +1,32 @@
...
+def available_space(directory = nil)
+  directory = '/' if directory.nil?
+  mountpoints = facts[:mountpoints]
+  until (mountpoint = mountpoints[directory.to_sym])
+    directory = File.dirname(directory)
+  end
+  mountpoint[:available_bytes]
+end
+
+if local_postgresql? && facts[:os][:release][:major] == '7' && module_enabled?('foreman')
```
evgeni: Same as in the other file :)

evgeni (changes requested):

hooks/pre/30-upgrade-postgresql.rb
ehelms: I will need to update the packaging to require it by the foreman-installer. Not so much because of this line, but I need to be able to stop all services for all scenarios.

ekohl: foreman-maintain isn't packaged for Debian so the installer can't depend on it.

ehelms: This hook is technically limited to EL7 only, so the EL7 installer packaging can depend on it.

hooks/pre_validations/30-upgrade-postgresql.rb
ehelms: Added a helper function to check for this.

hooks/pre/30-upgrade_postgresql.rb
```diff
@@ -0,0 +1,18 @@
...
+  execute('foreman-maintain service stop')
+  execute('yum -y install rh-postgresql10-postgresql-server')
+  execute(%Q{scl enable rh-postgresql10 "PGSETUP_INITDB_OPTIONS='--lc-collate=#{collate} --lc-ctype=#{ctype} --locale=#{collate}' postgresql-setup --upgrade"})
+  execute('yum -y remove postgresql postgresql-server')
+  execute('rm -f /etc/systemd/system/postgresql.service')
+  execute('yum -y install rh-postgresql10-syspaths')
+end
+
+if local_postgresql? && rhel7? && module_enabled?('foreman')
+  `rpm -q rh-postgresql10-syspaths`
```
evgeni: Wouldn't be checking `/var/opt/rh/rh-postgresql10/lib/pgsql/data` exists be "cheaper" (no shell out to `rpm`)?

hooks/pre_validations/30-upgrade_postgresql.rb
```diff
@@ -0,0 +1,32 @@
+def check_postgresql_storage
+  #Ensure there is at least 1x /var/lib/pgsql free disk space available on the main filesystem
+  current_postgres_dir = '/var/lib/pgsql'
+  new_postgres_dir = '/var/opt/rh/rh-postgresql10/lib/pgsql'
+
+  begin
+    postgres_size = `du -b -s #{current_postgres_dir}`.split[0].to_i
+
+    if available_space(new_postgres_dir) < postgres_size
+      fail_and_exit "The PostgreSQL upgrade requires at least #{(postgres_size / 1024) / 1024} MB of storage."
```
evgeni: The message should probably tell the user *where* the storage is required as the user is not aware of the move to `/var/opt/rh/rh-postgresql10/lib/pgsql` just yet.

hooks/pre_validations/30-upgrade_postgresql.rb
```diff
@@ -0,0 +1,32 @@
+def check_postgresql_storage
+  #Ensure there is at least 1x /var/lib/pgsql free disk space available on the main filesystem
+  current_postgres_dir = '/var/lib/pgsql'
+  new_postgres_dir = '/var/opt/rh/rh-postgresql10/lib/pgsql'
+
+  begin
+    postgres_size = `du -b -s #{current_postgres_dir}`.split[0].to_i
```
evgeni:
```suggestion
    postgres_size = `du --bytes --summarize #{current_postgres_dir}`.split[0].to_i
```

hooks/pre/30-upgrade_postgresql.rb
evgeni: Also, given this code exists in two places, should it be a helper function like `needs_postgresql_scl_upgrade?`

hooks/pre/30-upgrade_postgresql.rb
```diff
@@ -0,0 +1,15 @@
+def postgresql_10_upgrade
+  execute('foreman-maintain service start --only=postgresql')
+  (name, owner, enconding, collate, ctype, privileges) = %x{runuser postgres -c 'psql -lt | grep -E "^\s+postgres"'}.chomp.split('|').map(&:strip)
+  execute('foreman-maintain service stop')
```
ehelms: Checking, @ekohl are you OK with this being here with a requirement on RPM foreman-installer packaging to require foreman-maintain? This code is gated on `el7?` so Debian is unaffected.

evgeni: While I'm not Ewoud, I'd be OK with that. But I think I'd prefer the file to be called `30-el7_upgrade_postgresql` to denote it's only relevant on el7.

evgeni (approved):

ehelms:
Files renamed

ekohl (commented):
`katello/hooks/pre_validations/32-upgrade.rb` can now be removed, right? That's now transformed into `hooks/pre_validations/30-el7_upgrade_postgresql.rb`.

Edit: also a merge conflict now.

ehelms:
Welp, this happened http://mirror.centos.org/centos/7/sclo/x86_64/rh/rh-postgresql12/ so I suppose I'll go and test that in case I can just change this to that.

evgeni:
Yeah, it'd be cool if we can go straight to 12 with 2.0

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -40,6 +40,10 @@ def local_redis?
     (foreman_server? && !param_value('foreman', 'jobs_sidekiq_redis_url')) || param_value('foreman_proxy::plugin::pulp', 'pulpcore_enabled')
   end
 
+  def needs_postgresql_scl_upgrade?
+    !File.exist?('/var/opt/rh/rh-postgresql12/lib/pgsql/data') && File.exist?('/var/lib/pgsql/data')
```
ehelms, line 44: @evgeni I had to add a condition here for this to work on fresh installations. Before adding the `&&` part, it attempted an upgrade on a fresh install.

ehelms:
This has been updated instead to PostgreSQL 12. @tbrisker for your awareness that we'd be switching to PG 12 instead of 10 now that it's published (http://mirror.centos.org/centos/7/sclo/x86_64/rh/rh-postgresql12/). Bats tests passed for me. 

This scratch build can be used to test: http://koji.katello.org/koji/taskinfo?taskID=272814

ehelms:
My testing of this with Pulp 3 yielded no errors.

hooks/boot/01-kafo-hook-extensions.rb
evgeni: Makes sense.

This will still try to upgrade if the user installed postgres before running the installer, but I think that's fine as being labeled "unsupported"?

evgeni (approved):

evgeni:
I am not merging this in case you first want to fix Katello nightly pipelines.

ehelms:
Niightly is greeeen. This will, however, need https://github.com/theforeman/foreman-packaging/pull/4577 resolved as well to not fail on nightly tests.

ekohl (approved):
Could you make a post to discourse with some notes for users who have already upgraded to SCL PG 10?

PR #447, opened 2020-01-13
Fixes #28738: Reset bind host if :: is present

ehelms:
This is necessary with smart proxy moving away from Ruby 2.0 via
RPM packaging changes to use SCLs.

### DIFF
--- a/config/foreman.migrations/20200113133837_reset_bind-host-foreman-proxy.rb
+++ b/config/foreman.migrations/20200113133837_reset_bind-host-foreman-proxy.rb
@@ -1,0 +1,7 @@
+if answers['foreman_proxy'].is_a?(Hash) &&
+    answers['foreman_proxy']['bind_host'].is_a?(Array) &&
+    answers['foreman_proxy']['bind_host'].include?('::') &&
+    facts[:os][:release][:major] == '7' &&
+    facts[:os][:family] == 'RedHat'
+  answers['foreman_proxy']['bind_host'].delete
+end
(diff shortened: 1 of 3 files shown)


ekohl (commented):
You're not guaranteed it's an array. A string is also a valid input. Technically that's deprecated since 1.15 but we haven't made the change yet to drop it. I'd suggest checking with `.is_a?(Array)` just to be sure, also in case it's `nil`.

Technically this was a valid config on Debian-based systems to bind to IPv6-only and only affected EL7. Though it's quite unlikely users actually ran into this, it may make sense to add an OS check in `config/foreman.migrations`.

ehelms:
This check will com ein handy for that -- https://github.com/theforeman/foreman-installer/pull/444/files#diff-c6fdf0194de166ebab0dde67cd968490R32

Would it be safe to reset the value of this in all cases? Or are there instances where this will have been configured by a user?

ehelms:
Added array and EL7 check to the migrations.

config/foreman.migrations/20200113133837_reset_bind-host-foreman-proxy.rb
```diff
@@ -0,0 +1,6 @@
+if answers['foreman_proxy']['bind_host'].is_a?(Array) &&
```
ekohl: `answers['foreman_proxy']` can be `false` in which case it crashes. I don't think we can use `answers.dig('foreman_proxy', 'bind_host')` in Ruby 2.0 yet, but that would be the defensive way.

ehelms: Added an `is_a?(Hash)` check

ekohl (approved):

PR #453, opened 2020-01-16
Fixes #28779: Require English when using CHILD_STATUS

ehelms:


ehelms:
Encountered as a result of the rubocop addition. On newer rubies this explicit requires is not required.

mmoll:
merged, thanks @ehelms!

PR #458, opened 2020-01-23
Fixes #28828 - Enable pulpcore

wbclark:


config/katello.migrations/200123161606-enable-pulpcore.rb
```diff
@@ -0,0 +1,3 @@
+# enable pulpcore
+
+answers['foreman_proxy']['plugin']['pulp']['pulpcore_enabled'] = true if answers['foreman_proxy']
```
ekohl: The key is the class name. You should also be careful that it can be a boolean or a hash. Something like this probably makes sense.

```suggestion
cls = 'foreman_proxy::plugin::pulp'
if answers[cls].is_a?(Hash)
  answers[cls]['pulpcore_enabled'] = true
elsif answers[cls]
  answers[cls] = { 'pulpcore_enabled' => true }
end
```

config/katello.migrations/200123161606-enable-pulpcore.rb
```diff
@@ -0,0 +1,3 @@
+# enable pulpcore
```
ekohl: I'd leave out this comment since the name already describes it.

ekohl (changes requested):
I think you should also detect upgrades and force it to use Pulp 2 for the content types. You can do that setting the answer to true by default in the answers file:

https://github.com/theforeman/foreman-installer/blob/6e12ef181d9bad0dc4db98409f093c7195ee86dc/config/katello-answers.yaml#L40

With that you can see that `answers['foreman_proxy::plugin::pulp']['pulpcore_enabled']` is already true on fresh installs but false on upgrades or answers['foreman_proxy::plugin::pulp'] is `true`.

config/katello.migrations/200123161606-enable-pulpcore.rb
```diff
@@ -0,0 +1,5 @@
+cls = 'foreman_proxy::plugin::pulp'
```
ehelms: Where possible, I prefer, for readability, not to use shortened names for things. So something like `puppet_class` as the variable name is more expressive.

ekohl: Probably right. On a mostly unrelated note, I've been thinking about a helper in kafo to simplify this pattern but haven't been able to come up with one that actually makes it clear.

config/katello.migrations/200123161606-enable-pulpcore.rb
```diff
@@ -0,0 +1,5 @@
...
+if answers[cls].is_a?(Hash)
+  answers[cls]['pulpcore_enabled'] = true
```
ehelms: This is the default so I don't think we need this if-condition. @ekohl ?

This migration will also need to set the following on upgrade:

```
use_pulp_2_for_file: true
use_pulp_2_for_docker: true
```

ekohl: > This is the default so I don't think we need this if-condition. @ekohl ?

The code default is off: https://github.com/theforeman/puppet-foreman_proxy/blob/32b61e3f6a4ee150d65413c201b5cb902ad7ecd3/manifests/plugin/pulp/params.pp#L11

If you change supplied katello-answers.yaml it gives you a chance to detect upgrades.

> use_pulp_2_for_file: true
> use_pulp_2_for_docker: true

Note that these are on `katello`, not `foreman_proxy::plugin::pulp`.

wbclark: @ehelms @ekohl thanks.

> If you change supplied katello-answers.yaml it gives you a chance to detect upgrades.

@ekohl can you elaborate on this point a bit? what conditional can i use to detect upgrades and set the right params?

ekohl: Currently the answers looks like this:
```yaml
foreman_proxy::plugin::pulp: true
```

For users who already ran the installer this gets expanded to:
```yaml
foreman_proxy::plugin::pulp:
  enabled: true
  pulpnode_enabled: false
  # and others
```

I don't know for sure if the `pulpcore_enabled` default is already loaded during migrations, but probably not. If it is, it's `false` since that's the module default.

Now if you change the answers file to:
```yaml
foreman_proxy::plugin::pulp:
  pulpcore_enabled: true
```

This gives you the ability to distinguish a fresh install.

In the migration you can have the following cases that the installer can see and shouldn't crash on:
```ruby
answers['foreman_proxy::plugin::pulp'] == false # user turned off the module
answers['foreman_proxy::plugin::pulp'] == true # default in a previous version
answers['foreman_proxy::plugin::pulp'].is_a?(Hash) == true # the following cases imply this
answers['foreman_proxy::plugin::pulp']['pulpcore_enabled'] == nil # upgrading user
answers['foreman_proxy::plugin::pulp']['pulpcore_enabled'] == false # module default
answers['foreman_proxy::plugin::pulp']['pulpcore_enabled'] == true # fresh install
```

That makes me think the following should safely cover all cases:
```ruby
puppet_class = 'foreman_proxy::plugin::pulp'
param = 'pulpcore_enabled'

def upgrade
  answers[puppet_class][param] = true
  return unless answers['katello']
  answers['katello'] = {} unless answers['katello'].is_a?(Hash)
  answers['katello']['use_pulp_2_for_file'] = true
  answers['katello']['use_pulp_2_for_docker'] = true
end

if answers[puppet_class].is_a?(Hash)
  upgrade unless answers[puppet_class][param]
elsif answers[puppet_class]
  answers[puppet_class][param] = {}
  upgrade
end
```

You can see it ignores the fresh install which we can now safely detect. If the module is on in some way, we upgrade.

Writing this makes me think we need a migration helper in kafo:
```ruby
def set_answer(puppet_class, parameter, value)
  answers[puppet_class] = {} unless answers[puppet_class].is_a?(Hash)
  answers[puppet_class][parameter] = value
end
```

wbclark: @ekohl a big, sincere thanks for the detailed breakdown and recommendations.

config/katello.migrations/200123161606-enable-pulpcore.rb
```diff
@@ -0,0 +1,17 @@
+puppet_class = 'foreman_proxy::plugin::pulp'
```
ekohl: Maybe this needs to be `PUPPET_CLASS` so it's available within the method

ekohl (changes requested):
You're missing the change to the default answers.

config/katello.migrations/200123161606-enable-pulpcore.rb
wbclark: Once I made the change to the default answers file, that issue disappeared so I think it's OK

config/katello.migrations/200123161606-enable-pulpcore.rb
```diff
@@ -0,0 +1,17 @@
...
+param = 'pulpcore_enabled'
+
+def upgrade
+  answers[puppet_class][param] = true
+  return unless answers['katello']
+  answers['katello'] = {} unless answers['katello'].is_a?(Hash)
+  answers['katello']['use_pulp_2_for_file'] = true
+  answers['katello']['use_pulp_2_for_docker'] = true
+end
```
ehelms, line 15: We also need to set https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/manifests/init.pp#L131 to false so that on upgrade `/pulp/isos` is being served from Pulp 2

config/katello.migrations/200123161606-enable-pulpcore.rb
ekohl: I think it's now following a different code path where upgrade is not actually executed. That's one way of avoiding build failures, but won't help much in production ;)

config/katello.migrations/200123161606-enable-pulpcore.rb
```diff
@@ -0,0 +1,18 @@
+PUPPET_CLASS = 'foreman_proxy::plugin::pulp'
+PARAM = 'pulpcore_enabled'
+
+def upgrade
+  answers[PUPPET_CLASS][PARAM] = true
+  return unless answers['katello']
```
ekohl: This should be a block now (and indent the rest):
```suggestion
  unless answers['katello']
```

config/katello.migrations/200123161606-enable-pulpcore.rb
```diff
@@ -0,0 +1,18 @@
...
+  answers['katello'] = {} unless answers['katello'].is_a?(Hash)
+  answers['katello']['use_pulp_2_for_file'] = true
+  answers['katello']['use_pulp_2_for_docker'] = true
+  answers['foreman_proxy_content']['proxy_pulp_isos_to_pulpcore'] = false
```
ekohl: This needs a similar guard as lines 6 and 7 for `katello` to ensure it's safe. It's unlikely, but I like to be defensive in my code.

config/katello.migrations/200123161606-enable-pulpcore.rb
wbclark: Good point, thanks :)

config/katello.migrations/200123161606-enable-pulpcore.rb
wbclark: Done, thanks

config/katello.migrations/200123161606-enable-pulpcore.rb
wbclark: I've created separate `unless` blocks for `katello` and `foreman_proxy_content`. Thanks

ekohl:
Looks like you need to make rubocop happy. It'd also make me happy if you added a test to https://github.com/theforeman/foreman-installer/blob/develop/spec/migration_spec.rb which migrates an existing setup and verify the upgrade method does what it's supposed to.

ekohl (approved):

ekohl:
@ehelms since I was quite involved with the code, I'll let you approve and merge.

ehelms (approved):
Looks good!

PR #459, opened 2020-01-29
Fixes #28839 - Reset pulpcore db with reset

sjha4:
To test:
1. Spin up a nightly.
2. Run foreman-installer --foreman-proxy-plugin-pulp-pulpcore-enabled=true on the nightly box to enable pulpcore.
3. Pull in these changes. I did these changes directly on the installed foreman-installer files on the nightly under /usr/share/foreman-installer
4. Run foreman-installer --reset

You will need to be on the latest foreman-maintain.

### DIFF
--- a/hooks/boot/01-kafo-hook-extensions.rb
+++ b/hooks/boot/01-kafo-hook-extensions.rb
@@ -41,7 +41,11 @@ module HookContextExtension
   end
 
   def local_redis?
-    (foreman_server? && !param_value('foreman', 'jobs_sidekiq_redis_url')) || param_value('foreman_proxy::plugin::pulp', 'pulpcore_enabled') || devel_scenario?
+    (foreman_server? && !param_value('foreman', 'jobs_sidekiq_redis_url')) || pulpcore_enabled? || devel_scenario?
+  end
+
+  def pulpcore_enabled?
+    param_value('foreman_proxy_plugin_pulp', 'pulpcore_enabled')
   end
 
   def needs_postgresql_scl_upgrade?
--- a/katello/hooks/pre/10-reset_feature.rb
+++ b/katello/hooks/pre/10-reset_feature.rb
@@ -8,6 +8,7 @@ def reset
     reset_database
     reset_candlepin
     reset_pulp
+    reset_pulpcore if pulpcore_enabled?
 
   else
     Kafo::KafoConfigure.logger.warn 'Katello not installed yet, can not drop database!'
@@ -133,6 +134,26 @@ def reset_pulp
   execute(
     'rm -rf /var/lib/pulp/{distributions,published,repos}/*'
   )
+end
+
+def load_pulpcore_config
+  db_config = {}
+  db_config[:host] = param_value('foreman_proxy_content', 'pulpcore_postgresql_host') || 'localhost'
+  db_config[:port] = param_value('foreman_proxy_content', 'pulpcore_postgresql_port') || 5432
+  db_config[:database] = param_value('foreman_proxy_content', 'pulpcore_postgresql_db_name') || 'pulpcore'
+  db_config[:username] = param_value('foreman_proxy_content', 'pulpcore_postgresql_user')
+  db_config[:password] = param_value('foreman_proxy_content', 'pulpcore_postgresql_password')
+  db_config
+end
+
+def reset_pulpcore
+  Kafo::KafoConfigure.logger.info 'Dropping Pulpcore database!'
+  config = load_pulpcore_config
+  if remote_host?(config[:host])
+    empty_database!(config)
+  else
+    execute("sudo -u postgres dropdb #{config[:database]}")
+  end
 end
 
 def remote_host?(hostname)


katello/hooks/pre/10-reset_feature.rb
```diff
@@ -135,6 +136,11 @@ def reset_pulp
   )
 end
 
+def reset_pulcore
+  Kafo::KafoConfigure.logger.info 'Dropping Pulpcore database!'
+  execute('sudo runuser - postgres -c "dropdb pulpcore"')
```
ehelms: Will this error if there is no pulpcore installed? We should check to be safe. There is code in (https://github.com/theforeman/foreman-installer/blob/develop/hooks/boot/01-kafo-hook-extensions.rb#L40) at the end that could be turned into a helper method.

ekohl: Why does it use `sudo` and `runuser`? Just using `runuser` should be sufficient, right?

sjha4: @ehelms : For some reason, checking for param_value('foreman_proxy::plugin::pulp', 'pulpcore_enabled') returns nil inside this .. I checked the katello-answers.yaml in my setup and the pulpcore_enabled is set to True..

sjha4: So @jlsherrill helped me figure that param_value('foreman_proxy_plugin_pulp', 'pulpcore_enabled') works..Updated with the same..

katello/hooks/pre/10-reset_feature.rb
```diff
@@ -135,6 +136,11 @@ def reset_pulp
   )
 end
 
+def reset_pulcore
```
jjeffers: Kind of petty but should that be `reset_pulpcore`?

sjha4: Updated.. :+1:

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -37,7 +37,11 @@ def local_postgresql?
   end
 
   def local_redis?
-    (foreman_server? && !param_value('foreman', 'jobs_sidekiq_redis_url')) || param_value('foreman_proxy::plugin::pulp', 'pulpcore_enabled')
+    (foreman_server? && !param_value('foreman', 'jobs_sidekiq_redis_url')) || param_value('foreman_proxy_plugin_pulp', 'pulpcore_enabled')
```
ehelms: The end of this line can be updated with the helper function `pulpcore_enabled?`

katello/hooks/pre/10-reset_feature.rb
```diff
@@ -135,6 +136,11 @@ def reset_pulp
   )
 end
 
+def reset_pulpcore
+  Kafo::KafoConfigure.logger.info 'Dropping Pulpcore database!'
+  execute('runuser - postgres -c "dropdb pulpcore"') if pulpcore_enabled?
```
ehelms: Let's move the `if` to the top of the function as a guard. No use logging we are dropping the pulpcore database if we actually aren't.

ehelms: Also, should this function look more like https://github.com/theforeman/foreman-installer/pull/459/files#diff-ed86b943421faaceae73f8a3bff3e2edR59 since that uses code already present and covers remote database?

sjha4: I raised an issue the other day around remote dbs..I am not sure that is supported yet..
https://projects.theforeman.org/issues/28901

ehelms: It will be, so I would opt for supporting it to not have to come re-engineer this code. @wbclark is working on the remote support

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -37,7 +37,11 @@ def local_postgresql?
...
+  end
+
+  def pulpcore_enabled?
+    param_value('foreman_proxy_plugin_pulp', 'pulpcore_enabled')
```
ekohl, line 48: Is this correct? I think the class is actually `foreman_proxy::plugin::pulp` as it is in the current code.

sjha4: So I tried this:
mod = Kafo::KafoConfigure.config.modules.find { |m| m.name == "foreman_proxy_plugin_pulp" }
puts mod.params.flatten
and I see all the params correctly. With this foreman_proxy::plugin::pulp I see nil.

ekohl: Sounds like we need tests for these helpers because that's not what I expected.

ekohl:
There is now a merge conflict as well.

katello/hooks/pre/10-reset_feature.rb
sjha4: Updated with the remote db part and running reset_pulpcore only if pulpcore_enabled? is true.

sjha4:
This should be ready now..nighlty spins up with staging repos and the remote db changes are in. Updated the PR with the remote pulpcore db changes addressed.

ekohl (approved):
Untested, but the code looks correct.

ehelms (approved):

PR #460, opened 2020-02-03
Fixes #28915: Always run foreman upgrade rake tasks in post hooks

ehelms:
These upgrade rake tasks are designed to determine if they need to be run, so there is no need for us to gate them on `--upgrade`.


katello/hooks/pre/30-upgrade.rb
```diff
@@ -1,50 +1,15 @@
...
-def step_path(step)
-  File.join(STEP_DIRECTORY, step.to_s)
-end
-
 if app_value(:upgrade)
   log_and_say :info, 'Upgrading, to monitor the progress on all related services, please do:'
   log_and_say :info, '  foreman-tail | tee upgrade-$(date +%Y-%m-%d-%H%M).log'
   sleep 3
 
-  upgrade_step :stop_services, :run_always => true
```
ekohl: Does anything rely on actually stopped services or is this safe to remove? Feels unrelated to the subject of this PR.

ehelms: This is a bit more of cleaning up upgrade hooks. But nothing in this file relies on stopped services so this ought to be dropped.

ekohl (approved):

katello/hooks/pre/30-upgrade.rb
ekohl: The thing is that this is before the installer runs and puppet doesn't guarantee services restart. While I don't mind the change, I generally feel more comfortable if it's a separate commit to keep things isolated.

PR #461, opened 2020-02-05
Refs #28697 - Use systemctl instead of foreman-maintain

ekohl:
The benefit of this is that systemctl is quiet where foreman-maintain is very noisy. It also avoids the overhead of Ruby just to start a service that's already known.

### DIFF
--- a/hooks/pre/30-el7_upgrade_postgresql.rb
+++ b/hooks/pre/30-el7_upgrade_postgresql.rb
@@ -1,5 +1,5 @@
 def postgresql_12_upgrade
-  execute('foreman-maintain service start --only=postgresql')
+  execute('systemctl start postgresql')
   (_name, _owner, _enconding, collate, ctype, _privileges) = `runuser postgres -c 'psql -lt | grep -E "^\s+postgres"'`.chomp.split('|').map(&:strip)
   execute('foreman-maintain service stop')
   ensure_package('rh-postgresql12-postgresql-server', 'installed')


ehelms (changes requested):
I disagree with this change for the following reasons:

 1) This upgrade only happens one time, so I'm not worried about verbosity or speed of stopping a service
 2) This provides confusing code as there is no consistency of what tooling is used to manage services
 3) If the problem is with the foreman-maintain verbosity or implementation, let's fix it there because foreman-maintain isn't going away

ehelms:
@ekohl thoughts here?

ekohl:
I still would like to push this. It gets rid of the following output in forklift:
```
Running Start Services
    ================================================================================
    Check if command is run as root user:                                 [32m[1m[OK][0m
    --------------------------------------------------------------------------------
    Start applicable services: Starting the following service(s):
    postgresql
                                                                                    |                                                                                 /                                                                                 /                                                                                 -                                                                                 \                                                                                 |                                                                                 /                                                                                 -                                                                                 \                                                                                 |                                                                                 / starting postgresql                                                                                / starting postgresql                                                                                - starting postgresql                                                                                \ starting postgresql                                                                                | starting postgresql                                                                                / starting postgresql                                                                                - starting postgresql                                                                                \ starting postgresql                                                                                | starting postgresql                                                                                / starting postgresql                                                                                - starting postgresql                                                                                \ starting postgresql                                                                                | All services started                                                [32m[1m[OK][0m
    --------------------------------------------------------------------------------
    foreman-maintain service start --only=postgresql finished successfully!
    Running Stop Services
```

> If the problem is with the foreman-maintain verbosity or implementation, let's fix it there because foreman-maintain isn't going away

Personally I think foreman-maintain hasn't really arrived yet. Since there's no Debian support, we can't generally rely on it anyway. I also see it as generally foreman-maintain calls foreman-installer so we shouldn't go the other way in the dependency graph. That's just asking for problems.

FYI: https://projects.theforeman.org/issues/25630

ehelms:
I think that is a more general question worth raising but I disagree with this change here. Two lines later foreman-maintain is used so this is not avoiding it just mudding the waters. I believe I have an action item to expand this conversation around foreman-maintain elevation and would like to approach that.

ehelms:
Given the abstract layer now for foreman_maintain, can we close this in favor of that abstraction layer for now?

ekohl:
Sure, but I still want to get rid of foreman-maintain in the installer if possible. After dropping the mongo migration that will be easier.

PR #462, opened 2020-02-05
Refs #28928 - Remove checkpoint_segments from tuning

ekohl:
The checkpoint_segments setting no longer exists since PostgreSQL 10 and the server fails to start up if it's present.

This is not a complete fix since it doesn't clean it up on existing installations which can break upgrades.

### DIFF
--- a/config/foreman.hiera/tuning/sizes/extra-extra-large.yaml
+++ b/config/foreman.hiera/tuning/sizes/extra-extra-large.yaml
@@ -12,6 +12,5 @@ postgresql::server::config_entries:
   max_connections: 1000
   shared_buffers: 32GB
   work_mem: 8MB
-  checkpoint_segments: 32
   effective_cache_size: 32GB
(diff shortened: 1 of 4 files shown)


mmoll (approved):

ehelms:
Unless the postgresql module supports cleaning up configs then I agree there isn't much to do here other than release notes.

ekohl:
The module supports it on the `postgresql::server::config_entry` type, but we use it via hiera:
https://github.com/puppetlabs/puppetlabs-postgresql/blob/2c85ed36c3a10f4f9a6c651abbb3940c8c6e1cf0/manifests/server.pp#L183-L187

That doesn't allow setting `ensure => absent`. Not entirely sure how to deal with this.

ehelms:
Me either but this is currently not a deviation from how we handle other tuning parameters with respect to PostgreSQL.

PR #468, opened 2020-02-13
Update pulpcore to theforeman org

ekohl:
The module was released to theforeman rather than katello. When pinning modules this breaks.


chris1984 (approved):
ACK

tbrisker (approved):
Thanks @ekohl !

PR #470, opened 2020-02-18
Fixes #29052: Drop upgrade flag and treat all installer runs the same

ehelms:
I broke this change out from the one targeted at post install and expanded it to include fully dropping the upgrade flag. This would treat all installer runs the exact same going forward. The only consequence of this currently is that Candlepin and Pulp 2 migrations will be executed on each run of the installer. I don't forsee this as a huge problem or additional amount of time consumed by the installer.

Relevant to this will be: https://github.com/theforeman/puppet-pulp/pull/390

I will also need to look into orchestrating pulp services stopping before running migration.


katello/hooks/pre/30-upgrade.rb
```diff
@@ -1,57 +1,8 @@
...
-end
-
-if app_value(:upgrade)
-  log_and_say :info, 'Upgrading, to monitor the progress on all related services, please do:'
-  log_and_say :info, '  foreman-tail | tee upgrade-$(date +%Y-%m-%d-%H%M).log'
-  sleep 3
-
-  upgrade_step :stop_services, :run_always => true
-
-  if File.exist?(PULP2_MIGRATION_MARKER_FILE)
```
ekohl, line 48: Was removal of this intentional?

ekohl: Apologies, didn't see the related puppet-pulp PR.

ehelms: Yea, see comment on PR. I am intending this to follow the change that makes puppet-pulp check if pending migrations to determine if its migrates.

katello/hooks/pre/30-candlpine_migrations.rb
```diff
@@ -0,0 +1,8 @@
+require 'fileutils'
```
ekohl: Nice filename :)

katello/hooks/boot/30-upgrade.rb
```diff
@@ -1,12 +0,0 @@
-app_option(
```
ekohl, line 1: Rather than removing, should we keep it and error out with a message that it's no longer needed for a release or 2?

katello/hooks/pre/30-candlpine_migrations.rb
```diff
@@ -0,0 +1,8 @@
...
+
+# puppet-candlepin users this file to know whether it needs to run cpdb --update
```
ekohl:
```suggestion
# puppet-candlepin uses this file to know whether it needs to run cpdb --update
```

ekohl (commented):
I still like the `candlepine` migrations file ;) Did you see my comments about keeping `--upgrade` but displaying an error to the user if it's used?

ehelms:
I did away with the ole candlpine, sounded like a candlepin scent. The --upgrade flag is back, but will error to point users away from its use.

katello/hooks/boot/30-upgrade.rb
```diff
@@ -4,9 +4,3 @@
   "Run the steps necessary for an upgrade such as migrations, rake tasks, etc.",
```
ekohl: This line can use an update to state it's just there to warn users it's no longer needed.

ekohl (approved):

ekohl:
I'll let Travis finish. This also needs some doc updates, but I'm glad we can finally kill it.

PR #471, opened 2020-02-20
Fixes #28903: Stop installation if a package cannot be installed

ehelms:


hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -20,7 +20,11 @@ def ensure_package(package, state = 'installed')
     end
 
     bin_path = Kafo::PuppetCommand.search_puppet_path('puppet')
-    execute("#{bin_path} resource package #{package} ensure=#{state}")
+    response = `#{bin_path} resource package #{package} ensure=#{state}`
+
+    if state == 'installed' && response.include?('purged')
```
ekohl, line 25: Puppet has an exit code of 0 if nothing changed and 2 when there was a change. It would be better to catch that and state it couldn't install the package.

ekohl (changes requested):

ehelms:
That implies on subsequent runs, puppet will exit 0 and installer report
the package could not be installed which would be false.

ekohl:
I meant that if you catch all codes other than 0 and 2, you should halt the installer execution.

ehelms:
I'm missing something fundamental here. Let me show you my output:

```
[root@pulp vagrant]# puppet resource package rh-postgresql12 ensure='installed'
Error: Execution of '/bin/yum -d 0 -e 0 -y install rh-postgresql12' returned 1: Error: Nothing to do
Error: /Package[rh-postgresql12]/ensure: change from 'purged' to 'present' failed: Execution of '/bin/yum -d 0 -e 0 -y install rh-postgresql12' returned 1: Error: Nothing to do
package { 'rh-postgresql12':
  ensure   => 'purged',
  provider => 'yum',
}
[root@pulp vagrant]# $?
bash: 0: command not found
```

ekohl:
Interested, I guess the `resource` command behaves different from `puppet apply`.

ekohl:
In https://github.com/theforeman/foreman-installer/pull/474 I submitted a draft alternative. I'd like to hear what you think about that approach while I actually test it.

PR #472, opened 2020-02-20
Fixes #29097: Move puppet-agent-oauth reinstall to Foreman hooks

ehelms:


ekohl (changes requested):
I actually like this hook because it provides a nice guarantee that the gem actually works in that env. We also still support upgrading from Puppet 5 to Puppet 6 where this is needed. IMHO this should be moved to the general hooks.

ehelms:
Moved to main hooks and stopped it form outputting the check for the package.

ekohl (approved):
Code wise I think this is ok, but please have a look at the commit message and PR summary. I think generalizing it to all scenarios is more important than the output handling.

ehelms:
Updated commit and PR message

ekohl (approved):
I think this safely falls back on Debian since the `rpm` command will simply fail. Ideally we'd also have a check that uses `dpkg` and if so, executes the right `apt` command but let's consider that out of scope for now.

PR #473, opened 2020-02-21
Refs #28915 - Only run upgrade if the installer ran

ekohl:
If the user passed in --noop, no upgrade should take place. When the install failed, there's also no point in running the upgrades.

### DIFF
--- a/hooks/post/30-upgrade.rb
+++ b/hooks/post/30-upgrade.rb
@@ -1,3 +1,3 @@
-if foreman_server?
+if !app_value(:noop) && [0, 2].include?(@kafo.exit_code) && foreman_server?
   execute('foreman-rake upgrade:run')
 end


ekohl:
This needs a cherry pick to 2.0-stable (cc @tbrisker)

hooks/post/30-upgrade.rb
```diff
@@ -1,3 +1,3 @@
-if foreman_server?
+if !app_value(:noop) && [0, 2].include?(@kafo.exit_code) && foreman_server?
```
evgeni, line 1: `[0, 2].include?(@kafo.exit_code)` is such a crude way to say "did kafo finish successfully?", but this is not a new patern, so you get a pass ;)

evgeni (approved):

hooks/post/30-upgrade.rb
ehelms: We do use that `[0, 2]` check in a few places, could we move it to the hook extensions as a function?

ekohl: How about we merge this PR now and I'll update kafo to provide that + update develop to use it.

ehelms: Deal, here's the Redmine -- https://projects.theforeman.org/issues/29112

ehelms (approved):

ekohl:
> This needs a cherry pick to 2.0-stable (cc @tbrisker)

Disregard this, that part was done post-branch.

PR #474, opened 2020-02-21
Fixes #28903 - Stop installer when package install fails

ekohl:
The puppet resource command always exits with code 0, even when the package failed to install. This approach uses puppet apply with detailed exit codes to detect when it failed. It also unifies all PUP-2169 workarounds in a single file and execution.

This is a draft because it's untested.

### DIFF
--- a/hooks/boot/01-kafo-hook-extensions.rb
+++ b/hooks/boot/01-kafo-hook-extensions.rb
@@ -1,4 +1,5 @@
 require 'English'
+require 'open3'
 
 module HookContextExtension
   # FIXME: remove when #23332 is released
@@ -14,13 +15,25 @@ module HookContextExtension
     !File.exist?(success_file)
   end
 
-  def ensure_package(package, state = 'installed')
-    unless ['installed', 'absent', 'latest'].include?(state)
-      fail_and_exit('Incorrect package state supplied to ensure_package')
-    end
+  def ensure_packages(packages, state = 'installed')
+    return if packages.empty?
 
+    code = "package { ['#{packages.join('\', \'')}']: ensure => #{state} }"
+    logger.info("Ensuring #{packages.join(', ')} to package state #{state}")
+    stdout, stderr, status = apply_puppet_code(code)
+
+    unless [0, 2].include?(status.exitstatus)
+      log_and_say(:error, "Failed to ensure #{packages.join(', ')} #{(packages.length == 1) ? 'is' : 'are'} #{state}")
+      log_and_say(:error, stderr.strip) if stderr && stderr.strip
+      logger.debug(stdout.strip) if stdout && stdout.strip
+      logger.debug("Exit status is #{status.exitstatus.inspect}")
+      exit(1)
+    end
+  end
+
+  def apply_puppet_code(code)
     bin_path = Kafo::PuppetCommand.search_puppet_path('puppet')
-    execute("#{bin_path} resource package #{package} ensure=#{state}")
+    Open3.capture3("echo \"#{code}\" | #{bin_path} apply --detailed-exitcodes")
   end
 
   def fail_and_exit(message)
--- a/hooks/pre/30-el7_upgrade_postgresql.rb
+++ b/hooks/pre/30-el7_upgrade_postgresql.rb
@@ -2,17 +2,17 @@ def postgresql_12_upgrade
   execute('foreman-maintain service start --only=postgresql')
   (_name, _owner, _enconding, collate, ctype, _privileges) = `runuser postgres -c 'psql -lt | grep -E "^\s+postgres"'`.chomp.split('|').map(&:strip)
   execute('foreman-maintain service stop')
-  ensure_package('rh-postgresql12-postgresql-server', 'installed')
 
+  server_packages = ['rh-postgresql12-postgresql-server']
   if execute_command("rpm -q postgresql-contrib", false, false)
-    ensure_package('rh-postgresql12-postgresql-contrib', 'installed')
+    server_packages << 'rh-postgresql12-postgresql-contrib'
   end
+  ensure_packages(server_packages, 'installed')
 
   execute(%(scl enable rh-postgresql12 "PGSETUP_INITDB_OPTIONS='--lc-collate=#{collate} --lc-ctype=#{ctype} --locale=#{collate}' postgresql-setup --upgrade"))
-  ensure_package('postgresql-server', 'absent')
-  ensure_package('postgresql', 'absent')
+  ensure_packages(['postgresql', 'postgresql-server'], 'absent')
   execute('rm -f /etc/systemd/system/postgresql.service')
-  ensure_package('rh-postgresql12-syspaths', 'installed')
+  ensure_packages(['rh-postgresql12-syspaths'], 'installed')
 end
 
 if local_postgresql? && el7? && foreman_server? && needs_postgresql_scl_upgrade?
--- a/hooks/pre/30-install_redis_scl.rb
+++ b/hooks/pre/30-install_redis_scl.rb
@@ -1,4 +1,0 @@
-# Working around https://tickets.puppetlabs.com/browse/PUP-2169
-if el7? && local_redis?
-  ensure_package('rh-redis5-redis', 'installed')
-end
--- a/hooks/pre/31-el7_install_postgresql.rb
+++ b/hooks/pre/31-el7_install_postgresql.rb
@@ -1,4 +1,0 @@
-# Working around https://tickets.puppetlabs.com/browse/PUP-2169
-if local_postgresql? && el7?
-  ensure_package('rh-postgresql12-postgresql-server', 'installed')
-end
--- a/hooks/pre/31_install_scl_packages.rb
+++ b/hooks/pre/31_install_scl_packages.rb
@@ -1,0 +1,7 @@
+# Working around https://tickets.puppetlabs.com/browse/PUP-2169
+if el7?
+  packages = []
+  packages << 'rh-postgresql12-postgresql-server' if local_postgresql?
+  packages << 'rh-redis5-redis' if local_redis?
+  ensure_packages(packages, 'installed')
+end
--- a/spec/hook_context_spec.rb
+++ b/spec/hook_context_spec.rb
@@ -1,0 +1,145 @@
+require 'spec_helper'
+require 'kafo/hook_context'
+require_relative '../hooks/boot/01-kafo-hook-extensions'
+
+describe 'HookContextExtension' do
+  let(:kafo) { instance_double('KafoConfigure') }
+  let(:logger) { instance_double('Logger') }
+  let(:context) { Kafo::HookContext.new(kafo) }
+
+  before do
+    allow(context).to receive(:logger).and_return(logger)
+  end
+
+  context 'ensure_packages' do
+    let(:state) { 'installed' }
+    subject { context.ensure_packages(packages, state) }
+
+    context 'no packages' do
+      let(:packages) { [] }
+
+      it 'returns without performing any action' do
+        is_expected.to be_nil
+      end
+    end
+
+    context 'with packages' do
+      let(:stdout) { '  StdOut    ' }
+      let(:stderr) { ' StdErr  ' }
+      let(:status) { instance_double('Process::Status', exitstatus: exitstatus) }
+      let(:exitstatus) { nil }
+
+      before do
+        allow(logger).to receive(:info)
+        allow(context).to receive(:apply_puppet_code).and_return([stdout, stderr, status])
+      end
+
+      context 'single package' do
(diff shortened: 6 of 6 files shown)


ehelms:
I'll wait to see the outcome of using apply as a mechanism for this. As for this change, I would prefer that if this apply solves the problem, that we use it for the `ensure_package` function for a few reasons.

 1) I think this won't be an isolated issue for just these packages, any package that we want to ensure is installed should be handled this way
 2) I like the individual hooks as it makes each use case explicit

ekohl:
> I like the individual hooks as it makes each use case explicit

On the other hand, the user only has the Puppet initialization time once instead of twice which means a faster intaller.

I've tested a bit with this:
```
[root@dc31dcbacea9 foreman-installer]# foreman-installer --no-enable-foreman-proxy --no-enable-puppet
Ensuring rh-postgresql12-postgresql-server, rh-redis5-redis to package state installed
Failed to ensure rh-postgresql12-postgresql-server, rh-redis5-redis are installed
Notice: Compiled catalog for dc31dcbacea9.example.com in environment production in 0.48 seconds
Error: Execution of '/usr/bin/yum -d 0 -e 0 -y install rh-postgresql12-postgresql-server' returned 1: Error: Nothing to do
Error: /Stage[main]/Main/Package[rh-postgresql12-postgresql-server]/ensure: change from 'purged' to 'present' failed: Execution of '/usr/bin/yum -d 0 -e 0 -y install rh-postgresql12-postgresql-server' returned 1: Error: Nothing to do
Error: Execution of '/usr/bin/yum -d 0 -e 0 -y install rh-redis5-redis' returned 1: Error: Nothing to do
Error: /Stage[main]/Main/Package[rh-redis5-redis]/ensure: change from 'purged' to 'present' failed: Execution of '/usr/bin/yum -d 0 -e 0 -y install rh-redis5-redis' returned 1: Error: Nothing to do
Notice: Applied catalog in 0.91 seconds
[root@dc31dcbacea9 foreman-installer]# yum -y install centos-release-scl-rh
Loaded plugins: fastestmirror, ovl
Loading mirror speeds from cached hostfile
 * base: mirror.nl.datapacket.com
 * epel: ftp.nluug.nl
 * extras: mirror.ipserv.nl
 * updates: centos.mirror.transip.nl
Resolving Dependencies
--> Running transaction check
---> Package centos-release-scl-rh.noarch 0:2-3.el7.centos will be installed
--> Finished Dependency Resolution

Dependencies Resolved

========================================================================================================================================
 Package                                  Arch                      Version                             Repository                 Size
========================================================================================================================================
Installing:
 centos-release-scl-rh                    noarch                    2-3.el7.centos                      extras                     12 k

Transaction Summary
========================================================================================================================================
Install  1 Package

Total download size: 12 k
Installed size: 20 k
Downloading packages:
centos-release-scl-rh-2-3.el7.centos.noarch.rpm                                                                  |  12 kB  00:00:00     
Running transaction check
Running transaction test
Transaction test succeeded
Running transaction
  Installing : centos-release-scl-rh-2-3.el7.centos.noarch                                                                          1/1 
  Verifying  : centos-release-scl-rh-2-3.el7.centos.noarch                                                                          1/1 

Installed:
  centos-release-scl-rh.noarch 0:2-3.el7.centos                                                                                         

Complete!
[root@dc31dcbacea9 foreman-installer]# foreman-installer --no-enable-foreman-proxy --no-enable-puppet
Ensuring rh-postgresql12-postgresql-server, rh-redis5-redis to package state installed
.... # regular output
```

I'd like to hide the log_and_say output if it's debug level (unless the user specified `--verbose`) but I'll submit that as a separate PR.

ekohl:
```
[root@2a40cc5bf344 ~]# foreman-installer 
Ensuring rh-postgresql12-postgresql-server, rh-redis5-redis to package state installed
Failed to ensure rh-postgresql12-postgresql-server, rh-redis5-redis are installed
Notice: Compiled catalog for 2a40cc5bf344.example.com in environment production in 0.50 seconds
Error: Execution of '/usr/bin/yum -d 0 -e 0 -y install rh-postgresql12-postgresql-server' returned 1: Error: Nothing to do
Error: /Stage[main]/Main/Package[rh-postgresql12-postgresql-server]/ensure: change from 'purged' to 'present' failed: Execution of '/usr/bin/yum -d 0 -e 0 -y install rh-postgresql12-postgresql-server' returned 1: Error: Nothing to do
Error: Execution of '/usr/bin/yum -d 0 -e 0 -y install rh-redis5-redis' returned 1: Error: Nothing to do
Error: /Stage[main]/Main/Package[rh-redis5-redis]/ensure: change from 'purged' to 'present' failed: Execution of '/usr/bin/yum -d 0 -e 0 -y install rh-redis5-redis' returned 1: Error: Nothing to do
Notice: Applied catalog in 0.96 seconds
[root@2a40cc5bf344 ~]# yum -y install centos-release-scl-rh
# yum install output
[root@2a40cc5bf344 ~]# foreman-installer 
Ensuring rh-postgresql12-postgresql-server, rh-redis5-redis to package state installed
# .... regular output
```

ehelms:
From my testing, we are talking the difference of 2 or 3 seconds which isn't much to me if it means a design that is easier to understand and a function that works for all use cases. We have 8 instances of `ensure_package` being called by my count. I'd prefer a single implementation that all use since its a common use case. I like this design approach though given the way it handles missing packages more cleanly.

ekohl:
> I'd prefer a single implementation that all use since its a common use case. I like this design approach though given the way it handles missing packages more cleanly.

I'm not entirely sure what you want me to do here. Do you want me to update the `ensure_package` implementation so that becomes reliable? I can also factor out this code into `ensure_packages` which takes an array and let `ensure_package` call that.

ehelms:
Yep, that's what I was thinking.

ekohl:
That makes sense and I agree. This initial implementation was just to prove it was a viable strategy.

ekohl:
Updated. There were some changes in the method so I still need to do a test run.

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -14,13 +15,20 @@ def new_install?
...
-      fail_and_exit('Incorrect package state supplied to ensure_package')
-    end
+  def ensure_packages(packages, state = 'installed')
+    return if packages.empty?
 
+    code = "package { ['#{packages.join('\', \'')}']: ensure => #{state} }"
     bin_path = Kafo::PuppetCommand.search_puppet_path('puppet')
-    execute("#{bin_path} resource package #{package} ensure=#{state}")
+
+    logger.debug("Ensuring #{packages.join(', ')} to package state #{state}")
```
ehelms: I think this is nice to surface as info rather than debug since its a critical step in execution of the installer.

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -14,13 +15,20 @@ def new_install?
...
+    stdout_and_stderr, status = Open3.capture2e("echo \"#{code}\" | #{bin_path} apply --detailed-exitcodes")
+
+    unless [0, 2].include?(status.exitstatus)
+      log_and_say(:error, "Failed to ensure #{packages.join(', ')} are #{state}")
+      log_and_say(:error, stdout_and_stderr)
```
ehelms: Testing this, I don't think this information provides much value:

```
[ INFO 2020-02-25T13:29:10 verbose] Notice: Compiled catalog for puma.war.example.com in environment production in 0.63 seconds
Notice: Applied catalog in 0.87 seconds
```

ehelms:
Changed work great overall, just a few requests around the output

hooks/boot/01-kafo-hook-extensions.rb
ekohl: That's weird. Do you have more context because it should only be showing Puppet output when it fails but this looks like it did nothing at all.

hooks/boot/01-kafo-hook-extensions.rb
ekohl: IMHO the katello-installer is too verbose. The installer should be quiet unless you pass --verbose and this should show up then.

hooks/boot/01-kafo-hook-extensions.rb
ehelms: That output is actually form me printing it on successful install from testing to see if `stdout_and_stderr` provided any valuable logging output. When it does actually fail this line prints `nothing`:

```
Failed to ensure rh-postgresql12-postgresql-server, rh-redis5-redis are installed
[ERROR 2020-02-25T13:29:10 verbose] Failed to ensure rh-postgresql12-postgresql-server, rh-redis5-redis are installed
[ INFO 2020-02-25T13:29:10 verbose] Installer finished in 11.248545533 seconds
```

hooks/boot/01-kafo-hook-extensions.rb
ehelms: Correct... but this to me is not debug information this is `info` level information that should show up in the installer as such. That is, if I set my log level to `info`, I should see this output in my opinion. 

Related is that all of our scenarios have their log level set to `debug` which feels wrong as the default (https://github.com/theforeman/foreman-installer/blob/develop/config/foreman.yaml#L17). That's a separate issue to discuss.

ekohl: > Correct... but this to me is not debug information this is `info` level information that should show up in the installer as such. That is, if I set my log level to `info`, I should see this output in my opinion.

So essentially:
```suggestion
    logger.info("Ensuring #{packages.join(', ')} to package state #{state}")
```

That makes sense to me.



> Related is that all of our scenarios have their log level set to `debug` which feels wrong as the default (https://github.com/theforeman/foreman-installer/blob/develop/config/foreman.yaml#L17). That's a separate issue to discuss.

I think I agree with that, but indeed another discussion.

hooks/boot/01-kafo-hook-extensions.rb
ekohl: It feels like the status.exitstatus doesn't match `0` or `2`. Could you add a `logger.debug("Exit status is #{status.exitstatus.inspect}")` as well?

ehelms: After playing around a bit, I found the output from switching to:

```
stdout, stderr, status = Open3.capture3

log_and_say(:error, stderr.strip)
```

To be cleaner. The inclusion of stdout just adds this kinda of crufty output:

```
Notice: Compiled catalog for puma.war.example.com in environment production in 0.70 seconds
```

ekohl:
Updated the output handling.

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -14,13 +15,22 @@ def new_install?
...
-    end
+  def ensure_packages(packages, state = 'installed')
+    return if packages.empty?
 
+    code = "package { ['#{packages.join('\', \'')}']: ensure => #{state} }"
     bin_path = Kafo::PuppetCommand.search_puppet_path('puppet')
-    execute("#{bin_path} resource package #{package} ensure=#{state}")
+
+    logger.info("Ensuring #{packages.join(', ')} to package state #{state}")
+    stdout, stderr, stderr, status = Open3.capture3("echo \"#{code}\" | #{bin_path} apply --detailed-exitcodes")
```
ehelms: One too many stderr

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -14,13 +15,22 @@ def new_install?
...
+
+    unless [0, 2].include?(status.exitstatus)
+      log_and_say(:error, "Failed to ensure #{packages.join(', ')} are #{state}")
+      log_and_say(:error, stderr.strip) if stderr && stderr.strip
+      logger.info(stdout.strip) if stdout && stdout.strip
```
ehelms: I still think this is noise that we shouldn't print out. At best, this could be debug information.

ekohl:
Updated, now with tests.

ehelms (approved):

ehelms:
Thanks for the tests, they look like a good foundation!

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -14,13 +15,25 @@ def new_install?
...
+      logger.debug(stdout.strip) if stdout && stdout.strip
+      logger.debug("Exit status is #{status.exitstatus.inspect}")
+      exit(1)
     end
+  end
 
+  def apply_puppet_code(code)
     bin_path = Kafo::PuppetCommand.search_puppet_path('puppet')
-    execute("#{bin_path} resource package #{package} ensure=#{state}")
+    Open3.capture3("echo \"#{code}\" | #{bin_path} apply --detailed-exitcodes")
```
ekohl, line 36: Note that ideally we'd pass in code via stdin so we don't need to deal with escaping, but for now this code is good enough.

PR #477, opened 2020-02-26
Fixes #29184 - Add cdn-ssl param migration hook

chris1984:


katello/hooks/pre/32-cdn_setting.rb
```diff
@@ -0,0 +1,10 @@
+if module_enabled?('katello')
+    if param_value('katello', 'cdn_ssl_version')
+      param = param_value('katello', 'cdn_ssl_version')
+      logger.info 'cdn_ssl_version param found, migrating to a Katello setting'
+      `foreman-rake -- config -k cdn_ssl_version -v #{param}`
```
ekohl: In the pre hook I'm not sure you can rely on the setting being present in an upgrade scenario. The database might still need a migration and seeding. Is there are reason you're doing this in a pre job rather than doing it all in the post hook?

ekohl (commented):
Please have a look at the whitespace.

chris1984:
@ekohl updated to fix the rubocop issues. 

I did part of it in the pre hook since I noticed when doing everything in the post hook, we did not remove the parameter in the `/etc/foreman/plugins/katello.yaml` file. 

When I ran the installer again to confirm that the logic was working correctly, I noticed it removing the param from the `katello.yaml` file. I can do everything in the post hook but I think I would need to use `sed` to remove the entry from the `katello.yaml` file unless there is a better way?

ekohl:
> I did part of it in the prehook since I noticed when doing everything in the post hook, we did not remove the parameter in the `/etc/foreman/plugins/katello.yaml` file.

That's a good point and indeed needed. I was only thinking about the `foreman-rake config` part and forgot the rest.

katello/hooks/pre/32-cdn_setting.rb
```diff
@@ -0,0 +1,10 @@
+if module_enabled?('katello')
+  if param_value('katello', 'cdn_ssl_version')
```
ehelms: From all of my testing, this won't be available here if it's been dropped from the puppet module. You'll have to instead take a tactic of using a `pre_commit` hook and accessing `@kafo.config.answers` which will have the answers prior to being committed to disk and being overwritten.

chris1984: @ehelms if we drop the puppet code will the `/etc/foreman/settings.d/katello.yaml` get updated with the removal of the cdn_ssl_setting or will we have to manually remove it from that file directly? Since the parameter will be gone, we can't rely on the installer to remove it like we would normally.

ekohl: AFAIK Kafo doesn't drop unknown answers and just keeps them. If the puppet-katello module doesn't have the parameter, you don't need the whole pre hook, call foreman-rake in a post hook and then delete the answer.

chris1984: That is what I was planning on going with @ekohl but then I saw the comment from Eric. Thanks for clearing it up.

ehelms: My experimentation has shown that any parameters that are unknown to Kafo will not be propagated to the answers file when Kafo commits the parameters from memory to the file. Thus having to do this in `pre_commit` was how we solved it in another case.

ekohl: Interesting, so https://github.com/theforeman/foreman-installer/blob/develop/config/foreman.migrations/20190111180118_delete_removed_settings.rb was mostly redundant?

However, with pre_commit I wonder if that works. Can you actually create a setting using `foreman-rake config` or do you need to do it after running the migrations to ensure it's present (as I've been assuming)?

ehelms: foreman-rake config creates a setting in the Foreman database, I'm missing why that matters when its ran.

ekohl: https://github.com/theforeman/foreman/blob/26dda8dbbf32d357657c3d22018c9d247fb0cb4e/lib/tasks/config.rake#L102

It will hard fail if the database row with that settings name doesn't exist:
```console
# foreman-rake -- config -k doesnotexist -v foo
rake aborted!
NoMethodError: undefined method `settings_type' for nil:NilClass
/usr/share/foreman/lib/tasks/config.rake:105:in `block in run_key_values'
/usr/share/foreman/lib/tasks/config.rake:100:in `each'
/usr/share/foreman/lib/tasks/config.rake:100:in `run_key_values'
/usr/share/foreman/lib/tasks/config.rake:77:in `run'
/usr/share/foreman/lib/tasks/config.rake:167:in `block in <top (required)>'
/opt/rh/rh-ruby25/root/usr/share/gems/gems/rake-12.3.0/exe/rake:27:in `<top (required)>'
Tasks: TOP => config
(See full trace by running task with --trace)
```

Usually the database migrations take care of creating this, hence why it should be a post hook. Otherwise you're creating an upgrade issue.

ehelms: The data needs to be stored somewhere "temporarily" I think. Whether thats a file that gets cleaned up at the end of post section or on the scenario configuration. As we want to consider if the installer fails we don't want to have erased the data. So this could be the hook creating it or some more supported entity where old parameters are stored.

ekohl: That was my thinking as well. I had hoped we could use the answers file but looks like we should use the [store](https://github.com/theforeman/kafo#store) (`store_custom_config` and `get_custom_config`). I don't exactly know at which stage the `store_custom_config` method should be called so you'll need to experiment with that.

chris1984:
@ehelms @ekohl updating and working :) I tested with the puppet-katello pr and here is the results of my testing:

Pre Commit Hook:
```bash
[ INFO 2020-03-19T13:39:22 verbose] Executing hooks in group pre_commit
[ INFO 2020-03-19T13:39:28 verbose] cdn_ssl_version param found, storing for post hook
```

Removing it from `katello.yaml` file since puppet code is gone:
```bash
[ WARN 2020-03-19T13:41:37 verbose]  /Stage[main]/Katello::Application/File[/etc/foreman/plugins/katello.yaml]/content: 
[ WARN 2020-03-19T13:41:37 verbose] --- /etc/foreman/plugins/katello.yaml	2020-03-18 17:55:15.435217825 +0000
[ WARN 2020-03-19T13:41:37 verbose] +++ /tmp/puppet-file20200319-5570-ay4sq4	2020-03-19 13:41:37.581697239 +0000
[ WARN 2020-03-19T13:41:37 verbose] @@ -2,7 +2,6 @@
[ WARN 2020-03-19T13:41:37 verbose]  ## Module: puppet-katello
[ WARN 2020-03-19T13:41:37 verbose]  
[ WARN 2020-03-19T13:41:37 verbose]  :katello:
[ WARN 2020-03-19T13:41:37 verbose] -  :cdn_ssl_version: SSLv23
[ WARN 2020-03-19T13:41:37 verbose]    :rest_client_timeout: 3600
[ WARN 2020-03-19T13:41:37 verbose]  
[ WARN 2020-03-19T13:41:37 verbose]    :content_types:
[ INFO 2020-03-19T13:41:37 verbose]  Computing checksum on file /etc/foreman/plugins/katello.yaml
[ INFO 2020-03-19T13:41:37 verbose]  /Stage[main]/Katello::Application/File[/etc/foreman/plugins/katello.yaml]: Filebucketed /etc/foreman/plugins/katello.yaml to puppet with sum 1f74ed4f03f430a2574a4cca5f4e3feb
[ WARN 2020-03-19T13:41:37 verbose]  /Stage[main]/Katello::Application/File[/etc/foreman/plugins/katello.yaml]/content: content changed '{md5}1f74ed4f03f430a2574a4cca5f4e3feb' to '{md5}778fa03cb7efcc2f6e5752b558077bbb'
```

Post Hook Running:
```bash
[ INFO 2020-03-19T13:46:32 verbose] cdn_ssl_version param found, migrating to a Katello setting
[ INFO 2020-03-19T13:47:03 verbose] All hooks in group post finished
```

Screenshot showing it in the settings page:

https://nimbusweb.me/s/share/3985742/gy42ph4587lctbb1dw92

katello/hooks/post/31-cdn_setting.rb
```diff
@@ -0,0 +1,9 @@
+if module_enabled?('katello')
+  param = get_custom_config('cdn_ssl_version')
+  if param
+    logger.info 'cdn_ssl_version param found, migrating to a Katello setting'
+    `foreman-rake -- config -k cdn_ssl_version -v #{param}`
+  else
+    logger.debug 'cdn_ssl_version already migrated, skipping'
+  end
+end
```
ekohl: Please ensure there are newlines at the end

ekohl (changes requested):
You need to clear the value for `cdn_ssl_version` in the custom config to avoid calling the same rake task every time.

katello/hooks/post/31-cdn_setting.rb
```diff
@@ -0,0 +1,9 @@
+if module_enabled?('katello')
+  param = get_custom_config('cdn_ssl_version')
+  if param
+    logger.info 'cdn_ssl_version param found, migrating to a Katello setting'
+    `foreman-rake -- config -k cdn_ssl_version -v #{param}`
```
ehelms: I'd lean towards calling our builtin `execute` so that this gets logged and fails if this command errors out.

chris1984:
@ekohl how does this look, if it looks good ill test it again.

katello/hooks/post/31-cdn_setting.rb
```diff
@@ -0,0 +1,11 @@
...
+  param = get_custom_config('cdn_ssl_version')
+  if param
+    logger.info 'cdn_ssl_version param found, migrating to a Katello setting'
+    execute("foreman-rake -- config -k cdn_ssl_version -v #{param}")
+    store_custom_config('cdn_ssl_version', nil)
+  else
+    logger.debug 'cdn_ssl_version already migrated, skipping'
+  end
+end
+
```
ekohl: Note that this is an empty line, not a line ending.

katello/hooks/pre_commit/14-cdn_setting.rb
```diff
@@ -0,0 +1,10 @@
+if module_enabled?('katello')
+  param = @kafo.config.answers['katello']['cdn_ssl_version']
```
ekohl: Untested, but I wonder whether just `param_value('katello', 'cdn_ssl_version')` works.

chris1984: It does not, I was talking to @ehelms about this yesterday, when I added the puppet pr and installed puppet-strings it went away from the param so @ehelms told me to use this variable instead.

katello/hooks/post/31-cdn_setting.rb
```diff
@@ -0,0 +1,10 @@
+if module_enabled?('katello')
+  param = get_custom_config('cdn_ssl_version')
+  if param
+    logger.info 'cdn_ssl_version param found, migrating to a Katello setting'
+    execute("foreman-rake -- config -k cdn_ssl_version -v #{param}")
```
ekohl: It's a good habit to always quote variables
```suggestion
    execute("foreman-rake -- config -k cdn_ssl_version -v '#{param}'")
```

katello/hooks/pre_commit/14-cdn_setting.rb
ekohl: Would you mind adding that as a comment above it? This is something non-obvious and usually that's a good reason for a comment.

ekohl (commented):
Overall this looks correct though I didn't test it.

chris1984:
@ekohl updated and it works again with testing, only difference this time with using `execute` during the post hook we are being transparent to the user what we are doing which looks better:

```
[ INFO 2020-03-19T19:48:49 verbose] cdn_ssl_version param found, migrating to a Katello setting
cdn_ssl_version: SSLv23
foreman-rake -- config -k cdn_ssl_version -v 'SSLv23' finished successfully!
[ INFO 2020-03-19T19:49:15 verbose] All hooks in group post finished
```

katello/hooks/post/31-cdn_setting.rb
```diff
@@ -0,0 +1,10 @@
+if module_enabled?('katello')
```
ekohl: This should respect `noop`
```suggestion
if !app_value(:noop) && module_enabled?('katello')
```

katello/hooks/pre_commit/14-cdn_setting.rb
```diff
@@ -0,0 +1,12 @@
+if module_enabled?('katello')
+  # Using @kafo.config.answers because it will have the answers prior to being
+  # committed to disk and being overwritten
+  # https://github.com/theforeman/foreman-installer/pull/477#pullrequestreview-370785117
+  param = @kafo.config.answers['katello']['cdn_ssl_version']
```
ekohl: `param` is a method that exists in the hook context and it's not a good practice to overwrite this. I think `value` is a better name.

katello/hooks/pre_commit/14-cdn_setting.rb
```diff
@@ -0,0 +1,12 @@
+if module_enabled?('katello')
+  # Using @kafo.config.answers because it will have the answers prior to being
+  # committed to disk and being overwritten
+  # https://github.com/theforeman/foreman-installer/pull/477#pullrequestreview-370785117
```
ekohl: I think this is technically not entirely correct. `param` (and thus `param_value`) use the same data structure (`kafo.config.answers`). However, `param` goes through `Kafo::PuppetModule` which uses the actual Puppet module (via a cache in production) to see which params actually exist. Since it doesn't exist, `param('katello', 'cdn_ssl_version')` returns nil.

Would something like this be better?

```suggestion
  # Using @kafo.config.answers because param() goes through PuppetModule. At
  # this point the parameter doesn't exist in the module on disk so param()
  # returns nil.
```

ekohl (commented):
I'm debating whether it should be `pre_validations` rather than `pre_commit`. The values should already be present then and there's less code that could interfere with it (no validations nor the entire interactive mode). However, I have insufficient experience with hooks to know whether it can really be a problem. @ehelms any thoughts?

katello/hooks/pre_commit/14-cdn_setting.rb
ehelms: I find this confusing personally. But I find explaining all of it confusing to be honest. As there are nuances depending on where you are at in the process as to what each represents. At this stage:

 * kafo.config.answers represents the old parameters and values on disk
 * param represents the current state of params before the transaction commits the new values to disk

ekohl: You're right, but in the current committed comment I'm missing explicitly that `param()` doesn't work. The reason I ask for this is that it might be very tempting to think about refactoring it using built in hooks, forgetting that it doesn't work.

chris1984:
@ekohl @ehelms updated with the comment and requested changes.

ekohl (approved):
@ehelms any last comments?

ehelms (approved):

PR #479, opened 2020-03-02
Fixes #29243: Create new install file

ehelms:
Fixes regression where new install messages are being printed
all the time rather than only on new installs.

### DIFF
--- a/hooks/post/99-post_install_message.rb
+++ b/hooks/post/99-post_install_message.rb
@@ -24,4 +24,5 @@ else
   failure_message
 end
 
+File.write(success_file, '') if !app_value(:noop) && new_install?
 log_message(@kafo)


ekohl:
There's a merge conflict now

ehelms:
Thanks, fixed

ekohl (commented):
Thinking more about this: should it actually be stored in the scenario itself rather than a separate marker file? For reference, how version locking stores its config:

https://github.com/theforeman/foreman-installer/blob/481ee53f04cbcc98fea39bba1a9c2c388bf22f1c/katello/hooks/pre_commit/09-version_locking.rb#L3-L9

I also don't know how well `configure_application` in a post hook behaves so I don't mind merging this and migrating later, but wanted to get your opinion first

ehelms:
I think that is a good idea and I'd like to explore it when I have a bit more time so I filed https://projects.theforeman.org/issues/29262 to look into it. I'd like to go with this option for now as it restores prior behavior.

ekohl (approved):
Does this also need to be cherry picked into 2.0.0-rc3?

ehelms:
Yea, we should for cleaner install messages.

PR #482, opened 2020-03-05
Refs #29069 - Install pulpcore-selinux if needed

ekohl:
To work around PUP-2169, the selinux contexts need to be present before the main Puppet run. Because pulpcore-selinux depends on python3-pulpcore, the foreman-installer-katello package can't depend on it.

### DIFF
--- a/hooks/pre/31_install_scl_packages.rb
+++ b/hooks/pre/31_install_scl_packages.rb
@@ -3,5 +3,6 @@ if el7?
   packages = []
   packages << 'rh-postgresql12-postgresql-server' if local_postgresql?
   packages << 'rh-redis5-redis' if local_redis?
+  packages << 'pulpcore-selinux' if pulpcore_enabled?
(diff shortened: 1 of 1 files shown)


evgeni (approved):
You could argue `pulpcore-selinux` is not an SCL package, so the filename becomes "wrong", but I don't care too much.

Strictly speaking, we could also drop the `pulpcore-selinux` → `python3-pulpcore` dependency, as the policy really doesn't apply to any files inside that package, it only labels content folders, that are not created by installing `python3-pulpcore` at all.

That all said, the given patch solves the issue at hand, so go forth and merge.

ekohl:
> You could argue `pulpcore-selinux` is not an SCL package, so the filename becomes "wrong", but I don't care too much.

Fair point. Perhaps it should be `31_pup_2169_workarounds.rb`

ehelms (approved):

PR #483, opened 2020-03-05
Refs #28903 - Swap the postgresql{,-server} removal

ekohl:
postgresql can't be removed until postgresql-server is removed. 72d7c3f67fd4e1b131871c946d03808c9d26a124 swapped the order. This swaps the order around.

### DIFF
--- a/hooks/pre/30-el7_upgrade_postgresql.rb
+++ b/hooks/pre/30-el7_upgrade_postgresql.rb
@@ -10,7 +10,8 @@ def postgresql_12_upgrade
   ensure_packages(server_packages, 'installed')
 
   execute(%(scl enable rh-postgresql12 "PGSETUP_INITDB_OPTIONS='--lc-collate=#{collate} --lc-ctype=#{ctype} --locale=#{collate}' postgresql-setup --upgrade"))
-  ensure_packages(['postgresql', 'postgresql-server'], 'absent')
+  ensure_packages(['postgresql-server'], 'absent')
+  ensure_packages(['postgresql'], 'absent')
   execute('rm -f /etc/systemd/system/postgresql.service')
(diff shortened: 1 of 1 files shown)


ekohl:
I should not I haven't tested this before, just wanted to share this before I headed out.

ehelms:
Doesn't work :/ Puppet keeps ordering them itself so they need to be split up.

ekohl:
Pushed an updated version that I'm now testing.

hooks/pre/30-el7_upgrade_postgresql.rb
```diff
@@ -10,9 +10,25 @@ def postgresql_12_upgrade
...
+    ensure => absent,
+  } ->
+  package { 'rh-postgresql12-syspaths':
+    ensure => installed,
+  }
+  EOF
+  failure_message = "Failed to ensure rh-postgresql12-syspaths is the system PostgreSQL"
+
+  logger.info("Ensuring rh-postgresql12-syspaths is the system PostgreSQL")
+  apply_puppet_code(puppet_code, failure_message)
```
ehelms: I get the concept and power of this code, the interface as a developer is just now much uglier :/ This is also a larger re-factoring of the change compared to:

```
 ensure_packages(['postgresql-server'], 'absent')
 ensure_packages(['postgresql'], 'absent')
```

I'll be curious to see how testing goes, but my vote is still the above.

ekohl: I considered the same. Given it needs to be cherry picked, I'm tempted to agree with you.

ekohl:
Ready to review now.

ehelms (approved):

tbrisker:
2.0 - f08b784

PR #484, opened 2020-03-06
Fixes #29282: Add migration to enable Puma

sthirugn:


config/foreman.migrations/20200305172758_enable_puma.rb
```diff
@@ -0,0 +1 @@
+answers['foreman']['passenger'] = false if answers['foreman'].is_a?(Hash)
```
ekohl: If you add `&& answers['foreman']['passenger']` at the end it makes the tests happy
```suggestion
answers['foreman']['passenger'] = false if answers['foreman'].is_a?(Hash) && answers['foreman']['passenger']
```

sthirugn:
The test steps I used to test this PR successfully in my local environment:

1. Spin up katello server
2. Check the value of `passenger` in answers file
```sh
# cat /etc/foreman-installer/scenarios.d/katello-answers.yaml | grep 'passenger:'
   passenger: false
```
3. Go to ` /usr/share/foreman-installer/config/katello.migrations/` and create the migration file from this PR
4. Check `.applied` file in the above folder to make sure that the newly added migration file is not present.
4. run the installer: `foreman-installer --scenario katello --migrations-only`
5. Check the value of `passenger` in answers file
```sh
# cat /etc/foreman-installer/scenarios.d/katello-answers.yaml | grep 'passenger:'
   passenger: true
```
6. Check `.applied` file to make sure that the newly added migration file is present. To retest manually remove the new migration entry from this `.applied` file.

config/foreman.migrations/20200305172758_enable_puma.rb
sthirugn: Thank you for the suggestion

ekohl (approved):

sthirugn:
To create new migration files, I used the `create-migration` script like:
```sh
 /foreman-installer/bin/create-migration enable_puma /foreman-installer/config/foreman.migrations/
 /foreman-installer/bin/create-migration enable_puma /foreman-installer/config/katello.migrations/
```

PR #485, opened 2020-03-09
2.0.0-rc3 cherry picks

ekohl:
For the SELinux part I also need to do a puppet-pulpcore module release to correctly label the ports.


tbrisker (approved):
Thanks @ekohl !

PR #486, opened 2020-03-09
Refs #29052: Allow --upgrade to pass through

ehelms:


katello/hooks/pre_validations/29-upgrade.rb
```diff
@@ -1,3 +1,3 @@
 if app_value(:upgrade)
-  fail_and_exit 'The --upgrade option has been removed. The installer will ensure the system is in the right state. The use of --upgrade is no longer needed.'
+  log_and_say(:warning, 'The --upgrade option has been removed. The installer will ensure the system is in the right state. The use of --upgrade is no longer needed.')
```
ekohl, line 2: I don't mind making it non-fatal, but IMHO there's no reason to log this.

tbrisker (approved):
makes sense to me not to fail on it for at least a couple of releases

katello/hooks/pre_validations/29-upgrade.rb
ehelms: I find its standard practice to log deprecation messages as a way to raise awareness to the user to modify their scripts.

ekohl: Fair enough. My only concern is that we'll forget to remove it in Foreman 3.0. Does a TODO in the code make sense?

ehelms: I created https://projects.theforeman.org/issues/29296 and once there is a Foreman version in Redmine I can target it at it.

PR #488, opened 2020-03-12
Fixes #29339 - Show tuning option only for katello

sthirugn:


katello/hooks/boot/13-hiera.rb
```diff
@@ -3,9 +3,11 @@
 TUNING_SIZES = ['default'] + ['medium', 'large', 'extra-large', 'extra-extra-large']
 TUNING_FACT = 'tuning'.freeze
 
-app_option(
-  '--tuning',
-  'INSTALLATION_SIZE',
-  "Tune for an installation size. Choices: #{TUNING_SIZES.join(', ')}",
-  :default => get_custom_fact(TUNING_FACT)
-)
+if module_enabled?('katello')
```
ekohl: I doubt `module_enabled?` actually works at the boot phase. https://github.com/theforeman/kafo#hooks suggests it shouldn't.

An alternative approach would be to see if `get_custom_fact` has a value. Then in migrations we can make sure a value is set in the katello scenario. Technically migrations aren't executed before this, but in the RPM there's a script that runs the migrations so for the end user it'd work.

ekohl (commented):
Some more context would be good to have. With only for katello, do you mean

katello/hooks/boot/13-hiera.rb
ehelms: If I read the code correctly, the boot hook runs after the scenario and config have been both loaded and migrated which means kafo has in memory the migrated config and answers but prior to any parameters. The only parameters that could affect this would be `--no-enable` which we don't really support anyway. I tested this and encouraged @sthirugn to go this route. I'm running another test to double check my initial results and will report back.

https://github.com/theforeman/kafo/blob/master/lib/kafo/kafo_configure.rb#L89

ehelms:
```
[root@installer vagrant]# foreman-installer --tuning
Insufficient memory for tuning size
Tuning profile 'default' requires at least 8 GB of memory and 1 CPU cores
[root@installer vagrant]# vi /usr/share/foreman-installer/katello/hooks/boot/13-hiera.rb 
[root@installer vagrant]# cat /usr/share/foreman-installer/katello/hooks/boot/13-hiera.rb 
# TODO: automatically get the tuning sizes - standard doesn't exist and only
# loads base. The rest maps to config/foreman.hiera/tuning/sizes/$size.yaml
TUNING_SIZES = ['default'] + ['medium', 'large', 'extra-large', 'extra-extra-large']
TUNING_FACT = 'tuning'.freeze

if module_enabled?('blah')
app_option(
  '--tuning',
  'INSTALLATION_SIZE',
  "Tune for an installation size. Choices: #{TUNING_SIZES.join(', ')}",
  :default => get_custom_fact(TUNING_FACT)
)
end
[root@installer vagrant]# foreman-installer --tuning
ERROR: Unrecognised option '--tuning'

See: 'foreman-installer --help'
[root@installer vagrant]# vi /usr/share/foreman-installer/katello/hooks/boot/13-hiera.rb 
[root@installer vagrant]# cat /usr/share/foreman-installer/katello/hooks/boot/13-hiera.rb 
# TODO: automatically get the tuning sizes - standard doesn't exist and only
# loads base. The rest maps to config/foreman.hiera/tuning/sizes/$size.yaml
TUNING_SIZES = ['default'] + ['medium', 'large', 'extra-large', 'extra-extra-large']
TUNING_FACT = 'tuning'.freeze

if module_enabled?('katello')
app_option(
  '--tuning',
  'INSTALLATION_SIZE',
  "Tune for an installation size. Choices: #{TUNING_SIZES.join(', ')}",
  :default => get_custom_fact(TUNING_FACT)
)
end
[root@installer vagrant]# foreman-installer --tuning
Insufficient memory for tuning size
Tuning profile 'default' requires at least 8 GB of memory and 1 CPU cores
[root@installer vagrant]# 
```

ekohl: > If I read the code correctly, the boot hook runs after the scenario and config have been both loaded and migrated which means kafo has in memory the migrated config and answers but prior to any parameters.

That would be good. I only read the documentation and that gave me the impression it might not work.

> The only parameters that could affect this would be `--no-enable` which we don't really support anyway.

The alternative is `unless kafo.module('katello').nil?` which only checks if the module is present, regardless of the exact answer.

It should be rather easy to verify because you can run the installer locally and `--help` should work. You can also add a line line:
https://github.com/theforeman/foreman-installer/blob/20a1c2d28b798cf8cb67e8c861149cc56047fc58/.travis.yml#L29
Though we already have the output in Travis because we have the generic `--help`. I only see `--tuning` for Katello so I think that confirms this works. We have more CI than we think ;)

Can't we check for the `foreman` module instead of `katello`? That's probably more accurate since our tuning profiles are more to Foreman/Apache/PostgreSQL.

ehelms:
On Thu, Mar 12, 2020, 5:31 PM Ewoud Kohl van Wijngaarden <
notifications@github.com> wrote:

> *@ekohl* commented on this pull request.
> ------------------------------
>
> In katello/hooks/boot/13-hiera.rb
> <https://github.com/theforeman/foreman-installer/pull/488#discussion_r391907575>
> :
>
> > @@ -3,9 +3,11 @@
>  TUNING_SIZES = ['default'] + ['medium', 'large', 'extra-large', 'extra-extra-large']
>  TUNING_FACT = 'tuning'.freeze
>
> -app_option(
> -  '--tuning',
> -  'INSTALLATION_SIZE',
> -  "Tune for an installation size. Choices: #{TUNING_SIZES.join(', ')}",
> -  :default => get_custom_fact(TUNING_FACT)
> -)
> +if module_enabled?('katello')
>
> If I read the code correctly, the boot hook runs after the scenario and
> config have been both loaded and migrated which means kafo has in memory
> the migrated config and answers but prior to any parameters.
>
> That would be good. I only read the documentation and that gave me the
> impression it might not work.
>
> The only parameters that could affect this would be --no-enable which we
> don't really support anyway.
>
> The alternative is unless kafo.module('katello').nil? which only checks
> if the module is present, regardless of the exact answer.
>
> It should be rather easy to verify because you can run the installer
> locally and --help should work. You can also add a line line:
>
> https://github.com/theforeman/foreman-installer/blob/20a1c2d28b798cf8cb67e8c861149cc56047fc58/.travis.yml#L29
> Though we already have the output in Travis because we have the generic
> --help. I only see --tuning for Katello so I think that confirms this
> works. We have more CI than we think ;)
>
> Can't we check for the foreman module instead of katello? That's probably
> more accurate since our tuning profiles are more to
> Foreman/Apache/PostgreSQL.
>
Would be more future proof, right now this hook is only available to
katello scenario.

ekohl:
> Would be more future proof, right now this hook is only available to katello scenario.

That was also my thinking. Checking `foreman` gives the same end result but later it'll just be a move of files.

If you actually run it, it'll probably fail because this assumes there is some tuning.
https://github.com/theforeman/foreman-installer/blob/20a1c2d28b798cf8cb67e8c861149cc56047fc58/katello/hooks/pre_commit/13-hiera.rb#L16-L23

Also note that that part is now the system checks for CPU and memory. Just disabling that allows installing on too small systems.

katello/hooks/boot/13-hiera.rb
sthirugn: Before submitting the PR, I tested the code locally in both 6.7 satellite and capsule servers with the changes from this PR:

```console
# hostname
sat-6-7-qa-rhel7.apex.example.com

# satellite-installer --scenario satellite --help | grep 'tuning'
    --tuning INSTALLATION_SIZE    Tune for an installation size. Choices: default, medium, large, extra-large, extra-extra-large (default: "medium")
```
```console
# hostname
capsule-6-7-qa-rhel7.apex.example.com

# satellite-installer --scenario capsule --help | grep 'tuning'
<No Output>
```

sthirugn:
ok I tested with `foreman` and I can confirm it gives me the same result as in https://github.com/theforeman/foreman-installer/pull/488#discussion_r391998300. So do you all recommend that I proceed with the following code?
```bash
# cat /usr/share/foreman-installer/katello/hooks/boot/13-hiera.rb 
# TODO automatically get the tuning sizes - standard doesn't exist and only
# loads base. The rest maps to config/foreman.hiera/tuning/sizes/$size.yaml
TUNING_SIZES = ['default'] + ['medium', 'large', 'extra-large', 'extra-extra-large']
TUNING_FACT = 'tuning'

if module_enabled?('foreman')
  app_option(
    '--tuning',
    'INSTALLATION_SIZE',
    "Tune for an installation size. Choices: #{TUNING_SIZES.join(', ')}",
    :default => get_custom_fact(TUNING_FACT)
  )
end
```

ekohl:
Pretty much, though you can probably also add a line at the start `return unless module_enabled?('foreman')`. Please do have a look at `pre_commit` hook. It relies on a tuning size and may break on fresh installations but I haven't thought through all permutations.

sthirugn:
updated the check to `foreman` and confirmed local satellite/capsule tests worked fine. I will test fresh install now and report back.

sthirugn:
I tested with new install of katello and it worked great.

When I attempted testing with new install of foreman-proxy-content, it fails:
```console
# foreman-installer --scenario foreman-proxy-content --disable-system-checks
package puppet-agent-oauth is not installed
Invalid tuning profile
'' is not one of default, medium, large, extra-large, extra-extra-large
```

sthirugn:
@ewoud Added check in pre_commit/13-hiera.rb too to overcome proxy related failures.

ehelms:
Looks good -- small rubocop issue

katello/hooks/pre_commit/13-hiera.rb
```diff
@@ -1,48 +1,51 @@
-# A mapping of tuning profile to the required CPU cores and memory in GB
-TUNING_SIZES = {
-  'default' => { cpu_cores: 1, memory: 8 },
-  'medium' => { cpu_cores: 8, memory: 32 },
-  'large' => { cpu_cores: 16, memory: 64 },
-  'extra-large' => { cpu_cores: 32, memory: 128 },
-  'extra-extra-large' => { cpu_cores: 48, memory: 256 },
-}.freeze
-TUNING_FACT = 'tuning'.freeze
+if module_enabled?('foreman')
```
ekohl: This means we no longer perform any system checks on the other scenarios, meaning you can install on systems with 4GB RAM.

sthirugn: Thank you for the comment. I am working on this today to see how best to fix this.

sthirugn:
Doing a few more rounds of testing, but I think this latest update will take care of default testing for all scenarios as usual.

sthirugn:
@ekohl ready for review again. Now the proxy would catch installer run which doesn't have minimum requirements (8GB, 1cpu):

```console
# foreman-installer --scenario foreman-proxy-content
Insufficient memory for tuning size
Tuning profile 'default' requires at least 8 GB of memory and 1 CPU cores
```

ehelms:
Unfortunately, this now leaks the tuning concept which you are working to hide into the `foreman-proxy-content` scenario. I thought the whole point of the tuning option right now is that its only applicable to the server installation. And that we were not prescribing minimums for the proxy/content proxy case.

sthirugn:
The default system requirements for all scenarios (8GB, 1 CPU) is now embedded in the code for the tuning in general. So all we are trying to do is to hide `tuning` option in `foreman-proxy-content` scenario but the minimum requirements for `foreman-proxy-content` will still be checked. See also https://github.com/theforeman/foreman-installer/pull/488#discussion_r392505102 for the related comment by @ekohl 

To summarize the experience:
**katello**
```console
# foreman-installer --scenario katello
Insufficient memory for tuning size
Tuning profile 'default' requires at least 8 GB of memory and 1 CPU cores

# foreman-installer --scenario katello --tuning medium --disable-system-checks
...
...
Preparing installation Done                                              
  Success!
...
```

**foreman-proxy-content**
```console
# foreman-installer --scenario foreman-proxy-content
Insufficient memory for tuning size
Tuning profile 'default' requires at least 8 GB of memory and 1 CPU cores

# foreman-installer --scenario foreman-proxy-content --tuning medium
ERROR: Unrecognised option '--tuning'

See: 'foreman-installer --help'
```
But I see why it is confusing - the error message of `foreman-proxy-content` mentioning `Tuning`. May be I can look into fixing the error message for `foreman-proxy-content` scenario above to make it sound like a non-tuning error.

sthirugn:
I updated the PR now with a new error message to distinguish non-tuning system check error and the new output will look like this:

**katello**
```console
# foreman-installer --scenario katello
Insufficient memory for tuning size
Tuning profile 'medium' requires at least 32 GB of memory and 8 CPU cores
```

**foreman-proxy-content**
```console
# foreman-installer --scenario foreman-proxy-content
This scenario requires at least 8 GB of memory and 1 CPU cores
```

katello/hooks/pre_commit/13-hiera.rb
```diff
@@ -29,14 +33,22 @@
   # Check if it's actually 90% of the required. If a crash kernel is enabled
   # then the reported total memory is lower than in reality.
   if facts[:memory][:system][:total_bytes] < (required_memory * 1024 * 1024 * 1024 * 0.9)
-    say "<%= color('Insufficient memory for tuning size', :bad) %>"
-    say "Tuning profile '#{new_tuning}' requires at least #{required_memory} GB of memory and #{required_cores} CPU cores"
+    if module_enabled?('foreman')
```
ehelms: Things look good. I think this could be more a little more robust against the future by checking if `app_value(:tuning)`

sthirugn: I will test with this to see how it works.

sthirugn: updated now thanks

sthirugn:
Thank you for the excellent feedback so far. The code is now updated. Let me know if there is anything else.

ehelms (approved):

katello/hooks/pre_commit/13-hiera.rb
```diff
@@ -29,14 +33,22 @@
   # Check if it's actually 90% of the required. If a crash kernel is enabled
   # then the reported total memory is lower than in reality.
   if facts[:memory][:system][:total_bytes] < (required_memory * 1024 * 1024 * 1024 * 0.9)
-    say "<%= color('Insufficient memory for tuning size', :bad) %>"
-    say "Tuning profile '#{new_tuning}' requires at least #{required_memory} GB of memory and #{required_cores} CPU cores"
+    if app_value(:tuning)
+      say "<%= color('Insufficient memory for tuning size', :bad) %>"
+      say "Tuning profile '#{new_tuning}' requires at least #{required_memory} GB of memory and #{required_cores} CPU cores"
+    else
+      say "This scenario requires at least #{required_memory} GB of memory and #{required_cores} CPU cores"
```
ekohl: There's [scenario_id](https://github.com/theforeman/kafo/blob/06187fab9efb3ad2ef9708e5f2c9cea7501168c1/lib/kafo/hook_context.rb#L110-L113) as well if you want to use it. This is the same value as passed to `--scenario`, so `foreman`, `katello` or `foreman-proxy-content`.

ekohl (approved):

katello/hooks/pre_commit/13-hiera.rb
sthirugn: good idea, updated now.

sthirugn: the new output looks like:
```console
# foreman-installer --scenario foreman-proxy-content
puppet-agent-oauth-0.5.1-3.el7.noarch
This scenario 'foreman-proxy-content' requires at least 8 GB of memory and 1 CPU cores
```

sthirugn:
@ekohl final comment updated as well, thanks.

ekohl:
Now that I think a bit more about this, do we need to deal with upgrading users? I forgot that we introduced this in 1.24 so we may already have users who set their tuning to non-default but lose the ability to change it after this. Do we need a migration to reset it to default in the foreman-proxy-content scenario?

sthirugn:
> Now that I think a bit more about this, do we need to deal with upgrading users? I forgot that we introduced this in 1.24 so we may already have users who set their tuning to non-default but lose the ability to change it after this. Do we need a migration to reset it to default in the foreman-proxy-content scenario?

Good point, I will look into this.

sthirugn:
> > Now that I think a bit more about this, do we need to deal with upgrading users? I forgot that we introduced this in 1.24 so we may already have users who set their tuning to non-default but lose the ability to change it after this. Do we need a migration to reset it to default in the foreman-proxy-content scenario?
> 
> Good point, I will look into this.

Just for the records, this is fixed in https://github.com/theforeman/foreman-installer/pull/491

PR #489, opened 2020-03-13
Fixes #28799: Allow calling puppet from inside SCL context

ehelms:
There are a series of changes needed to run the installer within an SCL Ruby context which would allow us to drop older Ruby support all together. This stems from a need to unset the environment and break out of SCL jail to run commands from the Puppet Ruby environment. The set of changes is listed here. There is of course packaging changes required as well that I will include once I have a PR up to show what is required to make this change.

 * https://github.com/theforeman/kafo/pull/250
 * https://github.com/theforeman/kafo_parsers/pull/37

The relevant EL7 packaging changes that would go with it:

 * https://github.com/theforeman/foreman-packaging/pull/4858

### DIFF
--- a/hooks/boot/01-kafo-hook-extensions.rb
+++ b/hooks/boot/01-kafo-hook-extensions.rb
@@ -33,7 +33,7 @@ module HookContextExtension
 
   def apply_puppet_code(code)
     bin_path = Kafo::PuppetCommand.search_puppet_path('puppet')
-    Open3.capture3("echo \"#{code}\" | #{bin_path} apply --detailed-exitcodes")
+    Open3.capture3(*Kafo::PuppetCommand.format_command("echo \"#{code}\" | #{bin_path} apply --detailed-exitcodes"))
   end
 
   def fail_and_exit(message, code = 1)
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'kafo', '~> 4.0'
+gem 'kafo', '~> 4.1'
 gem 'librarian-puppet'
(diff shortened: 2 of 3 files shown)


hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -33,7 +33,7 @@ def ensure_packages(packages, state = 'installed')
 
   def apply_puppet_code(code)
     bin_path = Kafo::PuppetCommand.search_puppet_path('puppet')
-    Open3.capture3("echo \"#{code}\" | #{bin_path} apply --detailed-exitcodes")
+    Open3.capture3(PuppetCommand.format_command("echo \"#{code}\" | #{bin_path} apply --detailed-exitcodes"))
```
ekohl: I've been wondering if we should provide this entire method as a higher level abstraction from Kafo. Thoughts?

ehelms: I thought about it. Right now we, the installer, use this in two places: kafo and kafo_parsers. Kafo requires kafo_parsers so I contemplated putting it all in kafo_parsers (that just felt weird). This only needs to live for the lifespan of EL7 which has only a few Foreman releases left. And really, I see packaging the installer for SCL as a way to get the installer to drop Ruby 2.0 conventions and align El7 and EL8 Ruby versions going forward. So I am inclined towards the least amount of overhead to capture this functionality. Which I why I arrived at putting it in kafo_parsers where its needed and Kafo and then making sure the installer uses Kafo (since it already relies heavily on it).

ekohl (changes requested):
Please require Kafo 4.1:
https://github.com/theforeman/foreman-installer/blob/a3a0fe6dc707ac295eeb6858792f0dda08179ff2/Gemfile#L3

ehelms:
Gemfile is updated

ekohl (approved):
Let's not forget to do the same in packaging :)

PR #491, opened 2020-03-23
Refs #29339 - Reset tuning for foreman-proxy-content scenario

sthirugn:


config/foreman-proxy-content.migrations/200323151921-reset_tuning.rb
```diff
@@ -0,0 +1,2 @@
+scenario[:facts] = {} unless scenario.key?(:facts)
```
ekohl: A common short hand for this in Ruby is
```suggestion
scenario[:facts] ||= {}
```

However, due to https://github.com/theforeman/foreman-installer/blob/develop/config/foreman-proxy-content.migrations/190904144224-set_tuning_fact.rb it's also safe to leave it out.

```suggestion
```

ekohl (commented):
What I normally like to do is explain in the commit message why we're resetting this. For example, add to it:
> 6048fcc67c656025888e226f9e88c74feddeadd2 removed the option to set tuning for foreman-proxy-content. This ensures the tuning is the same for every installation.

In this case I'd also say this PR should be `Refs #29339` rather than fixes since we keep the convention that there's only one `Fixes`.

config/foreman-proxy-content.migrations/200323151921-reset_tuning.rb
sthirugn: I wanted to keep this specific migration safe and not dependent on the previous tuning migration. If for some reason, an advanced user hand edits this configuration, it is probably safe to check and initialize this to { } before acting on it.

ekohl: I'm not opposed to defensive coding and in fact think too often we're not defensive enough so I'm fine with it here.

sthirugn:
@ekohl thank you for the feedback. I updated them all now.

ekohl (approved):
Thanks!

PR #492, opened 2020-03-27
Fixes #29425 - Configure default mongo cache size

sthirugn:
Configuring default mongo cache size to 20% of total memory. See[1]
Otherwise mongo goes onto consume about 50% of total memory. See[2]

[1] https://access.redhat.com/solutions/4505561
[2] https://docs.mongodb.com/manual/core/wiredtiger/#memory-use

### DIFF
--- a/config/foreman.hiera/tuning/common.yaml
+++ b/config/foreman.hiera/tuning/common.yaml
@@ -22,3 +22,6 @@ postgresql::server::config_entries:
   log_line_prefix: '%t '
   log_min_duration_statement: 1000
   log_rotation_size: 200000
+
+mongodb::server::config_data:
+  storage.wiredTiger.engineConfig.cacheSizeGB: "%{facts.kafo.scenario.custom.mongo_cache_size}"
--- a/katello/hooks/pre_commit/13-hiera.rb
+++ b/katello/hooks/pre_commit/13-hiera.rb
@@ -7,10 +7,22 @@ TUNING_SIZES = {
   'extra-extra-large' => { cpu_cores: 48, memory: 256 },
 }.freeze
 TUNING_FACT = 'tuning'.freeze
+MONGO_CACHE_SIZE_FACT = 'mongo_cache_size'.freeze
 
 EXIT_INVALID_TUNING = 101
 EXIT_INSUFFICIENT_CPU_CORES = 102
 EXIT_INSUFFICIENT_MEMORY = 103
+
+# Setting Default Mongo WiredTiger Engine cache size to be 20% of total memory
+# See https://access.redhat.com/solutions/4505561
+# See https://docs.mongodb.com/manual/core/wiredtiger/#memory-use
+current_mongo_cache_size = get_custom_fact(MONGO_CACHE_SIZE_FACT)
+required_mongo_cache_size = (facts[:memory][:system][:total_bytes] * 0.2 / 1024 / 1024 / 1024).round(2)
+if current_mongo_cache_size != required_mongo_cache_size
+  store_custom_fact(MONGO_CACHE_SIZE_FACT, required_mongo_cache_size)
+  # Store the app config to disk
+  kafo.config.configure_application
+end
 
 current_tuning = get_custom_fact(TUNING_FACT)
 if module_enabled?('foreman')


sthirugn:
Local test passed, test procedure:

1. `vagrant up centos7-katello-3.15`
2. Apply the changes from this PR
3. `foreman-installer --scenario katello`
4. Review `journalctl -f`:
 ```console
Mar 27 18:01:33 centos7-katello-3-15.apex.example.com mongod.27017[8441]: [initandlisten] options: { command: [ "run" ], config: "/etc/opt/rh/rh-mongodb34/mongod.conf", net: { bindIp: "127.0.0.1" }, processManagement: { fork: true, pidFilePath: "/var/opt/rh/rh-mongodb34/run/mongodb/mongod.pid" }, security: { authorization: "disabled" }, storage: { dbPath: "/var/lib/mongodb", journal: { enabled: true }, wiredTiger: { engineConfig: { cacheSizeGB: 1.13 } } }, systemLog: { destination: "syslog", quiet: true } }
```
5. Note cacheSizeGB above - cacheSizeGB: 1.13
6. check total memory in system facts
```console
# facter -p memory | grep total_bytes
    total_bytes => 2147479552,
    total_bytes => 6087458816,  <-- this is the system memory info
```
7. 20% of this 6087458816 × (20÷100) ÷ 1024 ÷1024 ÷1024 = 1.133877563 ~ 1.13 matches the above mongo configuration from Step 4.

config/foreman-proxy-content.yaml
```diff
@@ -14,6 +14,7 @@
 :hiera_config: ./config/foreman-hiera.yaml
 :facts:
   tuning: 'default'
+  mongo_cache_size: 1
```
ekohl: Technically this isn't really needed. `get_custom_fact` should return `nil`. `nil != something` so that should store the fact. This is also the upgrade scenario now since there's no migration. The reason we set one for tuning is that there the value is read for the `--tuning` parameter and we need to show something.

ekohl (commented):
Overall it looks good. I've always been wondering about doing this, but then in the postgresql department. We may even be able to drop some of the explicit tuning profile options then.

config/foreman-proxy-content.yaml
sthirugn: good to know this, I was uncomfortable putting a random default value `1` here.

ekohl (approved):

katello/hooks/pre_commit/13-hiera.rb
```diff
@@ -7,11 +7,23 @@
...
 
+# Setting Default Mongo WiredTiger Engine cache size to be 20% of total memory
+# See https://access.redhat.com/solutions/4505561
+# See https://docs.mongodb.com/manual/core/wiredtiger/#memory-use
+current_mongo_cache_size = get_custom_fact(MONGO_CACHE_SIZE_FACT)
+required_mongo_cache_size = (facts[:memory][:system][:total_bytes] * 0.2 / 1024 / 1024 / 1024).round(2)
+if current_mongo_cache_size != required_mongo_cache_size
+  store_custom_fact(MONGO_CACHE_SIZE_FACT, required_mongo_cache_size)
+  # Store the app config to disk
+  kafo.config.configure_application
```
ekohl, line 24: Now that I think about this, this is something we do wrong in the other one as well I think but it should respect noop
```suggestion
  kafo.config.configure_application unless app_value(:noop)
```

ekohl: Ok, it's actually time for weekend. Kafo itself already handles that and there's no need to worry about it here.

sthirugn: ok great thank you for the info.

config/foreman.hiera/tuning/common.yaml
```diff
@@ -22,3 +22,6 @@ postgresql::server::config_entries:
   log_line_prefix: '%t '
   log_min_duration_statement: 1000
   log_rotation_size: 200000
+
+mongodb::server::config_data:
+  storage.wiredTiger.engineConfig.cacheSizeGB: "%{facts.kafo.scenario.custom.mongo_cache_size}"
```
ehelms, line 27: If a user hasn't upgraded to wiredTiger will this configuration option existing be a no-op for them? That is, will it break if not on wiredTiger?

ehelms:
Note, tests have not run yet. Travis is being flaky.

config/foreman.hiera/tuning/common.yaml
sthirugn: it shouldn't. Let me see if I can get an older version of server to test this. It may be tricky to get one though.

sthirugn: I can't get older versions of upstream / downstream boxes to get non-wiredTiger, is there a way to do it properly?

sthirugn: I found an old Satellite 6.3.5 server running mmapv1 and added the following to the custom-hiera.yaml and ran the installer. The mongo service came back up without any issues.
```
mongodb::server::config_data:
  storage.wiredTiger.engineConfig.cacheSizeGB: 2
```

PR #493, opened 2020-03-30
Fixes #29437 - Drop PuppetDB module

ekohl:
The intent was to make it possible to install PuppetDB but there’s no documentation, it’s tricky to set up and we don’t verify it (it may actually be broken). Adding to that, its maintainers are largely unresponsive. Adding support for new distribution versions and other module versions is very painful.

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -2,7 +2,6 @@ forge 'https://forgeapi.puppetlabs.com/'
 
 # Dependencies
 mod 'puppetlabs/postgresql',         '>= 5.6.0'
-mod 'puppetlabs/puppetdb'
 mod 'theforeman/dhcp',               :git => 'https://github.com/theforeman/puppet-dhcp'
 mod 'theforeman/dns',                :git => 'https://github.com/theforeman/puppet-dns'
 mod 'theforeman/git',                :git => 'https://github.com/theforeman/puppet-git'


tbrisker (approved):
I'm assuming we want this in 2.0-stable as well?

ekohl:
Indeed

mmoll:
Shouldn't we remove the parts in question from puppet-puppet first?

ekohl:
Initially I thought it should still be supported, but the installer shouldn't set it up. Looking at the code in puppet-puppet I'm now thinking about this. I at least started with https://github.com/theforeman/puppet-puppet/pull/732. The upsides are that it cleans things up properly, the downside is a major version bump in the last RC. However, it's breaking because the API changes, not because the changes are large. The installer doesn't really care about this API and only makes it more stable. Let me know what you think.

mmoll:
Technically I'm OK with either way. I don't think many people are using this feature, but it should be prominently indicated that it's removed or broken, I think we didn't announce anything for 2.0 regarding that, yet...

tbrisker:
Would dropping the module in the installer cause issues for puppet-puppet? If not, i'd suggest merging this and cleaning up puppet-puppet for 2.1. If there is risk this will cause the installer to break, i'd suggest postponing this change for 2.1 completely.

ekohl:
Without the linked puppet-puppet PR the `--puppet-server-puppetdb-*` options will cause a compilation error that the `puppetdb::config::master` class isn't found. As long as these options are not used (or technically only `-host`) then it'll be fine. The impact is that users can no longer fully configure the puppetserver to use puppetdb integration, even if the server is remote.

tbrisker:
Would it be possible that these options exist in the answers file and need proper clean up? i'm starting to lean more towards pushing this to 2.1 and not risking breaking something in 2.0 at this stage.

sthirugn:
+1 to push this after 2.0

ekohl:
So 2.1 is coming up and the removal from puppet-puppet was just merged. Thoughts on merging this now?

sthirugn:
+1 to merge

tbrisker (approved):
Please add it to the upgrade warnings, and be prepared to revert if this breaks in some unexpected fun way. 
Thanks @ekohl !

PR #495, opened 2020-04-01
Readd "Fixes #29282: Add migration to enable Puma"

sthirugn:
The commit bd02fbd reverted commit 0b244f5.

https://github.com/theforeman/puppet-foreman/pull/814 fixed the race
condition of systemctl start foreman returning immediately before the
service was up.

### DIFF
--- a/config/foreman.migrations/20200305172758_enable_puma.rb
+++ b/config/foreman.migrations/20200305172758_enable_puma.rb
@@ -1,0 +1,1 @@
+answers['foreman']['passenger'] = false if answers['foreman'].is_a?(Hash) && answers['foreman']['passenger']
--- a/config/katello.migrations/200306192827-enable_puma.rb
+++ b/config/katello.migrations/200306192827-enable_puma.rb
@@ -1,0 +1,1 @@
+answers['foreman']['passenger'] = false if answers['foreman'].is_a?(Hash) && answers['foreman']['passenger']


config/foreman.migrations/20200305172758_enable_puma.rb
```diff
@@ -0,0 +1,5 @@
+unless answers['foreman']['ipa_authentication']
+    unless answers['foreman']['keycloak']
+      answers['foreman']['passenger'] = false if answers['foreman'].is_a?(Hash) && answers['foreman']['passenger']
+    end
+end
```
ekohl: At this point it's not guaranteed that `answers['foreman']` is a hash and might result in `nil`. Then you have `nil['ipa_authentication']` which results in an error.

You could write it like:
```suggestion
if answers['foreman'].is_a?(Hash)
  mod_answers = answers['foreman']
  if mod_answers['passenger'] && !(mod_answers['ipa_authentication'] || mod_answers['keycloak'])
    answers['foreman']['passenger'] = false
  end
end
```

ekohl (commented):
Please start the line again with `Fixes` so our tooling recognizes it. To avoid confusion you can change it to `Fixes #29282 - Enable Puma on compatible upgrades` or similar.

config/foreman.migrations/20200305172758_enable_puma.rb
sthirugn: good point about `answers['foreman']`

sthirugn: @ekohl updated now, thanks

config/foreman.migrations/20200305172758_enable_puma.rb
```diff
@@ -0,0 +1,6 @@
+if answers['foreman'].is_a?(Hash)
+    mod_answers = answers['foreman']
+    if mod_answers['passenger'] && !(mod_answers['ipa_authentication'] || mod_answers['keycloak'])
```
ehelms: Could you explain why we are caring about ipa_authentication and keycloak here?

ekohl: So my thought was to capture https://github.com/theforeman/puppet-foreman/blob/1e10f6c43ba7ff8353b19b6b7d1a3dcf0b5310cd/manifests/init.pp#L351-L360 but you may be right that it's not needed. I was confusing Apache and Passenger.

ehelms: I did not test every permutation, but from my investigation and attempted testing I couldn't see why external auth wouldn't work with puma setup.

sthirugn: So should we remove the external auth check and enable migration for all? and see how things to?

ekohl: Yes, I was confused with a non-apache install. That is still a problem but one we don't see that often.

sthirugn:
Based on the feedback, reverted to a simple migration for all scenarios.

ehelms (approved):

ekohl (approved):

PR #497, opened 2020-04-09
Fixes #29512: Call foreman-maintain without SCL context

ehelms:


hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -36,6 +36,19 @@ def apply_puppet_code(code)
     Open3.capture3(*Kafo::PuppetCommand.format_command("echo \"#{code}\" | #{bin_path} apply --detailed-exitcodes"))
   end
 
+  def foreman_maintain(command)
```
ekohl: I don't want to rely on foreman_maintain, as I've made very clear. This makes it way too easy to use it. I think at most we should expose a `stop_all_services(exclude = [])` method. IMHO `start_service` should directly invoke `systemctl`. We don't have any more bulk service start code. We can also expose locking explicitly. That does mean a helper is useful so I'm wondering if we can make it private. AFAIK hooks are executed as an instance eval so they have access to private methods as well, which is slightly ugly.

ehelms: I'm sorry but I do not agree that code by obscurity is the way to prevent further use of foreman-maintain. Relying on it sparingly does not change the fact that its being used and needs to be used at the present moment. We have to tackle the root of the issue which is the overall design in my opinion. And I am working on that.

ekohl: I think that exposing the subcommands rather than a generic command is at least a better design since it abstracts away the exact tool and states the intention. Even if we don't hide `foreman_maintain` now, I think that `stop_all_services(exclude = [])` and `start_service` are good methods to implement.

ehelms: So I started writing this and I realized that I am writing more code that is effectively wrapping the subcommand in a "generic" command that looks just like the subcommand. So all the benefits of it looking and feeling like a generic command to me is lost. I can still go this route but it feels like a lot of overkill when I can just call foreman_maintain rather than hide that I cam calling it. Looking ahead we can drop some of this hook code that calls out to maintain and re-think how we handle these kind of "upgrade" tasks.

ehelms: @ekohl thoughts on my latest response?

ekohl: I still think they make sense because they make it easier to get rid of foreman_maintain.

katello/hooks/pre_commit/09-version_locking.rb
```diff
@@ -9,17 +9,13 @@
 end
 
 if lock_versions
-  unless system('command -v foreman-maintain > /dev/null')
-    fail_and_exit('Locking of package versions was requested but foreman-maintain is not installed')
-  end
-  unless system('foreman-maintain packages -h > /dev/null 2>&1')
+  unless foreman_maintain('packages -h')
     fail_and_exit('Locking of package versions was requested but foreman-maintain version installed does not support it')
```
ekohl: This doesn't work because `foreman_maintain()` will call `fail_and_exit` already.

katello/hooks/pre_commit/09-version_locking.rb
```diff
@@ -9,17 +9,13 @@
...
   end
 end
 
 # unlock packages if locked
-if system('foreman-maintain packages is-locked --assumeyes > /dev/null 2>&1')
+if foreman_maintain('packages is-locked --assumeyes')
```
ekohl: This will call `fail_and_exit` instead of gracefully ignoring the result.

pcreech:
From what I can see, this appears to be a sane enough abstraction of what is currently there, and is a good step in the right direction.   +1 from me

hooks/boot/03-foreman-maintain-extensions.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def stop_services(services)
+    command = "service stop"
+    command = "service stop --only=#{services.join(',')}" if services
+
+    foreman_maintain(command, true)
+  end
+
+  def foreman_maintain(command, allow_fail = false)
```
ekohl: Allow fail feels like an odd naming. My initial understanding of it was that it allowed to continue but the implementation is quite the opposite. How about `exit_on_fail`?

hooks/pre/30-el7_upgrade_postgresql.rb
```diff
@@ -1,7 +1,8 @@
 def postgresql_12_upgrade
-  execute('foreman-maintain service start --only=postgresql')
+  foreman_maintain('service start --only=postgresql')
```
ekohl: Old code?
```suggestion
```

ekohl (changes requested):
This is indeed the direction I was thinking about.

katello/hooks/pre/10-reset_feature.rb
```diff
@@ -5,6 +5,7 @@ def reset
   if foreman_installed?
 
     stop_services
+    start_services(['postgresq'])
```
ekohl:
```suggestion
    start_services(['postgresql'])
```

ekohl (approved):
There's one issue (see inline), other than that :+1: 

Unrelated question, but seeing the changes there makes me wonder: when do we drop the mongodb upgrade? 2.2?

ehelms:
Aye, I was thinking we drop the mongodb upgrad ebits after branch for 2.1

PR #498, opened 2020-04-15
Fixes #28352: Enable tuning config for foreman scenario

ehelms:


ekohl (commented):
Looks like the tests are failing.

ehelms:
Err, this is why it fails locally for me. @ekohl any idea? I've not encountered this before.

```
[ERROR 2020-04-15T10:01:19 main] connect: No such file or directory
Please make sure that the zfs-fuse daemon is running.
internal error: failed to initialize ZFS library
Error: Evaluation Error: Operator '[]' is not applicable to an Undef Value. (file: /home/ehelms/workspace/upstream/foreman-installer/tmp/share/foreman-installer/modules/puppet/manifests/params.pp, line: 39, column: 19) on node war.eagle
```

ekohl:
That looks like a facter issue. We probably have Facter 2 in our tests and started to use modern facts in our modules. It stems from something like `$facts['networking']['fqdn']` instead of `$facts['fqdn']` which Facter 2 doesn't support. Now that Facter 4 is out (which bundler can install), I think that should be a decent workaround. There's probably some gem pinned so it doesn't get considered by bundler.

@wbclark this is why I've been very hesitant in using modern facts

ehelms:
Passing tests now if we'd like this included in 2.1

wbclark (approved):
LGTM, and I support including this in 2.1 since imo it's a significant improvement to the user experience to implement these tuning profiles via the installer without having to add custom hiera.

ekohl (approved):
Untested, but :+1: for the general idea. Even though some parts (passenger, candlepin) aren't that important anymore, the postgresql tuning is useful.

It'd be good to add this to the manual and possibly a headline feature. You might be able to copy/move the parts from the Katello manual.

ehelms:
Opened here -- https://github.com/theforeman/theforeman.org/pull/1576

PR #500, opened 2020-04-15
Fixes #29541: Use Facter 4 in testing

ehelms:
Depends on https://github.com/voxpupuli/librarian/pull/10

I included the git path for now to test that this works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: false
 language: ruby
 addons:
   apt:
@@ -8,9 +7,7 @@ addons:
     - xsltproc
     - docbook-xsl
     - ca-certificates
-rvm: 2.4
-env:
-  - PUPPET_VERSION=5.0
+rvm: 2.5
 script:
   - bundle exec rake rubocop
   #
@@ -23,8 +20,11 @@ script:
   - INSTDIR=$(mktemp -d)
   - bundle exec rake build install PREFIX=$INSTDIR
   - bundle exec $INSTDIR/sbin/foreman-installer --help --scenario foreman
+  - cat $INSTDIR/var/log/foreman-installer/foreman.log
   - bundle exec $INSTDIR/sbin/foreman-installer --help --scenario foreman-proxy-content
+  - cat $INSTDIR/var/log/foreman-installer/foreman-proxy-content.log
   - bundle exec $INSTDIR/sbin/foreman-installer --help --scenario katello
+  - cat $INSTDIR/var/log/foreman-installer/katello.log
   - bundle exec $INSTDIR/sbin/foreman-proxy-certs-generate --help
   - bundle exec $INSTDIR/sbin/foreman-proxy-certs-generate --help |grep -q certs-update-server
   #
--- a/Gemfile
+++ b/Gemfile
@@ -3,13 +3,10 @@ source 'https://rubygems.org'
 gem 'kafo', '~> 4.1'
 gem 'librarian-puppet'
 gem 'puppet', ENV.key?('PUPPET_VERSION') ? "~> #{ENV['PUPPET_VERSION']}" : '~> 6.0'
-if RUBY_VERSION < '2.1'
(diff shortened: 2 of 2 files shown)


ekohl:
I think you need to bump Puppet to version 6 in Travis or just leave it out since the Gemfile sets it correctly:
https://github.com/theforeman/foreman-installer/blob/473629dc2736c54ec75c211e581a39cd986c9d33/.travis.yml#L12-L13

Might also be a good time to drop the Ruby < 2.4 hacks from Gemfile.

ehelms:
Updated Gemfile. We can do another follow up to bump the version of Rubocop we are using as well.

Gemfile
```diff
@@ -3,27 +3,17 @@ source 'https://rubygems.org'
...
 gem 'rake'
-gem 'rdoc', '< 6' if RUBY_VERSION < '2.2'
 
 group :test do
   gem 'rspec'
   gem 'rubocop', '~> 0.50.0'
 end
 
 group :development do
-  # Needed to pin dependencies
```
ekohl, line 20: This comment is still relevant: they're here for `rake pin_modules`. I agree pinning can be a bit overloaded in this context.

Gemfile
```diff
@@ -3,27 +3,17 @@ source 'https://rubygems.org'
 gem 'kafo', '~> 4.1'
 gem 'librarian-puppet'
 gem 'puppet', ENV.key?('PUPPET_VERSION') ? "~> #{ENV['PUPPET_VERSION']}" : '~> 6.0'
-if RUBY_VERSION < '2.1'
-  gem 'puppet-strings', '< 2.0.0'
-else
-  gem 'puppet-strings'
-end
+gem 'facter', '~> 4.0'
```
ekohl: Technically it should work with Facter 3 as well and in some installs you can use that. It just doesn't work with Facter 2 anymore.
```suggestion
gem 'facter', '>= 3.0'
```

ehelms:
Errors:

```
apply --config=/tmp/kafo_installation20200417-9343-1i3ff5u/puppet.conf 
[ERROR 2020-04-17T13:21:56 main] Warning: /tmp/tmp.YNaFfplrd6/etc/foreman-installer/custom-hiera.yaml: file does not contain a valid yaml hash
Error: Evaluation Error: Error while evaluating a Function Call, Is a directory @ io_fread - /tmp/tmp.YNaFfplrd6/share/foreman-installer/config/foreman.hiera (line: 148, column: 72) on node travis-job-93be06d8-0449-437a-b392-26940e5bee26.c.travis-ci-prod-2.internal
```

@ekohl seeing if you recognize this error from switching to Facter 4

ekohl:
That used to be a warning. It's surprising to me that that's become an error. I thought about it, but not sure how we can ship an example YAML that doesn't have something like a dummy key or `{}` and still get a valid hash.

ehelms:
I fixed that issue with a test env replacement into custom-hiera. Now its on to the next error that is the one I see locally:

```
[ERROR 2020-04-17T18:58:11 main] Error: Evaluation Error: Error while evaluating a Function Call, Is a directory @ io_fread - /tmp/tmp.YizEDJ9urQ/share/foreman-installer/config/foreman.hiera (line: 148, column: 41) on node travis-job-1b974b62-1ea6-4106-9c17-bf74fc99fb3c.c.travis-ci-prod-3.internal
```

ehelms:
@ekohl any ideas about this error? the directory exists but it seems to want to think it's a file? or should be a file

ehelms:
I think this is now blocking all PR tests.

.travis.yml
```diff
@@ -9,8 +9,6 @@ addons:
     - docbook-xsl
     - ca-certificates
 rvm: 2.4
```
ekohl:
```suggestion
rvm: 2.5
```

ekohl (approved):
We can also drop the `sudo: false`

PR #501, opened 2020-04-28
Refs #29541: Use Facter 4

ehelms:
Facter 3 from gem does not install easily on system as it has a
lot of build dependencies.


ekohl (changes requested):
This shouldn't be needed. Facter 3 is still ok to use and Puppet 5 hard requires `Facter < 4` so it would be impossible to install that combination anyway.

ehelms:
Then let's just drop from source testing like we did here in PR testing - https://github.com/theforeman/foreman-infra/pull/1351

PR #504, opened 2020-05-14
Add pulpcore data reset feature

ianballou:
The `docroot` directory stores all of the pulpcore content as artifacts.  This should happen only after a pulp database reset, which seems to be the case.

Questions: 
Is the `app_value(:clear_pulpcore_content)` okay? I'm not sure where it comes from, perhaps as a command line argument?

### DIFF
--- a/katello/hooks/pre/11-clear_pulpcore_data_feature.rb
+++ b/katello/hooks/pre/11-clear_pulpcore_data_feature.rb
@@ -1,0 +1,12 @@
+# See bottom of the script for the command that kicks off the script
+
+def clear_pulpcore_content
+  if File.directory?('/var/lib/pulp/docroot')
+    FileUtils.rm_rf(Dir.glob("/var/lib/pulp/docroot/*"))
+    logger.info 'Pulpcore content successfully removed.'
+  else
+    logger.warn 'Pulpcore content directory not present at \'/var/lib/pulp/docroot\''
+  end
+end
+
+clear_pulpcore_content if app_value(:clear_pulpcore_content) && !app_value(:noop) && module_enabled?('foreman_proxy_content')
--- a/katello/hooks/boot/11-clear_pulpcore_data_hook.rb
+++ b/katello/hooks/boot/11-clear_pulpcore_data_hook.rb
@@ -1,0 +1,7 @@
+# Add reset option
+app_option(
+  '--clear-pulpcore-content',
+  :flag,
+  'This option will clear all Pulpcore content from disk located in \'/var/lib/pulp/docroot/\'.',
+  :default => false
+)


ehelms:
To add the parameter, you will want something like https://github.com/theforeman/foreman-installer/blob/develop/katello/hooks/boot/11-clear_pulp_data_hook.rb

Please also open and add a Redmine issue

katello/hooks/pre/11-clear_pulpcore_data_feature.rb
```diff
@@ -0,0 +1,12 @@
...
+def clear_pulpcore_content
+  if File.directory?('/var/lib/pulp/docroot')
+    `rm -rf /var/lib/pulp/docroot/*`
+    logger.info 'Pulpcore content successfully removed.'
+  else
+    logger.warn 'Pulpcore content directory not present at \'/var/lib/pulp/docroot\''
+  end
+end
+
+clear_pulpcore_content if app_value(:clear_pulpcore_content) && !app_value(:noop)
```
ehelms: Could you add a `module_enabled?('foreman_proxy_content')` condition to the if check

ianballou: do you think clear_pulp_data_feature should have this too?

katello/hooks/pre/11-clear_pulpcore_data_feature.rb
```diff
@@ -0,0 +1,12 @@
+# See bottom of the script for the command that kicks off the script
+
+def clear_pulpcore_content
+  if File.directory?('/var/lib/pulp/docroot')
+    `rm -rf /var/lib/pulp/docroot/*`
```
ekohl: I don't like shelling out, especially since this discards any exit status. Perhaps `FileUtils.rm_rf` should be used? It also has docs on the security implications. Not sure if `rm -rf` has those too.

ekohl (commented):
There is no hook to create a command line parameter. Also wondering what the value is of just clearing the content on disk. Shouldn't the DB also be cleared?

katello/hooks/pre/11-clear_pulpcore_data_feature.rb
ianballou: @ekohl I've mirrored the `clear_pulp_content` method, should we address this issue there first?

And yeah the DB is cleared during the main reset task. Looks like clearing out the pulp content historically has been an add-on.

ianballou:
https://projects.theforeman.org/issues/29821

ianballou:
@ekohl @ehelms I've made some improvements.  `clear_pulpcore_data` differs a bit from `clear_pulp_data` now, so I had some questions about that above.

ehelms:
@wbclark With the work you did, I think we can close this?

wbclark:
@ehelms 

I implemented a complete Pulpcore data reset (including database and content on disk) as part of the larger `--reset-data` feature, but did not implement a way to only clear Pulpcore content from disk (equivalent to what `--clear-pulp-content` does).

To do it via the installer, that would require clearing the content from disk (this is where I simply copied Ian's work) and surgically updating the DB to forget that it has downloaded content, which per jlsherrill is not desirable since it would be an even worse hack than what we already do for Pulp2.

So it looks like Katello and Pulp agreed to implement https://pulp.plan.io/issues/5926 . The open question from my PoV is then whether the installer should wrap that in a `--clear-pulpcore-content` hook once it is ready (which would resolve https://projects.theforeman.org/issues/29821 ) or whether Katello should do it.

One advantage of still doing it via the installer is that it doesn't require the application to be currently running, while the advantage of doing it via the application is that it's simple point and click for the user (maybe we don't actually want to make it too simple though?).

@ianballou any correction or further comment?

ehelms:
My feel is that for Pulp 3 use case, with that issue, it should be handled by the application and not by the installer which should stick to focusing on installation only.

ianballou:
With the work that @wbclark did, I don't think we need this PR anymore.  The Pulpcore data reset (as part of this PR) wasn't meant to run without a database reset anyway.

ekohl:
Indeed. The application(s) will handle this at runtime and there's no need to do this in the installer. Thanks everyone for figuring out the path forward.

PR #507, opened 2020-05-26
Refs #29512 - Add a default to stop_services

ekohl:
d687c2fa56efb223781078564f45f19310df0306 added this but it's also called without arguments.

Fixes #29906


wbclark (approved):
LGTM.

PR #508, opened 2020-05-27
Fixes #29941 - Check modules are enabled

ekohl:
param_value() can return an answer for a disabled module. The result is that PostgreSQL is being installed on a Foreman Proxy when it's not needed.

### DIFF
--- a/hooks/boot/01-kafo-hook-extensions.rb
+++ b/hooks/boot/01-kafo-hook-extensions.rb
@@ -49,8 +49,16 @@ module HookContextExtension
     module_enabled?('katello_devel')
   end
 
+  def local_foreman_db?
+    foreman_server? && param_value('foreman', 'db_manage')
+  end
+
+  def local_candlepin_db?
+    module_enabled?('katello') && param_value('katello', 'candlepin_manage_db')
+  end
+
   def local_postgresql?
-    param_value('foreman', 'db_manage') || param_value('katello', 'candlepin_manage_db') || devel_scenario?
+    local_foreman_db? || local_candlepin_db? || devel_scenario?
   end
 
   def local_redis?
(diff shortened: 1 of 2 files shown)


ehelms:
Should we file an issue with Kafo to fix param_value to only check enabled modules?

ekohl:
param_value is not in kafo itself:
https://github.com/theforeman/foreman-installer/blob/94207e8ba6b153c3b8c86231fdb18931c1ec22d5/hooks/boot/01-kafo-hook-extensions.rb#L5-L8

I wondered about this and not sure if there are cases where it makes sense. Scoping and its lifecycle might make it unpredictable. Because this is a regression introduced in 2.0 I've tried to keep it small so it's safer to cherry pick.

ehelms:
Apologies, I should have indicated I am good with this change as is, but wanted to ask/encourage filing an issue to Kafo to fix this at a core Kafo hook level as part of the work @wbclark is starting to both merge hooks and migrate functionality to Kafo.

ekohl:
Before we decide what we need, I'd prefer to clean up all the Pulp 2 code first (where cleaning up means removing it).

ehelms:
I don't understand what Pulp 2 has to do with this core hook functionality. I'll file the issue.

ekohl:
Now we're both going off topic, but IMHO we should first clean up Pulp 2 related code before we decide what we actually want from Kafo. Perhaps there are various patterns that are in the current hooks that we don't really need anymore or should be done in a better way.

ehelms (approved):

tbrisker:
2.0 - 233aa162c8d45f3b4481b8a7be90911cc030abd4
2.1 - 710077c7da1ef0354fe2f40b9df5a553b8f0c114

PR #509, opened 2020-05-27
Fixes #29797 - Merge --reset-foreman-db and --reset into --reset-data

wbclark:


hooks/pre/10-reset_data.rb
```diff
@@ -5,10 +5,10 @@ def reset
   if foreman_installed?
 
     stop_services
-    start_services(['postgresql'])
+    start_services(['postgresql']) if local_postgresql?
     reset_database
```
ekohl: You should also check if Foreman is enabled.
```suggestion
    reset_foreman if foreman_server?
```

Note you also need to rename `reset_database`

hooks/pre/10-reset_data.rb
```diff
@@ -176,8 +176,8 @@ def empty_database!(config)
   system(pg_sql_statement(config, delete_statements)) if delete_statements
 end
 
-if app_value(:reset) && !app_value(:noop)
-  response = ask('Are you sure you want to continue? This will drop the databases, reset all configurations that you have made and bring the server back to a fresh install. [y/n]')
+if app_value(:reset_data) && !app_value(:noop)
+  response = ask('Are you sure you want to continue? This will drop the databases, reset all configurations that you have made and bring all application data back to a fresh install. [y/n]')
   if response.downcase != 'y'
     $stderr.puts '** cancelled **'
     kafo.class.exit(1)
```
ekohl:
```suggestion
    exit(1)
```

hooks/pre/10-reset_data.rb
```diff
@@ -176,8 +176,8 @@ def empty_database!(config)
   system(pg_sql_statement(config, delete_statements)) if delete_statements
```
ekohl: We should not use `system()` without checking whether it succeeded

ekohl (changes requested):
Foreman has a rake task to drop the database. I don't know if it works on remote DBs, but if it does, should we prefer it? On the plus side, we reuse logic. On the downside, it requires an updated config file which we don't know for sure at this point.

We should also replace `Kafo::KafoConfigure.logger` with just `logger` since that also works.

hooks/pre/10-reset_data.rb
```diff
@@ -5,10 +5,10 @@ def reset
   if foreman_installed?
```
ekohl: This also feels like bad naming

hooks/boot/10-reset_data.rb
```diff
@@ -1,8 +1,7 @@
-# Add reset option
-app_option(
-  '--reset',
+add_option(
+  '--reset-data',
   :flag,
-  'This option will drop the Katello database and clear all subsequent backend data stores. ' +
+  'This option will drop all databases for Foreman and, when present, Katello. ' +
```
ehelms: This opens up to adding in other plugins if they have some sort of data. I tend to prefer just mentioning Foreman and referencing backend systems as a general thing that will get their data cleared.

ehelms:
I tend to agree, foreman-rake db:drop should be enough to handle database being dropped in local and remote scenarios. I have no view into the past as to why it was implemented the way it is today.

hooks/boot/10-reset_data.rb
```diff
@@ -1,8 +1,7 @@
-# Add reset option
-app_option(
-  '--reset',
+add_option(
```
ehelms: The tests are failing on this. It's, despite your intuition, `app_option`. But I totally understand thinking `add_option` should be the name.

ekohl: There's something to say for both. One is declarative, the other imperative. Given the rest of the hooks I think imperative feels more in line with the rest.

hooks/pre/10-reset_data.rb
wbclark: Thanks. I've changed this to use `execute()` instead.

hooks/pre/10-reset_data.rb
wbclark: In the latest version I've pushed, I've changed this function to check for `which foreman-maintain` rather than `which foreman-rake`, which has the benefit we should now be able to reuse `--reset-data` on a smart proxy to reset pulp{,core} data.

ehelms (approved):
I think this looks good.

hooks/pre/10-reset_data.rb
```diff
@@ -2,22 +2,22 @@
 require 'English'
 
 def reset
-  if foreman_installed?
+  if foreman_maintain_present?
```
ekohl: I'll be more verbose: previously this only worked if `foreman-rake` was present. That's only provided by the Foreman RPM. However, this can clear out Pulp and Pulpcore as well which in the future will be present on hosts without Foreman as well. This entire check is now no longer needed.

hooks/pre/10-reset_data.rb
```diff
@@ -2,22 +2,22 @@
...
 
     stop_services
```
ekohl: This relies on foreman-maintain and the installer can't depend on that. It simply doesn't exist on Debian. That's a regression for the regular (non-katello) installer users.

hooks/pre/10-reset_data.rb
```diff
@@ -31,14 +31,14 @@ def load_foreman_config
   db_config
 end
 
-def reset_database
+def reset_foreman_database
   Kafo::KafoConfigure.logger.info 'Dropping database!'
```
ekohl: I made a comment about the logger before, but that was meant for all instances in the entire file

ekohl (changes requested):
:-1: on relying on foreman-maintain.

hooks/pre/10-reset_data.rb
wbclark: OK great. I originally changed the check to foreman-maintain because I believe it should be present on smart proxies while foreman-rake is not. But I wasn't aware that foreman-maintain doesn't exist on Debian.

With that said, even without this check, we are still relying on foreman-maintain -- it's used in the definition of stop_services which runs unconditionally, so if that's not present on Debian this will still fail.

hooks/pre/10-reset_data.rb
```diff
@@ -2,22 +2,22 @@
...
 def reset
-  if foreman_installed?
+  if foreman_maintain_present?
 
     stop_services
-    start_services(['postgresql'])
-    reset_database
-    reset_candlepin
-    reset_pulp
+    start_services(['postgresql']) if foreman_server? && local_postgresql?
```
ehelms: For now, I could see wrapping `stop_services` and `start_services` in the if foreman-maintain present check. 

Will clearing the Foreman database cover any Dynflow related data? @adamruzicka

hooks/pre/10-reset_data.rb
ehelms: @kgaikwad We really need foreman-maintain on Debian to prevent this kind of incongruence within our implementation.

ekohl: I'd strongly prefer to get rid of the foreman-maintain dependency in `stop_services` altogether. We can simply come up with a list of services and replace it all with:
```
def stop_services(services)
  running = services.filter do |service|
    system('systemctl', 'is-active', service, in: '/dev/null', out: '/dev/null', err: '/dev/null')
  end
  system('systemctl', 'stop', *services, in: '/dev/null', out: '/dev/null', err: '/dev/null')
end
```

All the priority things that foreman-maintain does is IMHO not not needed and systemd can figure it out. It does so on system shutdown and there it's not a problem.

hooks/pre/10-reset_data.rb
wbclark: I can answer that @ehelms, the answer is yes it will.

Dynflow data lives in these tables: dynflow_envelopes,dynflow_delayed_plans,dynflow_steps,dynflow_actions,dynflow_execution_plans,foreman_tasks_locks,foreman_tasks_tasks

All of those reside in the main foreman DB

hooks/pre/10-reset_data.rb
kgaikwad: @ehelms, 
I understand your concern. Though we resumed our work on upstream support for upgrade scenarios however we are currently focusing on rhel & centos support first. So, it may take little more time to have Debian support.

ekohl: I'm preparing a PR that removes the need for it, writing tests now.

ekohl: https://github.com/theforeman/foreman-installer/pull/510 needs polishing and testing.

ehelms:
I got to thinking about this a bit more. Why do we support data reset through the installer? All other data manipulation (backup, restore, health checks) we do through foreman-maintain. We could drop this complexity here and address some of the concerns by relegating that functionality to that tool.

ekohl:
I had a similar thought.

hooks/pre/10-reset_data.rb
adamruzicka: With sidekiq in place, it would be good to restart redis so it gets cleared as well

hooks/pre/10-reset_data.rb
```diff
@@ -31,14 +31,14 @@ def load_foreman_config
...
+def reset_foreman_database
+  logger.info 'Dropping database!'
 
   config = load_foreman_config
   if remote_host?(config[:host])
     empty_database!(config)
   else
-    execute('DISABLE_DATABASE_ENVIRONMENT_CHECK=1 foreman-rake db:drop 2>&1')
+    execute("sudo -u postgres dropdb #{config[:database]}")
   end
```
ekohl: This logic is the same as `reset_pulpcore` and `empty_candlepin_database`, other than the `load_x_config` part. I'd unify that. You can probably also add a more accurate logger there than just `Dropping database!`, which is ambiguous when you're dropping 3 databases.

ekohl (changes requested):
This is blocked on either #510, packaging foreman-maintain for Debian or some other solution. My preference is #510, but it's important to me that we don't regress in the `foreman` scenario on the supported platforms.

ehelms:
@wbclark The work from @ekohl to refactor how services are handled has been merged, I think you can now revisit an dupdate this PR.

hooks/pre/10-reset_data.rb
wbclark: Thanks @adamruzicka . With full service restarts in place, we should be good then.

hooks/pre/10-reset_data.rb
wbclark: I've unified loading config for all 3 postgresql databases into a single method `load_db_config(db)` and resetting all 3 postgresql databases into a single method `empty_db_in_postgresql(db)`

hooks/pre/10-reset_data.rb
```diff
@@ -2,58 +2,37 @@
...
-end
-
-def foreman_installed?
-  `which foreman-rake > /dev/null 2>&1`
-  $CHILD_STATUS.success?
-end
-
-def load_foreman_config
+  stop_services
+  start_services(['postgresql']) if foreman_server? && local_postgresql?
```
ekohl: Why `foreman_server?` Isn't this sufficient?
```suggestion
  start_services(['postgresql']) if local_postgresql?
```

hooks/pre/10-reset_data.rb
```diff
@@ -2,58 +2,37 @@
...
+def load_db_config(db)
+  module_name, user_param, name_param, param_prefix = if db == 'foreman'
+                  'foreman', 'username', 'database', 'db_'
+                elsif db == 'candlepin'
+                  'katello', 'user', 'db_name', 'candlepin_db_'
+                elsif db == 'pulpcore'
+                  'foreman_proxy_content', 'user', 'db_name', 'pulpcore_postgresql_'
+                else
+                  raise "installer module unknown for db: #{db}"
+                end
```
ekohl: This feels about as complex as duplicating the config

hooks/pre/10-reset_data.rb
```diff
@@ -2,58 +2,37 @@
...
-  db_config[:port] = param_value('foreman', 'db_port')
-  db_config[:database] = param_value('foreman', 'db_database') || 'foreman'
-  db_config[:username] = param_value('foreman', 'db_username')
-  db_config[:password] = param_value('foreman', 'db_password')
+  db_config[:host] = param_value(module_name, "#{param_prefix}host") || 'localhost'
+  db_config[:port] = param_value(module_name, "#{param_prefix}port") || 5432
+  db_config[:database] = param_value(module_name, "#{param_prefix}#{name_param}") || db
+  db_config[:username] = param_value(module_name, "#{param_prefix}#{user_param}")
+  db_config[:password] = param_value(module_name, "#{param_prefix}password")
   db_config
```
ekohl: I'd prefer to avoid the `db_config` variable. Just return
```ruby
{
  host: param_value(),
  port: param_value(),
  # ...
}

File: hooks/pre/10-reset_data.rb [hunk 15]
```diff
@@ -62,14 +41,13 @@ def empty_candlepin_database
 end
 
 def reset_candlepin
-  Kafo::KafoConfigure.logger.info 'Dropping Candlepin database!'
   commands = [
     'rm -f /var/lib/candlepin/cpdb_done',
     'rm -f /var/lib/candlepin/cpinit_done',
     'systemctl stop tomcat',
```
ekohl: We already have `stop_services` in the beginning. Is this still needed? (similar thing in reset_pulp`)

hooks/pre/10-reset_data.rb
```diff
@@ -62,14 +41,13 @@ def empty_candlepin_database
 end
 
 def reset_candlepin
-  Kafo::KafoConfigure.logger.info 'Dropping Candlepin database!'
```
ekohl: You drop this line, but keep it in Pulp. Feels inconsistent

hooks/pre/10-reset_data.rb
```diff
@@ -62,14 +41,13 @@ def empty_candlepin_database
...
   commands = [
     'rm -f /var/lib/candlepin/cpdb_done',
     'rm -f /var/lib/candlepin/cpinit_done',
```
ekohl: We can combine this in a single command

hooks/pre/10-reset_data.rb
```diff
@@ -2,58 +2,49 @@
...
+  stop_services
+  start_services(['postgresql']) if local_postgresql?
+  empty_db_in_postgresql('foreman') if foreman_server?
+  reset_candlepin if module_enabled?('katello')
+  reset_pulp if pulp_enabled?
+  empty_db_in_postgresql('pulpcore') if pulpcore_enabled?
+end
+
+def load_db_config(db)
+  param_helper = if db == 'foreman'
```
ekohl: I still question whether it isn't easier to have 3 methods. If you do go with this method, I'd go with `case`:
```ruby
param_helper = case db
               when 'foreman'
                 {} # full thing
               when 'candlepin'
                 {} # full thing
               when 'pulpcore'
                 {} # full thing
               else
                 raise
               end
```

I'd also consider not adding a `param_helper` hash but assign 4 variables.

hooks/pre/10-reset_data.rb
wbclark: It's not dropped, it's moved to `empty_db_in_postgresql`. Pulp2 doesn't use that function so I kept it there.

ekohl: Right. Makes sense. You could keep things the same by moving it in `empty_mongo` as well.

hooks/pre/10-reset_data.rb
wbclark: This seems a lot more readable. Thanks

hooks/pre/10-reset_data.rb
```diff
@@ -0,0 +1,143 @@
...
+    raise "installer module unknown for db: #{db}"
+  end
+  {
+    host: param_value(module_name, "#{param_prefix}host") || 'localhost'
+    post: param_value(module_name, "#{param_prefix}port") || 5432
+    database: param_value(module_name, "#{param_prefix}#{name_param}") || db
+    username: param_value(module_name, "#{param_prefix}#{user_param}")
+    password: param_value(module_name, "#{param_prefix}password")
+  }
+end
```
wbclark: I think at this point it's sufficiently readable and simpler than loading config separately for all 3 databases.

Thanks for the feedback on it @ekohl . At this point is it satisfactory to you?

wbclark: Fixed typos...

ekohl: Yes, though I'd keep a empty line after `end` (from `case`) to have a clear visual split.

wbclark: Done.

ekohl: Also, looking at it we may want to align the parameter names at some point. I always liked `username` because it's the same length as `password`. Let's keep this in mind for the refactor.

Btw, candlepin also suggests you'll end up with `candlepin_db_db_name`.

hooks/pre/10-reset_data.rb
```diff
@@ -0,0 +1,144 @@
...
+  when 'foreman'
+    module_name  = 'foreman'
+    user_param   = 'username'
+    name_param   = 'database'
+    param_prefix = 'db_'
+  when 'candlepin'
+    module_name  = 'katello'
+    user_param   = 'user'
+    name_param   = 'db_name'
+    param_prefix = 'candlepin_db_'
```
ekohl: Looks like this will result in `candlepin_db_db_name`

hooks/pre/10-reset_data.rb
```diff
@@ -0,0 +1,144 @@
...
+  reset_pulp if pulp_enabled?
+  empty_db_in_postgresql('pulpcore') if pulpcore_enabled?
+end
+
+def load_db_config(db)
+  case db
+  when 'foreman'
+    module_name  = 'foreman'
+    user_param   = 'username'
+    name_param   = 'database'
```
ekohl: I'd suggest something like `db_param` because `name` is ambiguous.

hooks/pre/10-reset_data.rb
wbclark: Do you mean alignment, as in adjusting whitespace in this file to make it more readable?

Or alignment as in, using common parameter names across the 3 modules so that the entire case block here would not be needed? Because I definitely agree, and now feels like the time to do it while we are already removing pulp2/mongodb and unifying around postgresql.

wbclark: I guess the whitespace is a nonstarter, as the linter says we cannot have nice things...

ekohl: > Or alignment as in, using common parameter names across the 3 modules so that the entire case block here would not be needed? Because I definitely agree, and now feels like the time to do it while we are already removing pulp2/mongodb and unifying around postgresql.

This.

wbclark: @ekohl agreed. Opened: https://projects.theforeman.org/issues/30388

ekohl:
Note: foreman-maintain calls `--reset` which now needs to be modified.

PR #510, opened 2020-05-29
Fixes #29963 - Manage services without foreman-maintain

ekohl:
Currently foreman-maintain is called to stop and start services in the EL7 upgrade. However, foreman-maintain is unavailable on Debian. It's also possibly a circular problem because foreman-maintain depends on the installer for certain actions so it is possible to end up in an infinite loop.

Successor to https://github.com/theforeman/foreman-installer/pull/461

Currently not entirely tested. The list of service names is also likely to be incomplete.

### DIFF
--- a/hooks/boot/04-services.rb
+++ b/hooks/boot/04-services.rb
@@ -1,0 +1,44 @@
+require 'open3'
+
+module ServicesHookContextExtension
+  ALL_POSSIBLE_SERVICES = [
+    '*mongo*', # Used by Pulp 2
+    '*redis*', # Used by Foreman/Dynflow and Pulpcore
+    'apache.service', # Apache on Debian
+    'dynflow*',
+    'foreman*',
+    'httpd.service', # Apache on Red Hat
+    'postgresql*', # Used by Foreman/Dynflow and Pulpcore
+    'pulp*',
+    'qdrouterd.service', # Used by Pulp 2
+    'qpidd.service', # Used by Pulp 2
+    'smart_proxy_dynflow_core.service', # Used by Foreman Proxy
+    'tomcat.service', # Candlepin
+    'squid.service', # Used by Pulp 2
+  ].freeze
+
+  def start_services(services)
+    raise "Services must be specified" if services.empty?
+
+    logger.debug("Starting services #{services.join(', ')}")
+    stdout_and_stderr_str, status = Open3.capture2e('systemctl', 'start', *services)
+    fail_and_exit("Failed to start services: #{stdout_and_stderr_str}", status.exitstatus) unless status.success?
+  end
+
+  def stop_services(services = ALL_POSSIBLE_SERVICES)
+    raise "Can't stop all services" if services.empty?
+
+    logger.debug('Getting running services')
+    stdout_str, stderr_str, status = Open3.capture3('systemctl', 'list-units', '--no-legend', '--type=service,socket', '--state=running', *services)
+    fail_and_exit("Failed to get running services: #{stderr_str}", status.exitstatus) unless status.success?
+    running = stdout_str.lines.map { |line| line.split.first }
+    logger.debug("Found running services #{running.inspect}")
+    return if running.empty?
+
+    logger.debug("Stopping running services #{running.join(', ')}")
+    stdout_and_stderr_str, status = Open3.capture2e('systemctl', 'stop', *running)
+    fail_and_exit("Failed to stop running services: #{stdout_and_stderr_str}", status.exitstatus) unless status.success?
+  end
+end
+
+Kafo::HookContext.send(:include, ServicesHookContextExtension)
--- a/hooks/boot/03-foreman-maintain-extensions.rb
+++ b/hooks/boot/03-foreman-maintain-extensions.rb
@@ -15,17 +15,6 @@ module ForemanMaintainHookContextExtension
     foreman_maintain('packages unlock --assumeyes', true)
   end
 
-  def start_services(services)
-    foreman_maintain("service start --only=#{services.join(',')}", true)
-  end
-
-  def stop_services(services = nil)
-    command = "service stop"
-    command = "service stop --only=#{services.join(',')}" if services
-
-    foreman_maintain(command, true)
-  end
-
   def foreman_maintain(command, exit_on_fail = false)
     command = "foreman-maintain #{command}"
     status = execute_command(command, false, true)
--- a/spec/services_hook_extensions_spec.rb
+++ b/spec/services_hook_extensions_spec.rb
@@ -1,0 +1,184 @@
+require 'spec_helper'
+require 'kafo/hook_context'
+require_relative '../hooks/boot/04-services'
+
+describe 'ForemanMaintainHookContextExtension' do
+  let(:kafo) { instance_double('KafoConfigure') }
+  let(:logger) { instance_double('Logger') }
+  let(:context) { Kafo::HookContext.new(kafo) }
+
+  before do
+    allow(context).to receive(:logger).and_return(logger)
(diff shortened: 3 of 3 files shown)


ekohl:
After this is merged, we can also drop https://github.com/theforeman/foreman-packaging/blob/9a6c36d20882630001d61d59f81f1e3197f603e5/packages/foreman/foreman-installer/foreman-installer.spec#L26

hooks/boot/04-services.rb
```diff
@@ -0,0 +1,36 @@
+require 'open3'
+
+module ServicesHookContextExtension
+  ALL_POSSIBLE_SERVICES = [
```
ekohl, line 4: I already know qpid and friends is currently missing. Question is, do we wait for qpid to disappear or also list it here?

ehelms:
And add it here https://github.com/theforeman/foreman-packaging/blob/9a6c36d20882630001d61d59f81f1e3197f603e5/packages/foreman/foreman-installer/foreman-installer.spec#L50

I get the theoretical concern around circular problems (though in a couple years we have yet to face any). And I get the issues with foreman-maintain not being available on Debian (though outside of packaging it I don't see why it couldn't be present on Debian). 

My concern is that this duplicates logic and it creates division between Foreman and Katello scenarios where our goal has been to bring congruence.

hooks/boot/04-services.rb
```diff
@@ -0,0 +1,36 @@
...
+    'httpd.service', 'apache.service', # RH vs Debian
+    'postgresql*',
+    'foreman*',
+    'dynflow*',
+    '*mongo*',
+    '*redis*',
+    'pulp*',
```
ehelms, line 12: If we are considering all services that are stopped as of today:

 * tomcat
 * puppet
 * puppetserver
 * smart_proxy_dynflow_core

Things going away soon:

 * qpidd
 * qdrouterd
 * squid

ekohl:
> I get the theoretical concern around circular problems (though in a couple years we have yet to face any).

Likely because until recently the installer didn't depend on foreman-maintain.

> My concern is that this duplicates logic and it creates division between Foreman and Katello scenarios where our goal has been to bring congruence.

On the other hand, I'd argue this is much better logic since it's much more focused and it utilizes the parallelization that systemd offers.

Also note that we don't care about starting all services, just the individual ones we need for an action. That means we can get away with globbing.

We can also improve things. For example, resetting the databases shouldn't require us to stop Apache. If the foreman service is unavailable, we can render a proper HTTP 502 page. Resetting the databases also doesn't require us to stop puppet and puppetserver. If we're going that route, there really is no point in using foreman-maintain.

ehelms:
> Likely because until recently the installer didn't depend on foreman-maintain.
> 
> >

Just adding data to the discussion, been relying on it since over 1.5+ years, dating back to the great merging of the installers even (https://github.com/theforeman/foreman-installer/commit/e8ce73a008d882aaf14ce31198aeac5624fd28a3#diff-ceb7d70cfab1a466aef1539a90401046)

hooks/boot/04-services.rb
ehelms: Can you add tomcat to the list?

hooks/boot/04-services.rb
```diff
@@ -0,0 +1,36 @@
...
+  ]
+
+  def start_services(services)
+    raise "Services must be specified" if services.empty?
```
ehelms, line 21: Are you expecting Kafo to catch this and deal with it? Otherwise I think we can to use `fail`

hooks/boot/04-services.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    logger.debug("Starting services #{services.join(', ')}")
+    stdout_and_stderr_str, status = Open3.capture2e('systemctl', 'start', *services)
+    fail_and_exit("Failed to start services: #{stdout_and_stderr_str}", status.exitstatus) unless status.success?
```
ehelms, line 25: Whats the reason behind the combined stdout and stderr here? (vs just putting stderr out)

ehelms:
Can you fix the tests up.

ekohl:
> Can you fix the tests up.

That's IMHO just Rubocop being stupid.

hooks/boot/04-services.rb
ekohl: Passing an empty array is incorrect API use and the goal of this is to slap the developer. Users should not see this. The alternative is to just return.

hooks/boot/04-services.rb
ekohl: I wasn't quite sure about the output. just grabbing the entire output allows to debug was my reasoning. Didn't think it through that much.

ehelms:
Rules are rules it blocks this change from going in :)

hooks/boot/04-services.rb
ehelms: Suppose its a starting place.

hooks/boot/04-services.rb
ehelms: This was mostly around the nuance of raise vs fail in Ruby -- https://gist.github.com/Integralist/8058ac54088408d9aadc

ekohl: TIL. Thanks!

ekohl:
So I've updated it and the tests pass. I'll see about testing this on a real system.

ehelms:
Is this ready to be taken out of Draft?

ekohl:
Code wise I think this is ready but I haven't tested this myself on an actual system.

ekohl:
Testing on a real system showed that I had 2 issues:
* Empty output was not dealt with
* The incorrect variable was used (services instead of running)

Will update and expand the tests to cover this.

ekohl:
Now tested on a real system, both with httpd.service running and not running.

hooks/boot/04-services.rb
```diff
@@ -0,0 +1,44 @@
...
+    'postgresql*', # Used by Foreman/Dynflow and Pulpcore
+    'pulp*',
+    'qdrouterd.service', # Used by Pulp 2
+    'qpidd.service', # Used by Pulp 2
+    'smart_proxy_dynflow_core.service', # Used by Foreman Proxy
+    'tomcat.service', # Candlepin
+    'squid.service', # Used by Pulp 2
+  ].freeze
+
+  def start_services(services)
```
ehelms, line 20: Can this default to ALL_POSSIBLE_SERVICES similar to stop_srevices?

ehelms: I see this doesn't handle well from systemd:

```
Failed to start services: Failed to start apache.service: Unit not found.
Failed to start smart_proxy_dynflow_core.service: Unit not found.
[ERROR 2020-07-01T14:57:11 verbose] Failed to start services: Failed to start apache.service: Unit not found.
Failed to start smart_proxy_dynflow_core.service: Unit not found.
```

So while I feel it should be possible to just start everything, forcing a developer to define what to start is OK too given this response from systemd.

ekohl: > Can this default to ALL_POSSIBLE_SERVICES similar to stop_srevices?

IMHO no since this is mostly used to start services prior to a migration. In that case you only want to start really what's needed.
```
$ rg start_services
spec/services_hook_extensions_spec.rb
15:  context 'start_services' do
16:    subject { context.start_services(services) }

hooks/boot/04-services.rb
20:  def start_services(services)

hooks/pre/30-el7_upgrade_postgresql.rb
2:  start_services(['postgresql'])

katello/hooks/pre/10-reset_feature.rb
8:    start_services(['postgresql'])
```

ekohl:
Updated with corrected indentation in the tests to please Rubocop.

ehelms (approved):

PR #511, opened 2020-05-29
Fixes #29798 - Move --lock-package-versions to hooks/

wbclark:


ehelms:
As part of this move two thoughts:

 1) Move https://github.com/theforeman/foreman-installer/blob/29f1bf12d80ac6427ebde7b0092c8c59602c8e5f/hooks/boot/09-version_locking.rb#L1 to a helper

 2) Wrap the logic of the hook in the check if the option exists and is true

ekohl:
>     1. if system('command -v foreman-maintain > /dev/null')

Alternatively, you can at least use `system('....', out: '/dev/null')`

wbclark:
>     2. Wrap the logic of the hook in the check if the option exists and is true

Currently the main logic of the hook is executed or not based on the custom_config_value.

If the cli_param exists, the custom_config_value will always be set to match it.

What should be the behavior if the cli_param does not exist but the custom_config_value is defined as true? As currently written, the installer will still try to execute the hook. Is that intended?

wbclark:
Thinking about it further, the cli_param should always exist iff foreman-maintain is installed.

So I agree with making the change to execute it only when the cli_param exists and has value true.

But in that case, what is the purpose of the custom_config_value existing at all?

ekohl:
The CLI param does not exist either if foreman-maintain is not installed. While currently the installer requires it in RPMs, I've made it very clear that I really want to get rid of that. The installer should not mandate foreman-maintain.

The hook does the right thing by gracefully ignoring it and doesn't lock. IMHO that's the expected behavior since locking is not something I'd want on my own systems and I think many community members share that opinion.

ekohl:
> But in that case, what is the purpose of the custom_config_value existing at all?

To pass state.

There are a few scenarios to consider at the start of the installer:
* foreman-maintain is not installed
* foreman-maintain is installed
  * The lock feature is not present. This was relevant when we wrote the patches, installer went in before foreman-maintain with it was released - might be relevant as well on Debian if f-m is released             
  * Packages are locked
  * Packages are unlocked

If packages are locked, the installer will unlock them. Then it needs to remember to lock them after running the installer. That's done using `custom_config_value`.

ehelms (approved):

hooks/pre_commit/09-version_locking.rb
```diff
@@ -8,6 +8,10 @@
   kafo.config.configure_application
 end
 
+if lock_versions && !foreman_maintain_installed?
+  fail_and_exit('Locking of package versions was requested but foreman-maintain is not installed')
+end
```
ekohl: I think this is not needed because `package_lock_feature?` implies `foreman_maintain_installed?`

hooks/boot/09-version_locking.rb
```diff
@@ -1,4 +1,4 @@
-if system('command -v foreman-maintain > /dev/null')
+if foreman_maintain_installed?
```
ekohl: Thinking out loud: should this be:
```suggestion
if package_lock_feature?
```

hooks/post/99-version_locking.rb
```diff
@@ -1,4 +1,4 @@
-if get_custom_config(:lock_package_versions)
+if package_lock_feature? && app_value(:lock_package_versions)
```
ekohl: I don't think this is needed since `app_value(:lock_package_versions)` should never be true if the feature isn't present (see the suggestion in the boot hook)

ekohl:
I took a stab at this in https://github.com/theforeman/foreman-installer/pull/536 with some additional improvements.

wbclark:
@ekohl shall I close this in favor of #536 ?

ekohl:
Yes, I think so.

PR #512, opened 2020-05-29
Fixes #30006 - Remove MongoDB Storage Engine Migration Hook

wbclark:


ekohl:
I wonder if we should start a Remove Pulp 2 tracker

ehelms:
A tracker for Pulp 2 removal will be good across installer and packaging. I think this can fall under hooks and just get it's own issue. If you haven't upgraded mongodb storage engine by now... oh well.

ekohl:
I still think this deserves a Redmine issue just to track what's going on.

ehelms:
@wbclark Could you attach a Redmine issue?

evgeni:
> If you haven't upgraded mongodb storage engine by now... oh well.

But, have we forced users to do so? There is nothing in the release notes of Katello (or Satellite, for that matter).

ekohl:
The next version of Katello will be 4.0 and drop Pulp 2 which removes the need for MongoDB.

ekohl (approved):
I don't think we should be cleaning up custom-hiera.yaml since this doesn't remove MongoDB.

evgeni:
Right, if that's 4.0+ only, then yeah.

PR #514, opened 2020-06-02
Fixes #29803 - Move --certs* to hooks/

wbclark:


hooks/pre/20-certs_update.rb
```diff
@@ -0,0 +1,66 @@
...
+  path = File.join(*[SSL_BUILD_DIR, hostname, cert_name].compact)
+  puts "Marking certificate #{path} for update"
+  if app_value(:noop)
+    puts "skipping in noop mode"
+  else
+    FileUtils.touch("#{path}.update")
+  end
+end
+
+if module_enabled?('katello')
```
ehelms: I think we could effectively move this to line 4 of this file and wrap all the logic in this check. That will prevent code from being run (e.g. `which katello-certs-check`) when it does not need to be run in other scenarios.

wbclark: Good point. I've made the change.

hooks/boot/20-certs_update.rb
```diff
@@ -0,0 +1,36 @@
+# Add options around regenerating certificates
+if module_enabled?('katello')
```
ekohl: Is this really correct? I think these cert options are also relevant in other cases, like content proxies. Isn't it sufficient to check if the certs module is enabled?
```suggestion
if module_enabled?('certs')
```

(repeated throughout this PR)

However, a potentially larger issue is access to variables in the boot stage. https://github.com/theforeman/kafo#hooks isn't explicit about this. IMHO these methods should raise exceptions when used without initialization.

hooks/pre/20-certs_update.rb
```diff
@@ -0,0 +1,66 @@
+require 'fileutils'
+require 'English'
+
+if module_enabled?('katello')
+  SSL_BUILD_DIR = param('certs', 'ssl_build_dir').value
+  CHECK_SCRIPT = `which katello-certs-check`.strip
```
ekohl: Why do we need `which`? Isn't it sufficient to use a non-qualified path later on?

hooks/pre/20-certs_update.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+  end
+  
+  if app_value('certs_update_server_ca') || app_value('certs_reset')
+    mark_for_update('katello-server-ca')
+  end
+  
+  if !app_value('certs_skip_check') &&
+     cert_file.to_s != "" &&
+     (app_value('certs_update_server_ca') || app_value('certs_update_server'))
```
ekohl: Just thinking out loud: does it make sense to just always check the certs? The script can also check for expired certs and in general it should be safe.

hooks/pre/20-certs_update.rb
```diff
@@ -0,0 +1,66 @@
...
+  CHECK_SCRIPT = `which katello-certs-check`.strip
+  
+  def mark_for_update(cert_name, hostname = nil)
+    path = File.join(*[SSL_BUILD_DIR, hostname, cert_name].compact)
+    puts "Marking certificate #{path} for update"
+    if app_value(:noop)
+      puts "skipping in noop mode"
+    else
+      FileUtils.touch("#{path}.update")
+    end
```
ekohl: I never like correlating multiple lines. Something with single lines like this has my preference:
```suggestion
    if app_value(:noop)
      puts "Marking certificate #{path} for update (noop)"
    else
      puts "Marking certificate #{path} for update"
      FileUtils.touch("#{path}.update")
    end
```

Note I kept the line mostly in tact so you are close to the original output. Alternatives are `Would mark certificate #{path} for update`.

hooks/boot/20-certs_update.rb
ehelms: What does used without initialization here mean?

ekohl: Thinking more about it, technically the boot hook shouldn't check for enabled but rather whether the module is present in the answers file. This is because answers are not yet determined when booting.

You can run `--no-enable-certs` (which would save the answers) and in the next run `--enable-certs --certs-update-server`. With the current approach that would fail because the option `--certs-update-server` is not added, but it's a valid flow. Probably more theoretical, but let's use this as a learning experience about how things are supposed to work.

[This](https://github.com/theforeman/kafo/blob/68ae1ad760f4073a6fedc9ce8cfe4f6ddd1a85e8/lib/kafo/hook_context.rb#L60-L66) is the source for `module_enabled?`. With that, it's easy to see how `module_present?(module_name)` would be implemented.

Again, not sure if at this stage it's really present. Easy to verify though:
```console
bundle update
rm -f Puppetfile.lock # TODO: might be nice if rake clean removed this when it's in .gitignore
bundle exec rake clean build
bundle exec ./bin/foreman-installer --scenario katello --help
```

Note it doesn't work no my machine because it fails on getting defaults (`This module does not support osfamily Gentoo`) but on Fedora it worked last time I checked.

ekohl: > What does used without initialization here mean?

During the boot phase (as you can see in the linked docs) kafo is not initialized. It means you don't have a logger (yet) and parameters have no values. Perhaps the modules are loaded since the configs are loaded but I'd really verify this.

wbclark: During `bundle exec rake clean build`,

```
W, [2020-07-13T13:57:12.530656 #2304132]  WARN -- : Parser cache ./_build/parser_cache/katello.yaml is from a different version of Kafo, skipping setup
W, [2020-07-13T13:57:13.552531 #2304132]  WARN -- : Parser cache ./_build/parser_cache/katello.yaml is from a different version of Kafo, skipping setup
W, [2020-07-13T13:57:14.595790 #2304132]  WARN -- : Parser cache ./_build/parser_cache/katello.yaml is from a different version of Kafo, skipping setup
```

And later,

```
W, [2020-07-13T13:58:12.694403 #2304347]  WARN -- : Parser cache ./_build/parser_cache/foreman-proxy-certs.yaml is from a different version of Kafo, skipping setup
W, [2020-07-13T13:58:13.721999 #2304347]  WARN -- : Parser cache ./_build/parser_cache/foreman-proxy-certs.yaml is from a different version of Kafo, skipping setup
```
Then, during `bundle exec ./bin/foreman-installer --scenario katello --help`:

`Could not get default values, check log file at ./_build//katello.log for more information`

Checking the log,

`[ERROR 2020-07-13T13:58:35 main] Could not get default values, cannot continue`

The Kafo version,

```
$ bundle info kafo
  * kafo (4.1.0)
	Summary: A gem for making installations based on puppet user friendly
	Homepage: https://github.com/theforeman/kafo
	Path: /usr/share/gems/gems/kafo-4.1.0
```

But I am unsure how to check what version was used with `./_build/parser_cache/{katellomforeman-proxy-certs}.yaml`

ekohl: Did you check `./_build//katello.log`?

wbclark: It turns out the issue is

```
$ grep Function _build/katello.log
[ERROR 2020-07-14T10:54:13 main] Error: Evaluation Error: Error while evaluating a Function Call, 'upcase' parameter 'arg' expects a value of type Numeric, String, Array, or Hash, got Undef (file: ~/Projects/foreman-installer/_build/modules/foreman_proxy/manifests/params.pp, line: 278, column: 29) on node prometheus

$ sed -n '277,278p;279q' ~/Projects/foreman-installer/_build/modules/foreman_proxy/manifests/params.pp
  $dns_zone               = $facts['networking']['domain']
  $dns_realm              = upcase($dns_zone)

$ facter networking.domain

$
```

I'm not sure how exactly facter determines this fact but it's not too surprising since I previously only assigned a shortname to this machine. I will get that sorted and try again.

ekohl: Yes, that's actually a bug that we should fix. Various modules we maintain don't gracefully fall back if there's no domain. Even if variables end up being unused (like realm being turned off) they hard fail. Code like this is actually what we should be used.

https://github.com/theforeman/puppet-dhcp/blob/5ba9415784e47c4e8a65f51a679d721568e82b3b/manifests/params.pp#L5-L9

Then when we know the variable is used, we should have checks.

hooks/pre/20-certs_update.rb
ehelms: Are you asking if it makes sense to elevate this check to top-level stand-alone check of the certificates on every run?

ekohl: Yes I am.

ehelms:
Given this and a few other PRs depend on the logic, I implemented the `module_present?` concept --https://github.com/theforeman/kafo/pull/262

hooks/boot/20-certs_update.rb
ehelms: This can now use `module_present?` helper

hooks/boot/20-certs_update.rb
```diff
@@ -0,0 +1,36 @@
+# Add options around regenerating certificates
+if module_present?('katello')
```
ekohl: Should this be the `certs` module? Previously it was also applied to the fpc scenario, now only katello.
```suggestion
if module_present?('certs')
```

hooks/pre/20-certs_update.rb
```diff
@@ -0,0 +1,66 @@
...
+    all_cert_names.each do |cert_name|
+      mark_for_update(cert_name, hostname)
+    end
+  end
+  
+  if app_value('certs_update_server_ca') || app_value('certs_reset')
+    mark_for_update('katello-server-ca')
+  end
+  
+  if !app_value('certs_skip_check') &&
```
ekohl: While we're at it, mind moving this to `pre_commit`? In `pre` it already committed the changes, but we should never commit anything invalid.

hooks/pre/20-certs_update.rb
```diff
@@ -0,0 +1,66 @@
...
+    unless $CHILD_STATUS.success?
+      fail_and_exit("Command '#{check_cmd}' exited with #{$CHILD_STATUS.exitstatus}:\n #{output}", 101)
+    end
+  end
+  
+  if app_value('certs_reset') && !app_value(:noop)
+    param('certs', 'server_cert').unset_value
+    param('certs', 'server_key').unset_value
+    param('certs', 'server_ca_cert').unset_value
+  end
```
ekohl: I also wonder how this worked. In `pre` all parameter values are stored in the answers file. Here it then resets them. Doesn't sound like that will persist between runs.

hooks/pre/20-certs_update.rb
```diff
@@ -0,0 +1,66 @@
...
+  
+  if !app_value('certs_skip_check') &&
+     cert_file.to_s != "" &&
+     (app_value('certs_update_server_ca') || app_value('certs_update_server'))
+    check_cmd = %(#{CHECK_SCRIPT} -c "#{cert_file}" -k "#{key_file}" -b "#{ca_file}")
+    output = `#{check_cmd} 2>&1`
+    unless $CHILD_STATUS.success?
+      fail_and_exit("Command '#{check_cmd}' exited with #{$CHILD_STATUS.exitstatus}:\n #{output}", 101)
+    end
```
ekohl: We have `execute` wrappers for this now. Should we use those instead?

hooks/pre/20-certs_update.rb
```diff
@@ -0,0 +1,66 @@
+require 'fileutils'
+require 'English'
+
+if module_enabled?('katello')
```
ekohl: Also, needs to work on fpc
```suggestion
if module_enabled?('certs')
```

hooks/pre/20-certs_update.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+  end
+
+  ca_file   = param('certs', 'server_ca_cert').value
+  cert_file = param('certs', 'server_cert').value
+  key_file  = param('certs', 'server_key').value
+  
+  if app_value('certs_update_server_ca') && !module_enabled?('katello')
+    fail_and_exit("--certs-update-server-ca needs to be used with katello", 101)
+  end
```
ekohl: Should also be in a `pre_commit` hook.

ekohl (changes requested):

hooks/pre/20-certs_update.rb
wbclark: I tested using the non-qualified path and it and it seems to work just fine

hooks/pre/20-certs_update.rb
wbclark: My understanding is that `--certs-reset` is intended to restore to the default paths -- it's a way of starting over.

The user would first run the installer with `--certs-reset` to restore to the katello defaults. Then if defaults aren't their end goal and they need to supply new certificates they would run the installer again and supply the new answers.

Do you think it should work differently?

hooks/pre/20-certs_update.rb
wbclark:
```suggestion
    check_cmd = %(katello-certs-check -c "#{cert_file}" -k "#{key_file}" -b "#{ca_file}")
    unless execute(check_cmd)
      fail_and_exit("Command '#{check_cmd}' failed, exiting.", 101)
    end
```

I think we could so something like this. `execute(check_cmd)` would still log and print output including any errors but I don't see how we could then pass any exit code or error output to `fail_and_exit` -- which I guess is not necessary at that point because `execute` is printing and logging it already?

hooks/pre/20-certs_update.rb
ekohl: Reading Kafo, it suggests that you are not guaranteed that it will commit the changes to disk this way. From the [Kafo Hooks docs](https://github.com/theforeman/kafo#hooks):
> pre - just before puppet is executed to converge system, after parameter values are stored in the answer file

My interpretation is that you should not change any parameters in the `pre` stage, but really before that.

From reading those same docs, `pre_values` makes the most sense to me. This allows you to use `foreman-installer --certs-reset --interactive` which currently is not possible.

Another argument is that this now doesn't work together with the `pre_commit` hook this change has. You run `katello-certs-check` on some values and then reset the values.

Now this is only from reading docs so some real tests would be good to see if the theory matches reality :)

hooks/pre_commit/20-certs_update.rb
```diff
@@ -0,0 +1,18 @@
+if module_enabled?('certs')
+  if app_value('certs_update_server_ca') && !module_enabled?('katello')
+    fail_and_exit("--certs-update-server-ca needs to be used with katello", 101)
+  end
+
+  ca_file   = param('certs', 'server_ca_cert').value
+  cert_file = param('certs', 'server_cert').value
+  key_file  = param('certs', 'server_key').value
+
+  if !app_value('certs_skip_check') && cert_file.to_s != "" &&
```
ekohl:
```suggestion
  if !app_value('certs_skip_check') && !cert_file.to_s.empty? &&
```

hooks/pre_commit/20-certs_update.rb
```diff
@@ -0,0 +1,18 @@
...
+     (app_value('certs_update_server_ca') || app_value('certs_update_server'))
```
ekohl: I'd argue we can drop this. Our certs check should just work. This also means that if we tighten the checks to catch more problems, users will be notified about this, if they don't use `--certs-update-server`.

hooks/pre_commit/20-certs_update.rb
```diff
@@ -0,0 +1,18 @@
...
+    check_cmd = %(katello-certs-check -c "#{cert_file}" -k "#{key_file}" -b "#{ca_file}")
+    output = `#{check_cmd} 2>&1`
+    unless $CHILD_STATUS.success?
```
ekohl: This line is why there was a `require 'English'` in the other file. However, I'd suggest using `execute` from our hooks which has a lot more functionality built into it for running commands and handling failures.

hooks/pre/20-certs_update.rb
```diff
@@ -0,0 +1,47 @@
...
+    end.uniq
+
+    all_cert_names.each do |cert_name|
+      mark_for_update(cert_name, hostname)
+    end
+  end
+
+  if app_value('certs_update_server_ca') || app_value('certs_reset')
+    mark_for_update('katello-server-ca')
+  end
```
ekohl: Does it make sense to gather an array on all of this? We now have `all_cert_names` but really we can gather them all. Explicitly gathering the paths would make `mark_for_update` redundant. Because I wanted to know how it looked, I took a stab at it:
```ruby
require 'fileutils'

if module_enabled?('certs')
  if param('foreman_proxy_certs', 'foreman_proxy_fqdn')
    hostname = param('foreman_proxy_certs', 'foreman_proxy_fqdn').value
  else
    hostname = param('certs', 'node_fqdn').value
  end

  SSL_BUILD_DIR = param('certs', 'ssl_build_dir').value
  HOST_BUILD_DIR = File.join(SSL_BUILD_DIR, hostname)

  certs_to_update = []

  if app_value('certs_update_server')
    certs_to_update << File.join(HOST_BUILD_DIR, "#{hostname}-apache")
    certs_to_update << File.join(HOST_BUILD_DIR, "#{hostname}-foreman-proxy")
  end
   
  if app_value('certs_update_all') || app_value('certs_update_default_ca') || app_value('certs_reset')
    certs_to_update += Dir.glob(File.join(HOST_BUILD_DIR, '*.noarch.rpm')).map do |rpm|
      rpm.sub(/-1\.0-\d+\.noarch\.rpm/, '')
    end
  end

  if app_value('certs_update_server_ca') || app_value('certs_reset')
    certs_to_update << File.join(SSL_BUILD_DIR, 'katello-server-ca')
  end

  certs_to_update.uniq.each do |path|
    if app_value(:noop)
      puts "Marking certificate #{path} for update (noop)"
    else
      puts "Marking certificate #{path} for update"
      FileUtils.touch("#{path}.update")
    end
  end
end
```

I think it looks and better shows the intention of this hook. Note I left out resetting the params since I believe that should be in a different hook (https://github.com/theforeman/foreman-installer/pull/514/files#r476563297).

ekohl (changes requested):
Note that `katello_certs/hooks/pre/20-certs_update.rb` is a symlink to the current certs update. That will need an update too.

ehelms:
Given this hook has worked by all accounts, it would be nice if this could be a **move** and then a **refactor** rather than combining the two. That will make tracking down issues easier.

hooks/pre_commit/20-certs_update.rb
```diff
@@ -0,0 +1,18 @@
+if module_enabled?('certs')
+  if app_value('certs_update_server_ca') && !module_enabled?('katello')
```
ekohl: I think app values are symbols
```suggestion
  if app_value(:certs_update_server_ca) && !module_enabled?('katello')
```

ehelms: Strings work, if you look at the code that was in the hooks that got moved.

ekohl: https://github.com/theforeman/kafo/blob/b7815c3af58c1fc42b6d561e16a2b6f086aad4c7/lib/kafo/hook_context.rb#L39-L44 suggests it does work, but is internally changed back to a symbol. However, this is the only file that actually uses strings. All others properly use symbols. IMHO we should always use symbols because it's more efficient and consistent.

ehelms: Would you mind filing a bug against Kafo to error if it's not a symbol? If we believe symbols are a better type, let's enforce it to prevent ambiguity.

wbclark: Opened https://projects.theforeman.org/issues/30781 and https://github.com/theforeman/kafo/pull/275

wbclark: Updated this PR to always use Symbol, to be consistent with our other usage in the Hooks.

hooks/pre/20-certs_update.rb
wbclark: Indeed, so I tested this on Katello 3.16 with the below approach:

1. Made a backup of my answers file, then edited answers to include:

```yaml
certs:
  server_cert: "/root/no-such-cert"
  server_key: "/root/no-such-key"
  server_ca_cert: "/root/no-such-ca-cert"
```

2. For now I commented lines related to required CPU/memory check since I am running these tests on a Forklifted box.

3. `# foreman-installer --certs-reset`

Then I checked the answers file and indeed the values I set in step 1 were persisted. I agree this doesn't seem correct. I'll experiment with moving it to another stage as suggested. Looking at some other examples we may also need `kafo.config.configure_application`

ekohl: I think the correct approach would be to move `unset_value` to `pre_commit` as I suggested. `pre` means `post_commit` and `pre_puppet_apply`. Using `kafo.config.configure_application` would mean saving things to disk again, which is unexpected.

hooks/pre_commit/20-certs_update.rb
```diff
@@ -0,0 +1,14 @@
...
+  if app_value(:certs_update_server_ca) && !module_enabled?('katello')
+    fail_and_exit("--certs-update-server-ca needs to be used with katello", 101)
+  end
+
+  ca_file   = param('certs', 'server_ca_cert').value
+  cert_file = param('certs', 'server_cert').value
+  key_file  = param('certs', 'server_key').value
+
+  if !app_value(:certs_skip_check) && !cert_file.to_s.empty? &&
+     (app_value(:certs_update_server_ca) || app_value(:certs_update_server))
```
ekohl, line 17: Not needed in this PR, but I'd still consider always running `katello-certs-check` if there are custom certs.

wbclark: What is the best way to detect custom certs?

I suppose `!param('certs', 'server_cert').nil? || !param('certs', 'server_key).nil? || !param('certs', 'server_ca_cert').nil?`

ekohl: It already has `cert_file.to_s.empty?`, which should suffice. After this is merged, I'd open a new PR that effectively drops this line.

wbclark:
@ekohl I fixed the symlinks and tests are now green. I also moved resetting certs answers to pre_commit. do you have any other feedback on this one?

btw, I also opened #579 to run katello-certs-check whenever non-default certs answers are provided.

hooks/pre_commit/20-certs_update.rb
```diff
@@ -0,0 +1,20 @@
+if module_enabled?('certs')
+  if app_value(:certs_update_server_ca) && !module_enabled?('katello')
+    fail_and_exit("--certs-update-server-ca needs to be used with katello", 101)
+  end
+
+  if app_value(:certs_reset) && !app_value(:noop)
```
ekohl: I think it's safe to unset in noop. In fact, if you don't you'll get different results between noop and actual runs.
```suggestion
  if app_value(:certs_reset)
```

In a noop run the answers are not saved, so that's not a problem.

wbclark: Good point... updated

katello_certs/hooks/boot/20-certs_update.rb
```diff
@@ -1 +1 @@
-../../../katello/hooks/boot/20-certs_update.rb
\ No newline at end of file
+../../../hooks/boot/20-certs_update.rb
```
ekohl, line 1: I always forget these and what they're exactly supposed to do. It would be great to have some integration tests to make sure the functionality actually does what it's supposed to.

wbclark: I'm working on https://github.com/theforeman/forklift/pull/1208 as a first step towards integration testing of installer PRs.

Could we go ahead and merge this, and add tests in a future PR?

ehelms: Yes, tests can come after. Please file a Redmine issue to not lose track of the need.

ekohl: I agree they can come later, but I'm wonder how much we actually verified this continues to work. With the rest of the code I'm decently familiar with how it should work but I'm not that familiar with `foreman-proxy-certs-generate`.

ehelms (approved):
Tested locally with an install test

PR #515, opened 2020-06-03
Fixes #30078 Add default_hostgroup plugin for the Katello scenario

apatelKmd:
Based on the recommendation provided in the support communication
https://community.theforeman.org/t/default-host-group-plugin-config-file-through-the-installer/18797/2

### DIFF
--- a/config/foreman.migrations/20200602080100_add_default_hostgroup.rb
+++ b/config/foreman.migrations/20200602080100_add_default_hostgroup.rb
@@ -1,0 +1,1 @@
+answers['foreman::plugin::default_hostgroup'] ||= false
--- a/config/katello.migrations/20200602080100-add-default_hostgroup.rb
+++ b/config/katello.migrations/20200602080100-add-default_hostgroup.rb
@@ -1,0 +1,1 @@
+answers['foreman::plugin::default_hostgroup'] ||= false
--- a/spec/fixtures/pulpcore-migration/katello-answers-after.yaml
+++ b/spec/fixtures/pulpcore-migration/katello-answers-after.yaml
@@ -11,6 +11,7 @@ foreman::compute::ovirt: false
 foreman::compute::rackspace: false
 foreman::compute::vmware: false
 foreman::plugin::azure: false
+foreman::plugin::default_hostgroup: false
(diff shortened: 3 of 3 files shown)


bin/create-migration
```diff
@@ -16,8 +16,8 @@ directories.each do |directory|
     exit 2
   end
 
-  if File.basename(directory) == 'foreman.migrations'
-    # Used by the foreman scenario
+  if ['foreman.migrations', 'katello.migrations'].include?(File.basename(directory))
```
ehelms: I don't believe this change is needed. Katello and Foreman scenarios currently use different timestamp formats for their migrations and hence why this code exists as it does (see the else block).

ehelms:
Could you please open a Redmine issue and attach it in the commit message?

bin/create-migration
apatelKmd: Remvoed

apatelKmd:
I have added the Redmine ticket number in the title https://projects.theforeman.org/issues/30078

ehelms:
@apatelKmd Almost there! Can you edit the commit message to be the same as the title? That way when it goes into the git log it will have the right info to link it back to the Redmine issue.

apatelKmd:
@ehelms Updated the commit message.

ehelms:
@wbclark added some new tests here, could you help @apatelKmd with what changes are needed with this PR and the new tests?

ekohl:
This whole change is not needed since it's already present:
https://github.com/theforeman/foreman-installer/blob/2cfd0523d41a85c024a7db17886b23c2bacfd2be/config/foreman-answers.yaml#L33
https://github.com/theforeman/foreman-installer/blob/2cfd0523d41a85c024a7db17886b23c2bacfd2be/config/katello-answers.yaml#L50

Looks like this was done prior to the installer merge so somewhere before 1.21 and my instructions on Discourse were incorrect.

https://github.com/theforeman/puppet-foreman/blob/master/manifests/plugin/default_hostgroup.pp would need to be modified to manage the config file. https://github.com/theforeman/puppet-foreman/pull/601 is a similar PR that could serve as inspiration.

My apologies for misleading here.

PR #516, opened 2020-06-04
Refs #30023 - don't use pulp3 for rpm content on upgrades

wbclark:


wbclark:
I think a possible problematic edge case here would be if the user previously ran this migration in an upgrade prior to this commit.

If I understand correctly then in that case, `answers['katello']` would already be a hash so the logic which should set `answers['katello']['use_pulp_2_for_yum'] = true` would not get executed.

ekohl (changes requested):
We never modify migrations (except bug fixes / clean up if really needed) exactly for the reason you mention: once it ran, it'll never be executed again.

This does mean it's a bit tricky to correctly detect the upgrade now. A common way to solve this is to specify the value in the answers file. That's always there on new upgrades but unused in upgrades.

I'm not sure why we didn't do this previously.

config/katello.migrations/200123161606-enable-pulpcore.rb
```diff
@@ -7,6 +7,7 @@ def upgrade
     answers['katello'] = {} unless answers['katello'].is_a?(Hash)
     answers['katello']['use_pulp_2_for_file'] = true
     answers['katello']['use_pulp_2_for_docker'] = true
+    answers['katello']['use_pulp_2_for_yum'] = true
```
evgeni: I think this should be a new migration.

wbclark:
I've taken the suggested approach of changing the param in katello-answers.yaml so that new installs will use pulp3 for rpm content by default, while upgrades will fall back on the module default of using pulp2.

Then a migration will be necessary once it's time for upgrades to also use pulp3 for rpm content.

config/katello-answers.yaml
```diff
@@ -91,7 +91,8 @@ foreman_proxy::plugin::pulp:
 foreman_proxy::plugin::remote_execution::ssh: false
 foreman_proxy::plugin::salt: false
 foreman_proxy_content: true
-katello: true
+katello:
+  user_pulp_2_for_yum: false
```
ehelms: I'm a bit confused, this is already the default value. The goal is new installs should use Pulp 3 for yum content and the module defaults to this (https://github.com/theforeman/puppet-katello/pull/347/files#diff-481a583a7371c0062cb0fcb5c039626cR22). The migration is needed to ensure upgrades continue to use Pulp 2 for yum content.

ekohl: Note the typo as well (which will be silently ignored):
```suggestion
  use_pulp_2_for_yum: false
```

wbclark: Ah, for some reason I convinced myself that the module would default to true and didn't double check.

OK, I'll re-write it as a separate migration this time, keeping the change in answers to distinguish between new installs and upgrades.

wbclark:
I've updated this with my new approach. A summary of changes:

1. changed the answers file to explicitly state the module default, in order to make it easier to detect new installs vs upgrades
2. added the migration to override the default for upgrades, so that pulp2 will still be used for RPM content
3. updated the fixtures of the test of the previous migration, so that the test will now test both migrations
4. created a new test that tests only the new migration, assuming the previous one has already run

config/katello.migrations/200605154320-dont-use-pulpcore-rpm-on-upgrades.rb
```diff
@@ -0,0 +1,9 @@
+PUPPET_CLASS = 'katello'.freeze
+PARAM = 'use_pulp_2_for_yum'.freeze
```
ehelms: I find generic constants make the code hard to read, as I have to do a lot of mental translating to review. I'm OK with constants but prefer, for understandability, they just be a reflection of the thing they represent, e.g.

```
KATELLO = 'katello'.freeze
USE_PULP_2_FOR_YUM = 'use_pulp_2_for_yum'.freeze
```

config/katello.migrations/200605154320-dont-use-pulpcore-rpm-on-upgrades.rb
```diff
@@ -0,0 +1,9 @@
...
+
+def upgrade
+  answers[PUPPET_CLASS] = {} unless answers[PUPPET_CLASS].is_a?(Hash)
+  answers[PUPPET_CLASS][PARAM] = true
+end
+
+upgrade if answers[PUPPET_CLASS] && answers[PUPPET_CLASS][PARAM].nil?
```
ehelms: Is it that the value of that param would be nil or that the param exists when the migrations are executed? I can't seem to definitively recall if migrations take into account params from modules at the time they are executed or just act on the answers file as it exists without any view of params from the current set of modules.

wbclark: Based on:

 https://github.com/theforeman/kafo/blob/master/lib/kafo/configuration.rb#L48
https://github.com/theforeman/kafo/blob/master/lib/kafo/configuration.rb#L296-L298
https://github.com/theforeman/kafo/blob/master/lib/kafo/configuration.rb#L302

It looks like answers is just defined as the logical content of the parsed answers file, at the time that migrations are run.

So I think my implementation works, but we could be more confident if the tests would actually run, which they haven't. I'll try closing and re-opening the PR to see if that triggers it.

wbclark:
I am at a loss. I do not understand why tests run on Travis every once in a while but mostly do not.

ehelms (approved):

tbrisker:
2.1 - 9a60c01

PR #518, opened 2020-06-11
Minimal Foreman scenario

ehelms:
This is largely for testing purposes to help find the knobs that need to be tweaked to have a minimal Foreman installation.

### DIFF
--- a/config/foreman-answers.yaml
+++ b/config/foreman-answers.yaml
@@ -11,6 +11,8 @@
 ---
 foreman:
   configure_epel_repo: false
+
+## CLI ##
 foreman::cli: true
 foreman::cli::ansible: false
 foreman::cli::azure: false
@@ -20,6 +22,8 @@ foreman::cli::openscap: false
 foreman::cli::remote_execution: false
 foreman::cli::tasks: false
 foreman::cli::templates: false
+
+## Compute Resources ##
 foreman::compute::ec2: false
 foreman::compute::gce: false
 foreman::compute::libvirt: false
@@ -27,6 +31,8 @@ foreman::compute::openstack: false
 foreman::compute::ovirt: false
 foreman::compute::rackspace: false
 foreman::compute::vmware: false
+
+## Foreman Plugins ##
 foreman::plugin::ansible: false
 foreman::plugin::azure: false
 foreman::plugin::bootdisk: false
@@ -53,7 +59,12 @@ foreman::plugin::setup: false
 foreman::plugin::snapshot_management: false
 foreman::plugin::tasks: false
 foreman::plugin::templates: false
-foreman_proxy: {}
+
+## Smart Proxy and Plugins ##
+foreman_proxy:
+  puppet: false
+  puppetca: false
+  manage_puppet_group: false
 foreman_proxy::plugin::ansible: false
 foreman_proxy::plugin::chef: false
 foreman_proxy::plugin::dhcp::infoblox: false
@@ -68,5 +79,6 @@ foreman_proxy::plugin::openscap: false
 foreman_proxy::plugin::pulp: false
 foreman_proxy::plugin::remote_execution::ssh: false
 foreman_proxy::plugin::salt: false
-puppet:
-  server: true
+
+## Integrated Services ##
+puppet: false
--- a/config/foreman-answers.yaml
+++ b/config/foreman-answers.yaml
@@ -11,6 +11,15 @@
 ---
 foreman:
   configure_epel_repo: false
+  server_ssl_crl: ""
+  server_ssl_ca: "/etc/foreman-pki/certs/ca/ca.crt"
+  server_ssl_chain: "/etc/foreman-pki/certs/ca/ca.crt"
+  server_ssl_cert: "/etc/foreman-pki/certs/apache/apache.crt"
+  server_ssl_key: "/etc/foreman-pki/certs/apache/apache.key"
+  client_ssl_ca: "/etc/foreman-pki/certs/ca/ca.crt"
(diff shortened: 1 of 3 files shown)


ekohl:
You may also want to set tftp to false until https://github.com/theforeman/puppet-foreman_proxy/pull/602 is merged.

config/foreman-answers.yaml
```diff
@@ -11,6 +11,17 @@
 ---
 foreman:
   configure_epel_repo: false
+  server_ssl_crl: ""
+  server_ssl_ca: "/etc/foreman-pki/certs/ca/ca.crt"
+  server_ssl_chain: "/etc/foreman-pki/certs/ca/ca.crt"
```
ekohl, line 16: I don't like this scheme naming. It implies a single CA but we know that's not true. I'd suggest at least symlinks: `/etc/foreman-pki/certs/apache/apache-ca.pem`

config/foreman-answers.yaml
```diff
@@ -11,6 +11,17 @@
...
+  server_ssl_cert: "/etc/foreman-pki/certs/apache/apache.crt"
```
ekohl, line 17: I'd consider using the pem extension since they're pem encoded files.
```suggestion
  server_ssl_cert: "/etc/foreman-pki/certs/apache/apache-crt.pem"
```

config/foreman-answers.yaml
```diff
@@ -53,7 +68,16 @@ foreman::plugin::setup: false
...
+
+## Smart Proxy and Plugins ##
+foreman_proxy:
+  puppet: false
+  puppetca: false
+  manage_puppet_group: false
+  ssl_ca: "/etc/foreman-pki/certs/ca/ca.crt"
+  ssl_cert: "/etc/foreman-pki/certs/smart-proxy/smart-proxy.crt"
+  ssl_key: "/etc/foreman-pki/certs/smart-proxy/smart-proxy.key"
+  ssldir: "/etc/foreman-pki/certs"
```
ekohl: This is wrong. ssldir is the Puppet CA SSL directory and only used to configure the Puppet CA cert provider. That's only used with Puppet 5, Puppet 6 uses Puppet's REST API.

You can argue the parameter has a bad name and you'd be right.

config/foreman-answers.yaml
ehelms: There is still "custom certificates" to come support wise which is why this is currently this way. It is built around a single CA concept as the intro use case. I'm still contemplating how to handle custom certificates and at what level. I think it would be nice if we could get away with the simple case of supplying a single CA rather than customizing only some certificates but feedback so far is that is not feasible for all use cases.

config/foreman-answers.yaml
ehelms: I've tried to research this and look at lots of examples. I landed on that in most cases I could find using the `.crt` and `.key` extensions appears to be fairly common and what folks care more about than the format especially given PEM is pretty standard.

config/foreman-answers.yaml
ehelms: You did a great job arguing the bad name for me :)

config/foreman-answers.yaml
ekohl: If we already use bundles of 3 (cert, key, ca) then we can easily support that in the future. If you symlink to the common CA, you still only have a single file. Perhaps we should also have a 4th (chain) for when we have client auth.

I'd also recommend looking at certbot (the Let's Encrypt client). It creates a structure:
* cert.pem (certificate)
* chain.pem (the CA chain, including intermediates)
* fullchain.pem (the full chain - from CA to cert)
* privkey.pem (private key)

It does so for every certificate. Perhaps we should follow the same where the directory indicates the purpose and the file names within the directory are always the same. That way you could create a bundle `private-ca` and a bundle `public-ca`. Then just symlink `apache` to `public-ca` and you're done. However, perhaps file permissions will get in the way.

config/foreman-answers.yaml
ekohl: I'm wondering what we should do. I hope to drop Puppet 5 support when it goes EOL in November. The REST API implementation also works with Puppet 5 but you miss some fields that are only shown in the UI so I think dropping that provider is fine. Then we can drop the parameter.

Given November is Foreman 2.3 or 2.4, I'm not sure renaming it for 2 or 3 releases is worth it.

config/foreman-answers.yaml
```diff
@@ -11,6 +11,17 @@
 ---
 foreman:
   configure_epel_repo: false
+  server_ssl_crl: ""
+  server_ssl_ca: "/etc/foreman-pki/certs/ca/ca.crt"
```
ekohl, line 15: Just to be explicit about this: I hope we're not ending up with passing this via answers and this is just so you can iterate quickly.

ehelms: This is to help with iterating. Is this asking that in the end these paths are the defaults in the puppet modules?

ekohl: Yes

ehelms: Definitely, assuming we arrived at an agreed upon file layout I planned to do that. You can see some of my testing code that did this here: https://github.com/theforeman/puppet-foreman/pull/856/files

ehelms:
Tracking current test scenario that results in a successful install for Foreman:

```
foreman:
  box: centos7-foreman-nightly
  ansible:
    variables:
      foreman_repositories_environment: staging
      katello_repositories_environment: staging
      koji_task_ids:
        - 356320
      foreman_installer_module_prs:
        - theforeman/foreman/856
        - theforeman/foreman_proxy/599
        - theforeman/katello/351
```

PR #521, opened 2020-06-16
Fixes #30136: Do not proxy isos or yum on foreman proxy content

ehelms:


wbclark (changes requested):
@ehelms This change in answers would only affect new installations, correct? Therefore for users that upgraded there should also be a migration to disable the proxying of isos and rpm content.

ehelms:
Yessir, you are right. Helps if you git add new files when creating a PR :)

jlsherrill:
@ehelms looks like some rubocop failures

ehelms:
Addressed rubocop @jlsherrill @wbclark

jlsherrill:
LGTM @wbclark ?

config/foreman-proxy-content.migrations/200616155948-disable-pulp-3-proxying.rb
```diff
@@ -0,0 +1,15 @@
...
+PROXY_ISOS = 'proxy_pulp_isos_to_pulpcore'.freeze
+
+def upgrade
+  answers[FOREMAN_PROXY_CONTENT] = {} unless answers[FOREMAN_PROXY_CONTENT].is_a?(Hash)
+  answers[FOREMAN_PROXY_CONTENT][PROXY_YUM] = false
+  answers[FOREMAN_PROXY_CONTENT][PROXY_ISOS] = false
+end
+
+if answers[FOREMAN_PROXY_CONTENT] &&
+   answers[FOREMAN_PROXY_CONTENT][PROXY_YUM].nil? &&
```
wbclark, line 13: I used this pattern previously when I was confident that answers[FPC] should already be a hash, but that wasn't the case for answers in this scenario previously, opening the possibility of a NoMethodError:

```
irb(main):012:0> answers
=> {"FPC"=>true}
irb(main):013:0> if answers['FPC']
irb(main):014:1>   puts "hello"
irb(main):015:1> end
hello
=> nil
irb(main):016:0> if answers['FPC'] && answers['FPC']['PROXY'].nil?
irb(main):017:1>   puts "hello"
irb(main):018:1> end
Traceback (most recent call last):
        2: from lib/tasks/console.rake:5:in `block in <top (required)>'
        1: from (irb):16
NoMethodError (undefined method `[]' for true:TrueClass)
irb(main):019:0> 
```

wbclark: Hmmm, my usage may have been incorrect as well. I'm investigating it a bit further.

config/foreman-proxy-content.migrations/200616155948-disable-pulp-3-proxying.rb
```diff
@@ -0,0 +1,15 @@
...
+   answers[FOREMAN_PROXY_CONTENT][PROXY_ISOS].nil?
```
wbclark:
```suggestion
if answers[FOREMAN_PROXY_CONTENT] == true ||
   (answers[FOREMAN_PROXY_CONTENT].is_a?(Hash) &&
   answers[FOREMAN_PROXY_CONTENT][PROXY_YUM].nil? &&
   answers[FOREMAN_PROXY_CONTENT][PROXY_ISOS].nil?)
```
I think this works in all cases.

wbclark (approved):
LGTM. Thanks @ehelms !

tbrisker:
2.1 - 1adc0d7

PR #523, opened 2020-06-18
Refs #30136 - Ensure on FPC proxy to pulpcore is false for iso and yum

ehelms:


ehelms:
There is a slightly tricky situation where `proxy_pulp_isos_to_pulpcore` is already available and configured to true in nightly. This needs to be fixed really in 3.15. However, to fix nightly this change should ensure this situation. Also, we want these to be false no matter what on upgrading a proxy right now.

jlsherrill (approved):

tbrisker:
2.1 - 46b7f14

PR #524, opened 2020-06-23
Fixes #30193 - properly re-install puppet-agent-oauth for version-locked, SCL-ed installer

evgeni:
- move version unlock to main `pre_validations` hooks
`hooks/pre_validations/31-puppet-agent-oauth.rb` tries to re-install `puppet-agent-oauth`, but for that to work the packages need to be unlocked first. it's not sufficient to move it to `katello/hooks/pre_validations` as those are executed after *all* main `pre_validations`

 - execute rpm commands outside the SCL
otherwise Ruby finds the SCL oauth gem and does not re-install the one for the AIO Puppet Ruby


evgeni:
okay, this is not enough, as hooks/pre_validations/31-puppet-agent-oauth.rb is executed before katello/hooks/pre_validations/09-version_locking.rb :roll_eyes:

evgeni:
nope. damn scl

evgeni:
@ehelms maybe "just rebuild the rpm" would have been easier, but now we know that our code is/was broken in a very subtle way ;-)

hooks/boot/03-foreman-maintain-extensions.rb
```diff
@@ -1,6 +1,3 @@
-require 'English'
```
evgeni, line 1: the changes here are obviously not required, but it felt better to unify the code.

hooks/pre_validations/31-puppet-agent-oauth.rb
```diff
@@ -1,7 +1,7 @@
 unless app_value(:noop)
   if File.exist?('/opt/puppetlabs/puppet/bin/ruby') && execute_command("rpm -q puppet-agent-oauth", false, false)
-    unless execute_command("/opt/puppetlabs/puppet/bin/ruby -e \"require 'oauth'\"", false, false)
-      execute("yum -y reinstall puppet-agent-oauth", false, true)
+    unless execute_command_nonscl("/opt/puppetlabs/puppet/bin/ruby -e \"require 'oauth'\"", false, false)
+      execute_command_nonscl("yum -y reinstall puppet-agent-oauth", false, true)
```
ekohl, line 4: Is the SCL context important here? If so, it indicates https://github.com/theforeman/foreman-packaging/blob/024d0a3d7a2b2a3cad98f9e89a12cb458d30593f/packages/foreman/puppet-agent-oauth/puppet-agent-oauth.spec#L32-L50 isn't properly isolated.

evgeni: Yes. If I execute it in SCL context, it will happily load the SCL oauth gem.

ekohl: "nice"

ekohl (approved):

hooks/pre_validations/31-puppet-agent-oauth.rb
evgeni: I wonder if the installer should _always_ execute external commands w/o SCL context.

ekohl: Looking at what's executed, yes, I think that makes sense.

evgeni: @ehelms what do you think?

Also, shall we put that in this PR or later?

ekohl: If we decide to non-scl everything I think it makes sense to do it here. Introducing a method and removing it later feels bad.

hooks/boot/03-foreman-maintain-extensions.rb
```diff
@@ -31,11 +28,9 @@ def stop_services(services = nil)
 
   def foreman_maintain(command, exit_on_fail = false)
     command = "foreman-maintain #{command}"
-    log_and_say :debug, "Executing: #{command}"
-
-    _stdout, stderr, status = Open3.capture3(*Kafo::PuppetCommand.format_command(command))
+    status = execute_command_nonscl(command, false, true)
 
-    fail_and_exit(stderr) if exit_on_fail && !status.success?
+    exit 1 if exit_on_fail && !status.success?
```
ekohl, line 33:
```
    /usr/share/foreman-installer/hooks/boot/03-foreman-maintain-extensions.rb:34:in `foreman_maintain': undefined method `success?' for false:FalseClass (NoMethodError)
```

Status is now a boolean.

PR #525, opened 2020-06-23
Refs #29602 - Remove dropped repository parameters

ekohl:
Since theforeman-foreman 15.0.0 these parameters no longer exist and by default no repository setup is done.

### DIFF
--- a/config/foreman-answers.yaml
+++ b/config/foreman-answers.yaml
@@ -9,8 +9,7 @@
 #
 # See params.pp in each class for what options are available
 ---
-foreman:
-  configure_epel_repo: false
+foreman: {}
 foreman::cli: true
 foreman::cli::ansible: false
 foreman::cli::azure: false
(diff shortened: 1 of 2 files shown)


ehelms:
Could we get a Redmine issue for this please.

config/foreman-answers.yaml
```diff
@@ -9,8 +9,7 @@
 #
 # See params.pp in each class for what options are available
 ---
-foreman:
-  configure_epel_repo: false
+foreman: {}
```
ekohl, line 12: Note I chose a hash since our migrations aren't written cleanly. `true` would be cleaner, but this matches `foreman_proxy` below.

ekohl:
Updated to ref the original change in puppet-foreman.

evgeni:
(travis somehow forgot that I am part of this org, nice)

ehelms (approved):

PR #526, opened 2020-06-24
Fixes #30193 - status is now a boolean

evgeni:


evgeni:
@ekohl sorry for the noise

ekohl (approved):

PR #528, opened 2020-06-24
Refs #30193 - always execute commands outside the SCL context

evgeni:


ekohl (approved):

evgeni:
nightly is green again, so let's break it

PR #529, opened 2020-06-25
Fixes #30214 - properly catch the case when a command is not present

evgeni:
Debian does not have this feature

### DIFF
--- a/hooks/boot/01-kafo-hook-extensions.rb
+++ b/hooks/boot/01-kafo-hook-extensions.rb
@@ -107,7 +107,12 @@ module HookContextExtension
   def execute_command(command, do_say, do_log)
     log_and_say(:debug, "Executing: #{command}", do_say, do_log)
 
-    stdout_stderr, status = Open3.capture2e(*Kafo::PuppetCommand.format_command(command))
+    begin
+      stdout_stderr, status = Open3.capture2e(*Kafo::PuppetCommand.format_command(command))
+    rescue Errno::ENOENT
+      log_and_say(:error, "#{command} not found!", do_say, do_log)
+      return false
+    end
     stdout_stderr.lines.map(&:chomp).each do |line|
       log_and_say(:debug, line, do_say, do_log)
     end
--- a/hooks/boot/01-kafo-hook-extensions.rb
+++ b/hooks/boot/01-kafo-hook-extensions.rb
@@ -73,8 +73,12 @@ module HookContextExtension
     !File.exist?('/var/opt/rh/rh-postgresql12/lib/pgsql/data') && File.exist?('/var/lib/pgsql/data')
   end
 
+  def el?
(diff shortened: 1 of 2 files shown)


ekohl (commented):
It's odd, because the code should detect that foreman-maintain isn't present or isn't capable of locking and then just continue. Somehow that detection isn't working?

ekohl (changes requested):
Also, looks like this doesn't really relate to the Redmine issue.

evgeni:
There is no "is not present" detection right now. And Open3 fails if the binary it should call is not there:

```console
% irb
irb(main):001:0> require 'open3'
=> true
irb(main):002:0> Open3.capture2e('/usr/bin/false')
=> ["", #<Process::Status: pid 14111 exit 1>]
irb(main):003:0> Open3.capture2e('/usr/bin/no-exist')
Traceback (most recent call last):
        8: from /usr/bin/irb:23:in `<main>'
        7: from /usr/bin/irb:23:in `load'
        6: from /usr/share/gems/gems/irb-1.2.3/exe/irb:11:in `<top (required)>'
        5: from (irb):3
        4: from /usr/share/ruby/open3.rb:390:in `capture2e'
        3: from /usr/share/ruby/open3.rb:208:in `popen2e'
        2: from /usr/share/ruby/open3.rb:213:in `popen_run'
        1: from /usr/share/ruby/open3.rb:213:in `spawn'
Errno::ENOENT (No such file or directory - /usr/bin/no-exist)
```

(same happens with the old `IO.popen` code)

ekohl:
Should we catch `Errno::ENOENT` then?

evgeni:
Like this?

```patch
--- hooks/boot/01-kafo-hook-extensions.rb
+++ hooks/boot/01-kafo-hook-extensions.rb
@@ -111,7 +111,12 @@ module HookContextExtension
   def execute_command(command, do_say, do_log)
     log_and_say(:debug, "Executing: #{command}", do_say, do_log)
 
-    stdout_stderr, status = Open3.capture2e(*Kafo::PuppetCommand.format_command(command))
+    begin
+      stdout_stderr, status = Open3.capture2e(*Kafo::PuppetCommand.format_command(command))
+    rescue Errno::ENOENT
+      log_and_say(:error, "#{command} not found!", do_say, do_log)
+      return false
+    end
     stdout_stderr.lines.map(&:chomp).each do |line|
       log_and_say(:debug, line, do_say, do_log)
     end
```

ekohl:
Yes. I think that works. We shouldn't log/say something when foreman-maintain isn't present but I think that's already done by passing the correct values to `do_say` and `do_log` now.

evgeni:
@tbrisker this also needs to go to 2.1 when merged

evgeni:
Closing in favor of https://github.com/theforeman/foreman-installer/pull/532

PR #530, opened 2020-06-25
Refs #30193 - also move other locking related files to main hooks

evgeni:


evgeni:
the `post` one was renumbered to put the messages last

ekohl (changes requested):
I think the first line of the boot hook should be `if package_lock_feature?`

hooks/pre_validations/09-version_locking.rb
```diff
@@ -2,17 +2,18 @@
 cli_param = app_value(:lock_package_versions)
 custom_config_value = get_custom_config(:lock_package_versions)
 lock_versions = cli_param.nil? ? !!custom_config_value : cli_param # rubocop:disable Style/DoubleNegation
+lock_feature = package_lock_feature?
```
ekohl, line 5: This calls `foreman-maintain` which is slow. I now realize it did that before as well, but now I'm wondering if we can avoid that.

Perhaps https://github.com/theforeman/kafo/blob/68ae1ad760f4073a6fedc9ce8cfe4f6ddd1a85e8/lib/kafo/hook_context.rb#L24-L38 needs an `app_option?` as well:
```ruby
def app_option?(option)
  self.kafo.config.app.key?(option.to_sym)
end
```

Then we can wrap this entire hook in `app_option?(:lock_package_versions)` and do the same thing again in post.

evgeni: And now it calls and fails anyways *sigh*

evgeni: So the failure is expected until we have #529 merged, that makes sense.

@ekohl I agree, having `app_option?` would be cool -- but I'd rather not wait on a Kafo fix for that right now?

ekohl: We can implement that in https://github.com/theforeman/foreman-installer/blob/develop/hooks/boot/01-kafo-hook-extensions.rb. `param_value` is another example that we probably need to move over but for now lives in this repo.

evgeni:
@ekohl could you by any chance finish that up? I won't have enough time until Monday when @tbrisker wants to release 2.1 :(

ekohl:
I took a good look at it and decided https://github.com/theforeman/foreman-installer/pull/531 was a better solution.

ekohl:
Closing since #531 was merged.

PR #531, opened 2020-06-25
Refs #30193 - Correct version locking

ekohl:
6e1a07bc35bfe97927dfac6477a1faf53092c58f moved version locking from Katello to the common hooks. This was unintentional and should have been kept for Katello, at least for now.

However, it did try to fix a real issue where yum was locked while the hooks tried to make changes. However, during pre_validations it may still find an issue and roll back so no changes should be made. To solve this, the hook is moved to pre. This is more correct because at this point Kafo, has committed all the answers and is going to make changes to the system using Puppet.

This replaces https://github.com/theforeman/foreman-installer/pull/530


ekohl:
@evgeni could you have a look? I think this unblocks Debian builds.

evgeni:
Looks correct to me.

evgeni (approved):

tbrisker:
I'm guessing this should also go to 2.1-stable, right?

evgeni:
Yes, however Debian nightlies are still red and I didn't look why yet.

tbrisker:
pulled to 2.1 as 66988c8

PR #533, opened 2020-06-29
Fixes #30245 - Ensure Pulp 2 is disabled on EL8

ehelms:


config/foreman-proxy-content.migrations/200630000120-disable-pulp-2-el8.rb
```diff
@@ -0,0 +1,4 @@
+if facts[:os][:release][:major] == '8' && facts[:os][:family] == 'RedHat'
```
ekohl: It feels a bit odd to read it this way because I'm so used to first checking the family and then the specific release. Doesn't really matter, just a habit I've developed. Probably because in Puppet modules you often have logic blocks for a family and nested a block for specific releases.

config/foreman-proxy-content.migrations/200630000120-disable-pulp-2-el8.rb
```diff
@@ -0,0 +1,4 @@
...
+  answers['foreman_proxy::plugin::pulp']['enabled'] = false
```
ekohl: Let's be defensive and first check that `answers['foreman_proxy::plugin::pulp']` is a hash. Otherwise you can get nasty errors. Might also be good to handle `true`.

ekohl (commented):
Untested, but generally looks like the way to go. I would like to be defensive though because in the future we might simplify the default answer to just `true` / `false` and then the migration would crash.

I'm also wondering if it makes sense to change the default to do this check:
https://github.com/theforeman/puppet-foreman_proxy/blob/c794f80bad96a2638881f879afaf94b71b317b3a/manifests/plugin/pulp/params.pp#L4
On the one hand, that would be more correct but on the other hand, it's going away and in few cases it would actually help.

config/foreman-proxy-content.migrations/200630000120-disable-pulp-2-el8.rb
ehelms: What is the case where it is not a hash?

ekohl: In the future we may change the answer file to just `foreman_proxy::plugin::pulp: true`. Users can also have set it to `false` (though that's unlikely). It's mostly that I'm allergic to code that assumes a value and crashes with an exception on other legal values.

ehelms: Be nice if there was a way to enforce this without it being still possible for the user to set a bad option and get themselves in trouble. I liked the way the hiera heirarchy works but the answers file ultimately overrides that. Being able to block parameters would also be nice.

ekohl: I too want to be more flexible in that. I thought I had a way to override the module defaults at the scenario level but misunderstood how Kafo does things today. Still on my list to fix.

> Being able to block parameters would also be nice.

Blocking values is possible with a `pre_commit` hooks. Being able to prevent parameters from being shown as parameters is currently not possible but also on my wish list.

Combining those two (scenario level defaults + hidden parameters) would be a powerful tool in reducing parameter counts.

ekohl (approved):

ekohl:
Looks like Rubocop isn't fully happy.

PR #536, opened 2020-07-02
Fixes #29798 - Move --lock-package-versions to hooks/

ekohl:
The primary goal of this PR is to move the hook from Katello to a shared one. An accidental move in 6e1a07bc35bfe97927dfac6477a1faf53092c58f resulted on this being executed on Debian which showed some flaws. Additional improvements are made to make it robust.

This uses the package_lock_feature? helper to determine whether locking is available. It then uses the newly introduced app_option? to determine if the option is available. This means it doesn't have to execute package_lock_feature? within the pre_commit hook again which saves an exec() of an external process.

In case locking could not be performed, the preference is cleared to allow the user to proceed. This saves them from having to edit the scenario manually. Instructions are printed to restore functionality.

This replaces https://github.com/theforeman/foreman-installer/pull/511


hooks/pre_commit/09-version_locking.rb
```diff
@@ -0,0 +1,27 @@
...
+    unlock_packages
+  end
+elsif get_custom_config(:lock_package_versions)
+  # This case can happen if previously packages were locked but this is no
+  # longer supported. For example, if foreman-maintain was removed. Clearing
+  # this ensures the post hook works properly
+  store_custom_config(:lock_package_versions, nil)
+  kafo.config.configure_application
+  fail_and_exit <<~MESSAGE
+  Locking of package versions was requested but foreman-maintain couldn't be used for locking.
```
wbclark, line 31:
```suggestion
  Locking of package versions was requested but foreman-maintain could not be used for locking.
```

If I'm reading it correctly, Rubocop is actually complaining about this?

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -7,6 +7,17 @@ def param_value(mod, name)
...
 
+  # Returns whether the given app option exists
+  #
+  # This is useful when there's a conditional option that is determined during
+  # boot. Later in the actual hooks it can simply use this method to check
+  # whether the hook is relevant.
+  #
+  # TODO: move to Kafo
+  def app_option?(option)
+    self.kafo.config.app.key?(option.to_sym)
```
wbclark:
```suggestion
    kafo.config.app.key?(option.to_sym)
```

hooks/pre_commit/09-version_locking.rb
```diff
@@ -0,0 +1,27 @@
...
+  The locking preference was cleared. Rerun the installer with the same arguments to proceed.
+  If locking is desired, ensure foreman-maintain can be used and rerun with --lock-package-versions.
```
wbclark, line 33:
```suggestion
  If locking is desired, ensure foreman-maintain can be used and rerun the installer with --lock-package-versions.
```

wbclark (changes requested):
I'll go ahead and implement the `app_option?` helper in Kafo, unless there is some particular reason you want to hold off on it for now.

ekohl:
> I'll go ahead and implement the `app_option?` helper in Kafo, unless there is some particular reason you want to hold off on it for now.

No, I think that would be good to have. This was just the quickest way and I actually considered doing the Kafo change directly.

wbclark:
Hey @ekohl , heads up that `app_option?` helper is available in Kafo with https://github.com/theforeman/kafo/commit/77dbd25ba2d4aa54576cb73c9ac19e5fd50c720b

ekohl:
Rebased. Also split the hook context to localize methods.

hooks/pre_commit/09-version_locking.rb
wbclark: Re-ran tests and it seems not to be an issue now. Marking as resolved.

wbclark (approved):
LGTM. Thanks @ekohl

PR #537, opened 2020-07-02
Switch to Github Actions

ekohl:
The goal of this PR is to also run on actual CentOS 7/8, Debian 10 and Ubuntu 18.04 containers in noop mode. That should also cover the hooks which currently aren't tested until after they're merged.


ekohl:
Oops, wrong branch.

PR #538, opened 2020-07-02
Switch to Github Actions

ekohl:
The goal of this PR is to also run on actual CentOS 7/8, Debian 10 and Ubuntu 18.04 containers in noop mode. That should also cover the hooks which currently aren't tested until after they're merged.


ehelms:
I think I recall if we commit an empty file then this one will actually run and we can see the results? I like this change and that style of tackling it so we can see the results of this before we merge.

ekohl:
I'm now iterating by running it in my own fork. Will see if that actually gives the result I want before removing it from draft status.

ekohl:
The further I get with this, the more it looks like doing this in containers is really limited.

ehelms:
That's what we do today with Travis right? or are you adding more testing with the switch?

ekohl:
Yes - today we only run on Ubuntu but here I try to run the installer in noop mode on all supported platforms (CentOS 7, 8, Debian 10 & Ubuntu 18.04). However, containers are very systemd unfriendly and don't have a working DNS forward/reverse. See https://github.com/ekohl/foreman-installer/actions/runs/160558319.

It's first running the tests, builds a tarball, passes it to the containers who then run it. I'm starting to feel like we should use CentOS CI with real VMs for this.

ehelms:
@ekohl I can help here if you like, I set this up once for some testing of containers I did in Forklift.

ekohl:
I'm going to close this for now since I'm hitting limits. Later we should investigate it again using real VMs.

PR #539, opened 2020-07-09
Fixes #30349 - Handle empty data dirs in PG upgrade

ekohl:
When rh-postgresql12-postgresql-server is installed prior to running the installer, the directory exists but is empty. This breaks upgrades.

### DIFF
--- a/hooks/boot/01-kafo-hook-extensions.rb
+++ b/hooks/boot/01-kafo-hook-extensions.rb
@@ -70,7 +70,9 @@ module HookContextExtension
   end
 
   def needs_postgresql_scl_upgrade?
-    !File.exist?('/var/opt/rh/rh-postgresql12/lib/pgsql/data') && File.exist?('/var/lib/pgsql/data')
+    old_path = '/var/lib/pgsql/data'
+    new_path = '/var/opt/rh/rh-postgresql12/lib/pgsql/data'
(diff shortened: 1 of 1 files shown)


ehelms:
Do you know if the upgrade will happen properly if that directory is not empty? Or if the user has to clear it out for the scripts to work?

ekohl:
No, I don't know. But right now the upgrade doesn't happen at all if the directory exists but is empty. This can be seen in https://community.theforeman.org/t/upgrade-from-1-24-3-to-2-0-fails/19516.

ehelms:
I feel like I recall that if that directory exists the upgrade scripts will also fail. After a quick test, this will fail and why we had to go this route. Even if `/var/opt/rh/rh-postgresql12/lib/pgsql/data` is present but empty the upgrade will fail. I suppose you could check that and have the upgrade fail if `/var/opt/rh/rh-postgresql12/lib/pgsql/data` exists but is empty with a message to remove it and re-run.

For develop branch, we can likely remove this upgrade code for the Foreman 2.2 release? Or at least Foreman 2.3/3.0

ehelms:
@ekohl Thoughts here? Do you want to try to find another work around, remove the upgrade code or close this PR?

ekohl:
I'm going to close this PR. This appears to be an uncommon case and I don't have the capacity to finish this.

PR #540, opened 2020-07-15
Fixes #29799 - Move --clear-pulp-content to hooks/

wbclark:
I implemented this as separate options for Pulp2 and Pulpcore. This has at least two key advantages -- first, the user may wish to clear one but not the other. Second, when Pulp2 is later removed it makes it very easy to remove `--clear-pulp-data` without touching `--clear-pulpcore-data`.


hooks/boot/12-clear_pulp_content.rb
```diff
@@ -0,0 +1,8 @@
+if pulp_enabled?
```
ekohl: With Pulp 2 removal planned and being worked on for 4.0, wouldn't it be better to just remove it from Katello and drop it?

hooks/pre/12-clear_pulp_content.rb
```diff
@@ -0,0 +1,74 @@
...
+  config[:host] = host || 'localhost'
+  config[:port] = port || '27017'
+  config[:database] = param_value('katello', 'pulp_db_name') || 'pulp_database'
+  config[:username] = param_value('katello', 'pulp_db_username')
+  config[:password] = param_value('katello', 'pulp_db_password')
+  config[:ssl] = param_value('katello', 'pulp_db_ssl') || false
+  config[:ca_path] = param_value('katello', 'pulp_db_ca_path')
+  config[:ssl_certfile] = param_value('katello', 'pulp_db_ssl_certfile')
+  config
+end
```
wbclark: Most functions here (and similarly in `hooks/pre/10-reset_data.rb` and in `hooks/pre/13-clear_pulpcore_content.rb`) could be moved to `hooks/boot/01-kafo-hook-extensions.rb` or something new like `hooks/boot/05-data-reset-helpers.rb` to prevent duplication of code.

hooks/boot/12-clear_pulp_content.rb
wbclark: I think it could be kept until the removal is really complete? Having it split into separate files makes it trivially easy to drop.

hooks/pre/12-clear_pulp_content.rb
ekohl: Before you merged that, it was actually going to be my comment that we should just drop Pulp 2 before migrating things over.

hooks/boot/12-clear_pulp_content.rb
ekohl: We're actively working on killing Pulp 2 in all modules. Resetting and clearing is IMHO a good place to start. Katello 4.0 is the big Pulp 2 removal version and we slightly less than a month (so the end of this sprint) before branching. I'm in favor of doing the work just once: remove first, then move over what's left.

ehelms: Maybe we hedge our bets, and go this route until we've actually removed Pulp 2. This keeps us flexible but easily able to rip this out closer to release. Despite the project being `Pulp` so much naming has had to be `pulpcore` that I do think using that for the Pulp 3 hook makes sense. Provides consistency.

ehelms: I thought about it more and I think this is the right approach until we know the Katello plugin is fully ready to drop Pulp 2 given their timeline has slipped due to various bugs found in the migrations and adaptation to Pulp 3 RPM content.

hooks/pre/12-clear_pulp_content.rb
ehelms: Moving to a dedicated data helpers file in `boot` makes sense.

hooks/boot/12-clear_pulp_content.rb
ehelms: Given Katello project will be having two more releases, I think the two options is the route to go.

hooks/pre/12-clear_pulp_content.rb
ehelms: @wbclark Do you want to tackle the re-factor mentioned here in this PR or a follow up?

wbclark:
@ehelms I introduced `hooks/boot/05-data-helpers.rb`. Now we have minimal `hooks/pre/{10-reset_data.rb,12-clear_pulp_content.rb,13-clear_pulpcore_content.rb}` which call the reusable helpers.

hooks/boot/05-data-helpers.rb
```diff
@@ -0,0 +1,126 @@
...
+  def empty_mongo
+    logger.info "Dropping Pulp database!"
+  
+    mongo_config = load_mongo_config
+    if remote_host?(mongo_config[:host])
+      empty_remote_mongo(mongo_config)
+    else
+      execute(
+        [
+          'systemctl start rh-mongodb34-mongod',
```
ekohl: Should we use `start_services(['rh-mongodb34-mongod'])` here?

hooks/boot/05-data-helpers.rb
```diff
@@ -0,0 +1,126 @@
...
+      port: param_value(module_name, "#{param_prefix}port") || 5432,
+      database: param_value(module_name, "#{param_prefix}#{db_param}") || db,
+      username: param_value(module_name, "#{param_prefix}#{user_param}"),
+      password: param_value(module_name, "#{param_prefix}password"),
+    }
+  end
+  
+  def pg_command_base(config, command, args)
+    port = "-p #{config[:port]}" if config[:port]
+    "PGPASSWORD='#{config[:password]}' #{command} -U #{config[:username]} -h #{config[:host]} #{port} #{args}"
```
ekohl: The downside of using it like this is that you can see the password in `ps aux`. It would be much better to pass an environment variable via native Ruby methods rather than relying on shell.

hooks/boot/05-data-helpers.rb
```diff
@@ -0,0 +1,126 @@
...
+      host: param_value(module_name, "#{param_prefix}host") || 'localhost',
+      port: param_value(module_name, "#{param_prefix}port") || 5432,
+      database: param_value(module_name, "#{param_prefix}#{db_param}") || db,
+      username: param_value(module_name, "#{param_prefix}#{user_param}"),
+      password: param_value(module_name, "#{param_prefix}password"),
+    }
+  end
+  
+  def pg_command_base(config, command, args)
+    port = "-p #{config[:port]}" if config[:port]
```
ekohl: This doesn't make a lot of sense since at line 29 you already guarantee there is a port set.

hooks/boot/05-data-helpers.rb
```diff
@@ -0,0 +1,126 @@
...
+    host, port = seed.split(':') if seed
+    config[:host] = host || 'localhost'
+    config[:port] = port || '27017'
+    config[:database] = param_value('katello', 'pulp_db_name') || 'pulp_database'
+    config[:username] = param_value('katello', 'pulp_db_username')
+    config[:password] = param_value('katello', 'pulp_db_password')
+    config[:ssl] = param_value('katello', 'pulp_db_ssl') || false
+    config[:ca_path] = param_value('katello', 'pulp_db_ca_path')
+    config[:ssl_certfile] = param_value('katello', 'pulp_db_ssl_certfile')
+    config
```
ekohl: I prefer a single definition:
```ruby
config = {
  host: host || 'localhost',
  port: port || 27017
  # ...
}
```

ekohl (changes requested):

hooks/boot/05-data-helpers.rb
wbclark: Checking https://github.com/theforeman/foreman-installer/blob/develop/hooks/boot/01-kafo-hook-extensions.rb#L119 and https://github.com/theforeman/kafo/blob/master/lib/kafo/puppet_command.rb#L49 , I think it makes sense to solve this in a separate PR.

hooks/boot/05-data-helpers.rb
wbclark: I suppose if we have a function defined for this exact purpose then we should use it. I've made the change.

hooks/boot/13-clear_pulpcore_content.rb
```diff
@@ -0,0 +1,8 @@
+if pulpcore_enabled?
```
ehelms: Being in boot phase, this will need to use the new `module_present?` helper (https://github.com/theforeman/kafo/pull/262) and likely introduce a `pulpcore_present?` to also help draw the distinction.

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -73,10 +73,18 @@ def pulp_enabled?
     module_enabled?('foreman_proxy_plugin_pulp') && (param_value('foreman_proxy_plugin_pulp', 'pulpnode_enabled') || param_value('foreman_proxy_plugin_pulp', 'enabled'))
   end
 
+  def pulp_present?
```
wbclark, line 84: It's actually identical to https://github.com/theforeman/foreman-installer/pull/540/files#diff-c6fdf0194de166ebab0dde67cd968490R84 but in the interest of having a consistent API I defined both.

The alternative would be relying on `pulpcore_present?` at https://github.com/theforeman/foreman-installer/pull/540/files#diff-d063850194f4a36e3364bdee036f452fR1 which doesn't feel correct even if it would accomplish the same thing while saving a few lines of codes that we are planning to remove in the not so distant future anyway.

ehelms (approved):
Untested, but the structure looks good

hooks/pre/13-clear_pulpcore_content.rb
```diff
@@ -0,0 +1,18 @@
+# See bottom of the script for the command that kicks off the script
+require 'English'
+
+def reset
+  stop_services
+  start_services(['postgresql']) if local_postgresql?
+  empty_db_in_postgresql('pulpcore') if pulpcore_enabled?
```
ekohl: This whole reset doesn't make sense if pulpcore is not enabled, right? Essentially this whole function becomes a noop then.

hooks/pre/13-clear_pulpcore_content.rb
```diff
@@ -0,0 +1,18 @@
...
+end
+
+if app_value(:clear_pulpcore_content) && !app_value(:noop)
+  response = ask('Are you sure you want to continue? This will clear all Pulpcore content including repositories and synced content. [y/n]')
```
ekohl: It doesn't remove synced content, only the database. Not sure it matters to end users, but they may be surprised to find the disk usage is still high.

hooks/pre/13-clear_pulpcore_content.rb
wbclark: Moved the condition to the block that actually calls reset, and added in the corresponding place for the pulp reset as well.

hooks/pre/13-clear_pulpcore_content.rb
wbclark: I took inspiration from (read: shamelessly copied) @ianballou 's work on https://github.com/theforeman/foreman-installer/pull/504 into a new helper in `hooks/boot/05-data-helpers.rb` and incorporated it in `hooks/pre/13-clear_pulpcore_content.rb`

ehelms:
Looking a bit closer, I am realizing that this does change Pulp 2 clear content behavior and raises the question for what does that mean for Pulp 3.

In Pulp 2, the clear content command just erased the content on disk and made a small database update but left the database in tact. This allowed users to either re-sync or switch to on-demand and free up disk space. This change appears to fully reset the content on disk and the database when using `--clear-pulp-content`.

As for Pulp 3, @ianballou @jlsherrill can just the content be cleared off disk for Pulp 3? Anything that needs to be done to the database (e.g. in Pulp 2 -- https://github.com/theforeman/foreman-installer/blob/develop/katello/hooks/pre/11-clear_pulp_data_feature.rb#L6)

wbclark:
I think then I should implement two separate options for each of pulp{,core}

A --clear-content option that only clears content from disk, and a --reset option that resets the DB and clears content from disk.

If this still makes sense to do for Pulp3, it should be pretty easy from my perspective to make the changes.

ehelms:
I believe that we don't ever want to reset the Pulp database without resetting the Foreman database and that is why we never had a `--reset-pulp-database` type command in the past. We could clear the content off disk without harm, but not the database.

ianballou:
After a conversation with the Pulp team, it sounds like Pulp won't take kindly to having content deleted from the disk. The Pulp database will still think the content is there and will not properly restore the content files.  I've filed an issue that should help: https://pulp.plan.io/issues/7340

Perhaps we should file another story to add a Pulp API endpoint for clearing out content?  It still might be a bit of a hack to clear out Pulp content like we did for Pulp 2.

ehelms:
@ianballou Good info! I sorta wonder then if, we should avoid this for Pulp 3 and provide, if it does not exist, the ability from Katello to turn a repository from immediate to on-demand with storage cleanup (if such a thing does not already exist). 

@wbclark sounds like this should drop the --clear-pulpcore-content all together and just have --clear-pulp-content for Pulp 2 (or perhaps we even drop --clear-pulp-content as part of getting users used to the fact that this won't be supported with Pulp 3)

jlsherrill:
@ehelms there's actually an upstream RFE to do this: https://pulp.plan.io/issues/5926

+1 to dropping it in the installer, the pulp2 way was a hack and the pulp 3 way would be an much worse hack IMO

wbclark:
Thanks all, I appreciate the helpful input.

I've removed `--clear-pulpcore-content` but kept `--clear-pulp-content` for the time being. I did keep the new helpers for pulpcore data reset as part of the complete data reset feature.

hooks/boot/05-data-helpers.rb
```diff
@@ -0,0 +1,135 @@
...
+  def reset_pulp
+    execute('rm -f /var/lib/pulp/init.flag')
+    empty_mongo
+    execute('rm -rf /var/lib/pulp/{distributions,published,repos}/*')
+  end
+end
+
+def clear_pulpcore_content
+  if File.directory?('/var/lib/pulp/docroot')
+    FileUtils.rm_rf(Dir.glob("/var/lib/pulp/docroot/*"))
```
ekohl: The installer can recreate the docroot. Can't we just remove `/var/lib/pulp/docroot` instead of globbing?

hooks/pre/12-clear_pulp_content.rb
```diff
@@ -0,0 +1,17 @@
+# See bottom of the script for the command that kicks off the script
+require 'English'
+
+def reset
+  stop_services
+  reset_pulp
+end
```
ekohl: I think this is small enough that it can be inlined instead of a separate method.

hooks/pre/12-clear_pulp_content.rb
```diff
@@ -0,0 +1,17 @@
+# See bottom of the script for the command that kicks off the script
```
ekohl: This isn't all that relevant anymore now.

hooks/pre/12-clear_pulp_content.rb
```diff
@@ -0,0 +1,17 @@
...
+require 'English'
```
ekohl: I don't think this is actually used.

hooks/pre/12-clear_pulp_content.rb
```diff
@@ -0,0 +1,17 @@
...
+def reset
+  stop_services
+  reset_pulp
+end
+
+if app_value(:clear_pulp_content) && pulp_enabled? && !app_value(:noop)
+  response = ask('Are you sure you want to continue? This will drop all Pulp 2 data including repositories and synced content. [y/n]')
+  if response.downcase != 'y'
+    $stderr.puts '** cancelled **'
+    exit(1)
```
ekohl: We shouldn't exit in `pre`. I think the confirmation should happen in `pre_commit` or perhaps even `pre_validations`.

ekohl (changes requested):

hooks/boot/05-data-helpers.rb
wbclark: I'll change this to `execute('rm -rf /var/lib/pulp/docroot/*')` to be consistent with what we are already doing in other hooks and helpers

hooks/pre/12-clear_pulp_content.rb
wbclark: Could you please elaborate on your reasoning here?

hooks/boot/05-data-helpers.rb
ekohl: Well, that still ends up globbing. I also think it doesn't remove hidden files. I think the previous approach was taken because `/var/lib/pulp` may be a mountpoint and you can't remove that. However, the media storage shouldn't be on its own mountpoint (all the temp dirs should be on the same mount). That means I think it's safe to remove the whole directory and let the installer recreate it.

hooks/pre/12-clear_pulp_content.rb
ekohl: https://github.com/theforeman/kafo#hooks has a list of hooks. In `pre` all state has been written to disk. That means the answers are going to be persisted. That means you can end up in a weird situation where things are out of sync.

Imagine you call `foreman-installer --my-password-param --clear-pulp-content`. foreman-maintain actually reads some values (including passwords) from the answers. If you cancel, these are now out of sync which means foreman-maintain may be broken.

However, if you exited in `pre_commit` nothing has been written to disk and no changes have been made. In other words, any calls to `exit()` in `pre` is highly suspicious.

This was clearly already wrong in katello-installer, but I'm using these moves as a good chance to review and properly fix behavior.

wbclark: Great, thanks. I've moved that logic to `pre_validations`

hooks/boot/05-data-helpers.rb
```diff
@@ -0,0 +1,135 @@
...
+  def reset_pulp
+    execute('rm -f /var/lib/pulp/init.flag')
+    empty_mongo
+    execute('rm -rf /var/lib/pulp/{distributions,published,repos}/*')
+  end
+end
+
+def clear_pulpcore_content
+  if File.directory?('/var/lib/pulp/docroot')
+    execute('rm -rf /var/lib/pulp/docroot')
```
ekohl: I think you can still use the native `FileUtils.rm_rf` to remove it. It's probably good to add a `logging.debug` line before it though.

hooks/boot/05-data-helpers.rb
```diff
@@ -0,0 +1,135 @@
...
+    logger.info 'Pulpcore content successfully removed.'
+  else
+    logger.warn 'Pulpcore content directory not present at \'/var/lib/pulp/docroot\''
```
ekohl: You're repeating the same path multiple times. A variable is a good idea.

hooks/pre/12-clear_pulp_content.rb
ekohl: Can you elaborate why you think `pre_validations` should be used over `pre_commit`?

wbclark: If we are giving the user the option to abort after realizing what `--clear-pulp-content` implies, it's better to do that before starting the interactive wizard.

wbclark:
@ekohl added https://github.com/theforeman/foreman-installer/pull/540/files#diff-39c98fdf403a2c84a86d3e68bae71945R1 and reverted to using `FileUtils`

additionally added debug level logging before content removal

also updated to use `content_dir` as a parameter of the method. currently it's always getting the value `/var/lib/pulp/docroot` but I think once we settle on https://github.com/theforeman/puppet-pulpcore/pull/115 or similar we can determine the value dynamically from answers.

hooks/pre/12-clear_pulp_content.rb
ekohl: I also debated this internally, but wanted to know your opinion. I also saw the argument that you want to exit early. However, technically you can turn off Pulp during the interactive phase, so the interaction becomes:

* Run installer with argument
* Confirm in dialog
* Turn Pulp off
* Hook doesn't run because `pulp_enabled?` returns false

So the logic I've been following is that if you use answers, `pre_commit` is the right phase.

This is a downside of using a helper like `pulp_enabled?`: it obfuscates that you use answers and what the implications are.

Btw, you can also wonder why we even need an `ask()` at all. The user passed in the argument and you're expected to know what you're doing.

I also realize now that `--reset-data` suffers from the same problem.

wbclark: So, a similar problem exists if you turn off pulp with a command line argument, which in my experience is more common than using the interactive mode (in fact, I've never met a single user that uses interactive mode):

- Run `foreman-installer --clear-pulp-content --no-enable-foreman-proxy-plugin-pulp`
- Regardless of whether the confirmation exists in `pre_commit` or `pre_validations`, the confirmation dialogue doesn't run because `pulp_enabled?` returns false (either due to being disabled in answers, or due to answers being unavailable)
- The reset logic itself doesn't run for the same reason, so the user believes they have disabled the plugin and reset all its data, but in fact the data remains.

With that said, I think there are bigger problems with disabling the plugin after it has been installed: puppet will do nothing to remove the existing pulp server, services, vhosts, mongoDB, etc -- it simply won't go out of its way any longer to ensure they remain present in the correct state either. So as we get closer to removing Pulp2 we need to come up with a strategy to really remove it rather than merely no longer ensuring its presence.

wbclark: Thinking about it more, I suppose that problem (pulp2 removal) is why the upgrade strategy to the point of pulp3 only installations won't be an in place upgrade at all but a data migration to a fresh install.

So I'm therefore not that worried about edge cases involving disabling the pulp2 plugin when it was previously enabled, since that already doesn't work the way an user might expect for a host of other reasons.

Still, as you pointed out, we can't rely on `pulp_enabled?` in `pre_validations`, even in the more common case where we are not trying to disable the plugin. So I think the options are to either move the confirmation to `pre_commit` after all as you suggested, or else remove it entirely.

ekohl: You're right that there are other issues. My goal right now is to figure out where things belong and get a good feeling for that. It helps to have these examples.

An in my mind I'm still going back and forth between the two. Either one have benefits and downsides. To me a strong argument about doing it before entering interactive mode is that you can't spend a lot of time answering questions and then end up throwing those away.

Another way to deal with it is to prompt at `pre_validations` and print a warning (probably in `pre`) that roughly says "we know you requested this, but we're not going to do it. That feels like the best compromise.

> puppet will do nothing to remove the existing pulp server, services, vhosts, mongoDB, etc

Well, it probably removes the vhosts since Apache is set up to purge unknown content. I know there are some customers who have turned that off, but they should know that's unsupported for upgrades.

There are Puppet modules which at the top level accept arguments like `service_ensure => stopped` and `package_ensure => absent`. There is a parameter `version` which we can set to `absent`, but then a lot of other things will break because the rest of our module is not designed for it.

wbclark: I like this idea.

I removed the `pulp_enabled?` condition from the confirmation dialogue in `pre_validations`. I'm considering removing the `!app_value(:noop)` condition from the confirmation dialogue as well, since the behavior of the interface should be the same regardless.

I then modified the logic in `pre` to skip the reset and print warnings if `!pulp_enabled?` or `app_value(:noop)`.

wbclark: Regarding confirmation dialogue with `--noop`, I ultimately elected to display a modified version of the dialogue in that that makes it clear nothing will actually happen. This keeps the user experience consistent.

ekohl (changes requested):
There's now a merge conflict.

That made me think. I would like to keep the hook context small. Given we're not using most methods in `DataResetHookContextExtension`, would you mind moving the changes back to the reset hook? Perhaps even nicer is to have the actual pulp clearing code in `hooks/pre/10-reset_data.rb`. That keeps all reset code in the same place. There you may even be able to detect passing in both `--reset-data` and `--clear-pulp-content`.

Some background info: that would also make this patch easier to review. Number of lines changed is a large factor in that. This can also be mitigated by splitting in to multiple smaller patches, even if they're in the same PR. After changing that, splitting may not be needed anymore.

hooks/boot/05-data-helpers.rb
```diff
@@ -0,0 +1,138 @@
+require 'fileutils'
+
+module DataResetHookContextExtension
```
ekohl: I've started to think we shouldn't do have this. Rather, all data reset functionality should be in `hooks/pre/10-reset_data.rb`. We should not use these things in other hooks.

Another effect of moving the pulp reset hook handling in that file is that you can prevent the double mongo clear if `--reset-data` and `--clear-pulp-content` are both passed in with a simple `if/else`.

wbclark:
@ekohl I split this into two commits to make it easier to review

ekohl:
I still think the entire hook is a bad idea. It means we'll have larger and larger hook context that becomes hard to reason about. You have no documentation about all the methods. It's something we shouldn't do too much. Having it all in a single file keeps it reasonably self documenting.

Thinking even further (and I'm not sure if we can actually do this) is that you would implement `--reset-data=pulp` at some point in the future (or `--reset-data=foreman`).

wbclark:
> I still think the entire hook is a bad idea. It means we'll have larger and larger hook context that becomes hard to reason about. You have no documentation about all the methods. It's something we shouldn't do too much. Having it all in a single file keeps it reasonably self documenting.
> 
> Thinking even further (and I'm not sure if we can actually do this) is that you would implement `--reset-data=pulp` at some point in the future (or `--reset-data=foreman`).

I'm thinking about a few key points here:

1. The Katello team has said that Katello can't really easily handle an external (i.e. via installer) reset of the Pulpcore component the same way it currently handles the reset of Pulp data.

2. As an user, I don't see a real use for separate reset of data by component, outside Pulp. Blowing away synced content on a content proxy or even Katello can make sense during troubleshooting paths but I don't see an obvious need for such granularity with other components.

Given that Pulp2 will ultimately be removed, why don't we keep its reset as a separate hook for now and not worry about otherwise harmless inefficiencies like resetting it twice in the unlikely case of `--clear-pulp-content --reset-data`?

I also originally implemented the `DataResetHookContextExtension` when it looked like `--clear-pulpcore-data` was a thing that made sense to do, to avoid duplication of methods. Since that assumption has changed and ultimately Pulp2 will be removed, I could move those methods back into `hooks/pre/10-reset_data.rb` and we could deal for now with having a few of them duplicated in `hooks/pre/12-clear_pulp_content.rb` with the understanding that the entire hook will be removed in the not so distant future. What do you think @ekohl ?

ekohl:
I don't really see a reason why you should split it over multiple files. The way I see it is that `hooks/pre/10-reset_data.rb` deals with resetting data. Clearing Pulp content is resetting data. That was what I tried to point out with that you could (in theory) map everything to a single parameter. Not that it should actually be done.

And if it's going away soon, that's probably even more reason not to move too many things around. Moving everything into `DataResetHookContextExtension` and then on removal back to the hook is a lot of noise in the git log for no real benefit.

ehelms:
I land on, given Pulp 2 will go away, let's just move the hook files as is, wrapping it in the proper checks for module present or enabled, to reduce the diff and achieve the goal here which is to be able to drop `katello/hooks` directory.

ekohl:
If we're going to say it's going away, I'd actually prefer to postpone the dropping of `katello/hooks` and wait till we can remove the hook. There's only a small benefit in removing the `katello/hooks` directory. Polluting the git log with just moving things around is a net negative value. It makes debugging over various releases harder.

ekohl:
Something I didn't notice at all was that in the old hook it actually didn't fully wipe mongodb but rather just executed a single command. That means that just moving it isn't duplicating code and also invalidates my argument to move it in with the reset data hook.

My main take away from this discussion is that we should not `exit()` in `pre`. I've submitted https://github.com/theforeman/foreman-installer/pull/570 to correct reset data.

wbclark:
I've added a 3rd commit which reverts the HookContextExtension and fixes up the `--clear-pulp-content` functionality.

The intention is that if the approach looks good, I will squash these before merging.

TODO: Undo changes in `hooks/pre/10-reset_data.rb` here and open a separate PR for those. I will do that after my next afternoon meeting.

wbclark:
I split off the changes which no longer make sense to be included here into a separate PR, https://github.com/theforeman/foreman-installer/pull/571

I then rebased this and squashed into a single commit which focuses on the `--clear-pulp-content` feature.

hooks/pre/12-clear_pulp_content.rb
```diff
@@ -0,0 +1,73 @@
+require 'fileutils'
+
+def clear_pulp_content
+  content_dir = '/var/lib/pulp/content'
+  if File.directory?(content_dir)
+    logger.debug "Removing Pulp content inside \'#{content_dir}\'."
+    FileUtils.rm_rf('/var/lib/pulp/content/*')
```
ekohl: https://ruby-doc.org/stdlib-2.4.1/libdoc/fileutils/rdoc/FileUtils.html#method-c-rm_r suggests you need to call glob explicitly.
```suggestion
    FileUtils.rm_rf(Dir.glob(File.join(content_dir, '*')))
```

I'm also wondering about the docs for it that mention a security risk. I guess this is a timing attack where it would start to process it, read a directory and then replace that with a symlink. I wouldn't be too worried about this.

hooks/pre/12-clear_pulp_content.rb
```diff
@@ -0,0 +1,73 @@
+require 'fileutils'
+
+def clear_pulp_content
+  content_dir = '/var/lib/pulp/content'
+  if File.directory?(content_dir)
+    logger.debug "Removing Pulp content inside \'#{content_dir}\'."
```
ekohl: You don't need to escape `'` inside `"`.

hooks/pre/12-clear_pulp_content.rb
```diff
@@ -0,0 +1,73 @@
...
+  }
+end
+
+def reset_repo_importers
+  mongo_config = load_mongo_config
+  if remote_host?(mongo_config[:host])
+    reset_repo_importers_remote_mongo(mongo_config)
+  else
+    start_services(['rh-mongodb34-mongod'])
+    execute("mongo pulp_database --eval 'db.repo_importers.update({"scratchpad": {$ne: null}}, {$set: {"scratchpad.repomd_revision": null}}, {"multi":true})'")
```
ekohl: You need to escape here. Previously it was in backticks but now that it's in quotes, you should. Given there are no variables, I'd suggest
```suggestion
    execute('mongo pulp_database --eval \'db.repo_importers.update({"scratchpad": {$ne: null}}, {$set: {"scratchpad.repomd_revision": null}}, {"multi":true})\'')
```

ekohl (changes requested):

wbclark:
Thanks for the reviews @ekohl . It is updated with proper escaping

ehelms (approved):

hooks/pre/12-clear_pulp_content.rb
```diff
@@ -0,0 +1,73 @@
+require 'fileutils'
+
+def clear_pulp_content
+  content_dir = '/var/lib/pulp/content'
+  if File.directory?(content_dir)
+    logger.debug "Removing Pulp content inside '#{content_dir}'."
+    FileUtils.rm_rf(Dir.glob(File.join(content_dir, '*')))
+    logger.debug "Removed Pulp content inside '\#{content_dir}\', now resetting Pulp Repo importers."
```
ekohl: I assumed I didn't need to repeat the escaping comment everywhere, but it was relevant for the whole method.

hooks/pre/12-clear_pulp_content.rb
```diff
@@ -0,0 +1,73 @@
...
+end
+
+def reset_repo_importers
+  mongo_config = load_mongo_config
+  if remote_host?(mongo_config[:host])
+    reset_repo_importers_remote_mongo(mongo_config)
+  else
+    start_services(['rh-mongodb34-mongod'])
+    execute('mongo pulp_database --eval \'db.repo_importers.update({"scratchpad": {$ne: null}}, {$set: {"scratchpad.repomd_revision": null}}, {"multi":true})\'')
+  end
```
ekohl: Rather than repeating the command, I'd put the eval content in a separate method. This avoids bugs where you make a modification in one place, but forget the other. The alternative is to make `reset_repo_importers_remote_mongo` handle both local and remote hosts. I suspect it already does this since the host defaults to localhost.

This means you have fewer code paths which is a good thing.
```suggestion
  start_services(['rh-mongodb34-mongod']) unless remote_host?(mongo_config[:host])
  reset_repo_importers_remote_mongo(mongo_config)
```

Then you probably also want to remote `_remote_mongo` from the method name.

hooks/pre/12-clear_pulp_content.rb
```diff
@@ -0,0 +1,73 @@
...
+    reset_repo_importers
+    logger.info 'Pulp content successfully removed.'
+  else
+    logger.warn "Pulp content directory not present at \'#{content_dir}\'."
+  end
+end
+
+def remote_host?(hostname)
+  !['localhost', '127.0.0.1', `hostname`.strip].include?(hostname)
+end
```
ekohl: This may make sense in the shared context since it's copied from the reset data hook.

ekohl (changes requested):

wbclark:
@ekohl I fixed the remaining issues with escaping and now use a single method `reset_repo_importers` for both local and remote mongodb.

I then added a 2nd commit which moves shared helpers to the hook context extension.

hooks/pre/12-clear_pulp_content.rb
```diff
@@ -0,0 +1,45 @@
+require 'fileutils'
+
+def clear_pulp_content
+  content_dir = '/var/lib/pulp/content'
+  if File.directory?(content_dir)
+    logger.debug "Removing Pulp content inside '#{content_dir}'."
+    FileUtils.rm_rf(Dir.glob(File.join(content_dir, '*')))
+    logger.debug "Removed Pulp content inside '#{content_dir}', now resetting Pulp Repo importers."
```
ekohl, line 10: I'd add an empty line for readability. You'll end up with 3 separate "paragraphs": stopping, clearing filesystem content, clearing DB.

```suggestion
    logger.debug "Removed Pulp content inside '#{content_dir}', now resetting Pulp Repo importers."

```

hooks/pre/12-clear_pulp_content.rb
```diff
@@ -0,0 +1,45 @@
+require 'fileutils'
+
+def clear_pulp_content
+  content_dir = '/var/lib/pulp/content'
+  if File.directory?(content_dir)
```
ekohl, line 5: I'd move `stop_services` into this section so you don't stop anything unless you really need to
```suggestion
  if File.directory?(content_dir)
    stop_services

```

hooks/pre/12-clear_pulp_content.rb
```diff
@@ -0,0 +1,45 @@
...
+
+if app_value(:clear_pulp_content)
+  if !pulp_enabled?
+    logger.warn 'Skipping Pulp content reset because Pulp feature is disabled in scenario answers.'
+  elsif app_value(:reset_data)
+    logger.warn "Skipping Pulp content reset because all Pulp data will be reset with '\--reset-data\'."
+  elsif app_value(:noop)
+    logger.warn 'Skipping Pulp data reset due to \'--noop\' flag.'
+  else
+    stop_services
```
ekohl: And then remove it here.
```suggestion
```

hooks/pre/12-clear_pulp_content.rb
```diff
@@ -0,0 +1,45 @@
...
+  cmd_base = '--eval \'db.repo_importers.update({"scratchpad": {$ne: null}}, {$set: {"scratchpad.repomd_revision": null}}, {"multi":true})\''
+  cmd = "mongo #{config[:database]} #{username} #{password} #{host} #{ssl} #{ca_cert} #{client_cert} #{cmd_base}"
+  execute(cmd)
+end
+
+if app_value(:clear_pulp_content)
+  if !pulp_enabled?
+    logger.warn 'Skipping Pulp content reset because Pulp feature is disabled in scenario answers.'
+  elsif app_value(:reset_data)
+    logger.warn "Skipping Pulp content reset because all Pulp data will be reset with '\--reset-data\'."
```
ekohl: Perhaps just make this empty. The user doesn't need to be warned about this IMHO.
```suggestion
```

wbclark: @ekohl If we are going to prevent `--clear-pulp-content` from running when using `--reset-data` then I think it absolutely makes sense to warn the user that using them together doesn't exactly make sense to do.

Then, I just noticed that `--reset-data` clears out `/var/lib/pulp/{distributions,published,repos}/*` but not `/var/lib/pulp/content`...

So either that directory should be included as well (I believe that is the case), or we shouldn't actually prevent both from running.

The way I'm thinking about it right now, I lean towards leaving the warning and opening another PR to make sure `--reset-data` also clears `/var/lib/pulp/content`

Do you agree @ekohl ?

wbclark:
@ekohl Updated the first commit to address the above comments, except the one about the `--reset-data` warning... please see above discussion about it and let me know your thoughts.

hooks/pre/12-clear_pulp_content.rb
ekohl: sing them together is redundant, but IMHO not a real issue. Perhaps the way to deal with this is to skip the database clean if `--reset-data` is also passed in?

Another option is to completely ignore it. It's been like this for ages and planned to be removed. Rather than trying to figure out how to deal with it properly, I'd be ok with ignoring it.

wbclark: @ekohl I think the handling already implemented in this PR becomes correct if `--reset-data` also clears `/var/lib/pulp/content`

I also think that `--reset-data` should also clear `/var/lib/pulp/content` given that it resets the pulp database completely and clears other directories in `/var/lib/pulp`... this seems like a bug to me that we should fix.

Do you disagree with either point?

ekohl: I wonder if `--reset-data` should clear the content. I've never used `--reset-data` nor stored any data in Pulp so I don't know what the expectation is. Perhaps Pulp can pick up content that's already on disk to prevent resyncing but doesn't publish it in any way?

ehelms: The expectation and use of this has been to clear the database and all content on disk. I think if Pulp 3 can handle clearing database and leaving content to be smartly picked up we could implement an additional tweak to the hook but shouldn't optimize that at this point in time.

wbclark: I confirmed with the Katello team it is indeed a bug -- there is no reason to preserve `/var/lib/pulp/content/*` when clearing the Pulp database -- so I opened https://projects.theforeman.org/issues/30776 and the fix with #572

ekohl: Ok, so now #572 is merged I think we can drop this warning. It's now purely a redundant parameter when the user passes it in but that's fine. No need for any warning. Note I'm **not** saying you should remove `elsif app_value(:reset_data)`, only the warning. You can also downgrade it to a debug message.

wbclark: OK, I changed the log level of the message to debug.

hooks/pre/12-clear_pulp_content.rb
```diff
@@ -0,0 +1,47 @@
...
+  cmd_base = '--eval \'db.repo_importers.update({"scratchpad": {$ne: null}}, {$set: {"scratchpad.repomd_revision": null}}, {"multi":true})\''
+  cmd = "mongo #{config[:database]} #{username} #{password} #{host} #{ssl} #{ca_cert} #{client_cert} #{cmd_base}"
+  execute(cmd)
+end
+
+if app_value(:clear_pulp_content)
+  if !pulp_enabled?
+    logger.warn 'Skipping Pulp content reset because Pulp feature is disabled in scenario answers.'
+  elsif app_value(:reset_data)
+    logger.debug "Skipping \'--clear-pulp-content\' because Pulp data and content will be reset with '\--reset-data\'."
```
ekohl: Still has redundant escaping. The warn below as well.

wbclark: Sorry, this should be in single quotes I think. Once that is fixed, the escaping becomes necessary.

wbclark: @ekohl There are similar issues in `hooks/pre/10-data_reset.rb` so I added a 3rd commit that fixes them. But if you prefer I can move that commit to a new branch and PR.

ekohl (approved):
Untested, but the code looks good to me.

PR #541, opened 2020-07-15
Fixes #29296 - remove --upgrade

wbclark:
This option currently does nothing except print a notice that `--upgrade` is no longer needed. This PR will remove it entirely.


ehelms (approved):
Cleanup for the win!

PR #542, opened 2020-07-16
Fixes #29801 - move --tuning to hooks/

wbclark:


ekohl (changes requested):
We also had a migration to set the default tuning profile. I think that was needed. It's also missing the boot hook. It also ends up with 2 hooks to set hiera.

I'm also :-1: on this for now. The tuning sizes are absolutely wrong and need to take more options into account. This raises the minimum required memory for vanilla installations from 4GB to 8GB for example.

A lot of tuning is also irrelevant. Most of it is setting up passenger which is not used anymore. I think we really need to rethink this before we make it globally available.

ehelms:
Two thoughts since combining hooks/ directories is not the same as turning on every hook for every scenario. 

 1) Could we store the tuning profile in a hiera file or the configuration per scenario?
 2) Could we enable tuning as an option based on a value in the scenarios configuration file?

ekohl:
> Could we store the tuning profile in a hiera file or the configuration per scenario?

We can check if the key is in the scenario to enable/disable it:
https://github.com/theforeman/foreman-installer/blob/79f509685e75141038f03e0505018ba7e0a920de/config/katello.yaml#L15-L16

So something like
```ruby
# At boot
if has_custom_fact?(:tuning)
  app_option(...)
end

# At pre_commit
if app_option?(:tuning)
end
```

It does require implementing `has_custom_fact?` and getting a release out, but that would allow us to define it per scenario. IMHO it's cleaner than the current method of checking whether the Foreman module is enabled.

ehelms:
@wbclark This is waiting on you to tackle the additional check of the fact

wbclark:
Trying a `has_custom_fact?` helper with https://github.com/theforeman/kafo/pull/266

hooks/pre_commit/13-tuning.rb
```diff
@@ -0,0 +1,76 @@
+# A mapping of tuning profile to the required CPU cores and memory in GB
```
ehelms: Should the logic of this whole file be also wrapped in the `has_custom_fact?('tuning')` check?

ekohl: It's already checking `if app_option?(:tuning)` which I think is cleaner. If we decide to expose the option on a different condition, it'll still work. I'm not sure about the constants.

hooks/pre_commit/13-tuning.rb
```diff
@@ -0,0 +1,76 @@
...
+    required_memory = required[:memory]
+
+    # Check if it's actually 90% of the required. If a crash kernel is enabled
+    # then the reported total memory is lower than in reality.
+    if facts[:memory][:system][:total_bytes] < (required_memory * 1024 * 1024 * 1024 * 0.9)
+      if app_value(:tuning)
+        say "<%= color('Insufficient memory for tuning size', :bad) %>"
+        say "Tuning profile '#{new_tuning}' requires at least #{required_memory} GB of memory and #{required_cores} CPU cores"
+      else
+        say "This scenario '#{scenario_id}' requires at least #{required_memory} GB of memory and #{required_cores} CPU cores"
```
ekohl: I know it was part of the previous code, but will this be easier to read?
```suggestion
        say "The #{scenario_id} scenario requires at least #{required_memory} GB of memory and #{required_cores} CPU cores"
```

ekohl (changes requested):
We have a tuning fact in the foreman-proxy-content scenario. This change will enable it again, so you also need a scenario migration where the fact is removed. This also needs to happen in the defaults.

hooks/pre_commit/13-tuning.rb
wbclark: I'll push an update wrapping the constants as well

wbclark:
> We have a tuning fact in the foreman-proxy-content scenario. This change will enable it again, so you also need a scenario migration where the fact is removed. This also needs to happen in the defaults.

Good catch. I added a second commit with the configuration change, migration, and a test for the migration.

config/foreman-proxy-content.migrations/200818160950-remove_tuning_fact.rb
```diff
@@ -0,0 +1,2 @@
+scenario[:facts] ||= {}
+scenario[:facts].delete('tuning') if scenario[:facts].key?('tuning')
```
ekohl: There's no need to introduce facts if you only want tor remove something.
```suggestion
scenario[:facts].delete('tuning') if scenario[:facts] && scenario[:facts].key?('tuning')
```

config/foreman-proxy-content.yaml
```diff
@@ -12,8 +12,7 @@
 :hook_dirs: ['./katello/hooks']
 :parser_cache_path: ./_build/parser_cache/foreman-proxy-content.yaml
 :hiera_config: ./config/foreman-hiera.yaml
-:facts:
-  tuning: 'default'
+:facts: {}
```
ekohl: I'd leave it out altogether
```suggestion
```

ekohl (commented):
We have older migrations which put the scenario fact in. I'd remove those since they no longer serve a purpose.

ekohl:
Note you can run tests locally using `bundle install && bundle exec rake spec`. Similarly, there's also a `rubocop` job.

wbclark:
@ekohl all green.

Before moving forward should we remove passenger parameters from the profiles in `config/foreman.hiera/tuning/*` ?

Do you want to make a change to memory/CPU requirements?

config/foreman-proxy-content.migrations/200818160950-remove_tuning_fact.rb
```diff
@@ -0,0 +1,2 @@
+scenario[:facts].delete('tuning') if scenario[:facts] && scenario[:facts].key?('tuning')
+scenario.delete(:facts) if scenario[:facts] && scenario[:facts] == {}
```
ekohl: Given the repetition of `if scenario[:facts]`, I'd expect a regular `if`
```suggestion
if scenario[:facts]
  scenario[:facts].delete('tuning') if scenario[:facts].key?('tuning')
  scenario.delete(:facts) if scenario[:facts].empty?
end
```

Note this also uses the cleaner `.empty?` method.

ekohl (commented):
> Before moving forward should we remove passenger parameters from the profiles in `config/foreman.hiera/tuning/*` ?

No, Passenger is still a supported solution. Having them present doesn't hurt non-passenger setups either.

> Do you want to make a change to memory/CPU requirements?

For now the code doesn't introduce the tuning parameter to regular Foreman, so I don't think there's a need to change anything right now.

wbclark:
Thanks @ekohl . Updated following your style suggestions and spec tests and rubocop are all green for me locally.

hooks/pre_commit/13-tuning.rb
```diff
@@ -0,0 +1,76 @@
...
+      end
+      exit EXIT_INSUFFICIENT_MEMORY
+    end
+
+    if facts[:processors][:count] < required_cores
+      if app_value(:tuning)
+        say "<%= color('Insufficient CPU cores for tuning size', :bad) %>"
+        say "Tuning profile '#{new_tuning}' requires at least #{required_memory} GB of memory and #{required_cores} CPU cores"
+      else
+        say "This scenario '#{scenario_id}' requires at least #{required_memory} GB of memory and #{required_cores} CPU cores"
```
ekohl: Probably good to align this to line 55
```suggestion
        say "The #{scenario_id} scenario requires at least #{required_memory} GB of memory and #{required_cores} CPU cores"
```

spec/fixtures/foreman-proxy-content-remove-tuning-fact/foreman-proxy-content.yaml
```diff
@@ -0,0 +1,29 @@
...
+  - foreman_proxy
+  - foreman_proxy::plugin::pulp
+  - foreman_proxy_content
+  - puppet
+
+# If using the Debian ruby-kafo package, uncomment this
+# :kafo_modules_dir: /usr/lib/ruby/vendor_ruby/kafo/modules
+
+# Unused, but remains present for older config migrations
+:mapping: {}
```
ekohl: It may be good to minimize the file and remove irrelevant parts. I think these can be safely dropped
```suggestion
```

ekohl (approved):
Overall :+1: but 2 small nits.

spec/fixtures/foreman-proxy-content-remove-tuning-fact/foreman-proxy-content.yaml
wbclark: In the process of cleaning this up, I made a few more changes. Particularly, I needed a new fixture for the `after` config, which I should have had all along rather than than relying on the default config to also serve as the fixture.

ekohl (approved):
Ok, let's see if we break nightly with this. Could you also prepare PRs to foreman-packaging to require Kafo 5.1? Both Debian and RPM.

tbrisker:
Looks like we did break nightlies - https://community.theforeman.org/t/katello-nightly-rpm-pipeline-649-failed/20226

PR #543, opened 2020-07-17
Refs #30423 - preserve old pulpcore MEDIA_ROOT on upgrades

wbclark:
Requires: https://github.com/theforeman/puppet-pulpcore/pull/113
Requires: https://github.com/theforeman/puppet-foreman_proxy_content/pull/275


config/foreman-proxy-content-answers.yaml
```diff
@@ -14,6 +14,7 @@ certs:
 foreman_proxy_content:
   proxy_pulp_isos_to_pulpcore: false
   proxy_pulp_yum_to_pulpcore: false
+  pulpcore_media_root: /var/lib/pulp
```
ekohl, line 17: This matches the default, so no need to include this.

ekohl (changes requested):
I don't think this is right. IMHO we should get everyone on the same layout and move files around if needed. The problem is that this means every upgrading user will forever be on a different layout than fresh installs which is hard to support.

Let's first agree on the correct layout (https://github.com/theforeman/puppet-pulpcore/pull/113#pullrequestreview-450769300) and then think about how to deal with upgrades.

config/foreman-proxy-content-answers.yaml
wbclark: It's included to make it easy to tell new installs apart from upgrades.

wbclark: https://github.com/theforeman/foreman-installer/pull/543/files#diff-77cd90f81394cb4c12f94334eb5e1489R9

wbclark:
> I don't think this is right. IMHO we should get everyone on the same layout and move files around if needed.

Is there anything stopping us from doing this within a migration @ekohl ?

ekohl:
Yes, it's not the appropriate place. Migrations are only for the configs and run in a RPM `%post` hook. Moving the files there is guaranteed to break things.

wbclark:
> Yes, it's not the appropriate place. Migrations are only for the configs and run in a RPM `%post` hook. Moving the files there is guaranteed to break things.

How then do you propose moving the uploaded content to the new MEDIA_ROOT directory on upgrades? I think you must have something in mind other than asking users to do it manually?

ekohl:
> How then do you propose moving the uploaded content to the new MEDIA_ROOT directory on upgrades? I think you must have something in mind other than asking users to do it manually?

A hook in `pre` makes the most sense to me. However, the steps we need to take (and we're extremely late in doing this):
* Agree on a directory layout
* Get the proper SELinux policy
* Work out a migration

This should really have been done before we even started to ship it in nightly. Now I just wrote something in puppet-pulpcore but forgot a TODO so others didn't see it.

ekohl:
Closing in favor of the alternative approach.

PR #544, opened 2020-07-28
Fixes #30506: Ensure selinux packages are installed before puppet run

ehelms:


hooks/pre/31_install_scl_packages.rb
```diff
@@ -1,8 +1,16 @@
...
+  end
+
+  packages << 'foreman-selinux' if foreman_server?
+  packages << 'katello-selinux' if katello_enabled?
+  packages << 'candlepin-selinux' if katello_enabled?
   packages << 'pulpcore-selinux' if pulpcore_enabled?
-  ensure_packages(packages, 'installed')
 end
+
+ensure_packages(packages, 'installed') unless packages.empty?
```
ekohl: `ensure_packages` already has the shortcircuit, no need to repeat this:
https://github.com/theforeman/foreman-installer/blob/79f509685e75141038f03e0505018ba7e0a920de/hooks/boot/01-kafo-hook-extensions.rb#L18-L19

```suggestion
ensure_packages(packages, 'installed')
```

ehelms:
To support the Puppet work around and allow breaking dependencies in packaging: https://github.com/theforeman/foreman-packaging/pull/5591

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -45,6 +45,10 @@ def foreman_server?
     module_enabled?('foreman')
   end
 
+  def katello_enabled?
+    module_enabled?('katello')
+  end
```
ekohl, line 50: This makes me wonder if we should replace all uses of `module_enabled?('katello')` but we didn't replace other uses of `module_enabled?`

ekohl (commented):
Overall :+1:, just a small nit at the end.

hooks/boot/01-kafo-hook-extensions.rb
ehelms: I will follow up with a PR not to muddy this one with more changes.

ekohl (approved):

PR #421, opened 2019-12-02
Refactor Katello entry points

ekohl:
Depends on https://github.com/theforeman/puppet-katello/pull/308.

This uses the separate classes in puppet-katello to expose in a more logical way to the end user.

The generated help text is below. Note that there's now `--candlepin-*` and `--pulp-*`. The content types are `--global-*`. With Pulp 3 these can be moved to `--pulp-*` parameters but for Pulp 2 Katello 2 needs to be statically configured. There's qpid module with a rather advanced parameter. Currently Kafo can't hide a single parameter as advanced.

```
Usage:
    foreman-installer [OPTIONS]

Options:

= Generic:
    --reset                       This option will drop the Katello database and clear all subsequent backend data stores. You will lose all data!
                                  Unfortunately, we can't detect a failure, so you should verify success manually.
                                  Dropping can fail when the DB is in use. (default: false)
    --clear-pulp-content          This option will clear all Pulp content from disk located in '/var/lib/pulp/content/'. (default: false)
    --clear-puppet-environments   This option will clear all published Puppet environments from disk. (default: false)
    --tuning INSTALLATION_SIZE    Tune for an installation size. Choices: default, medium, large, extra-large, extra-extra-large (default: "default")
    --disable-system-checks       This option will skip the system checks for memory. (default: false)
    --force-upgrade-steps         This option will force upgrade steps to run that are normally only run once. (default: false)
    --certs-update-server         This option will enforce an update of the HTTPS certificates (default: false)
    --certs-update-server-ca      This option will enforce an update of the CA used for HTTPS certificates. (default: false)
    --certs-update-all            This option will enforce an update of all the certificates for given host (default: false)
    --certs-reset                 This option will reset any custom certificates and use the self-signed CA instead. Note that any clients will need to be updated with the latest katello-ca-consumer RPM, and any external proxies will need to have the certs updated by generating a new certs tarball. (default: false)
    --certs-skip-check            This option will cause skipping the certificates sanity check. Use with caution (default: false)
    --upgrade-mongo-storage-engine Run the steps necessary to upgrade the MongoDB storage engine. (default: false)
    --upgrade                     Run the steps necessary for an upgrade such as migrations, rake tasks, etc. (default: false)
    --disable-resolve-mismatches  This will disable the resolving of mismatches between the application and backend services, during upgrade.  The steps will still run in a non-commit mode to show what would have been changed. (default: false)
    --[no-]colors                 Use color output on STDOUT (default: true)
[...]

PR #546, opened 2020-07-28
Fixes #30507: Use enabled hook helpers for Katello and Foreman

ehelms:


hooks/pre/10-reset_data.rb
```diff
@@ -5,7 +5,7 @@ def reset
   stop_services
   start_services(['postgresql']) if local_postgresql?
   empty_db_in_postgresql('foreman') if foreman_server?
-  reset_candlepin if module_enabled?('katello')
+  reset_candlepin if katello_enabled?
```
ekohl: Just thinking out loud: do we also want to start to prepare for an optional candlepin and introduce `candlepin_enabled?`?

katello/hooks/pre/15-check_java.rb
```diff
@@ -15,7 +15,7 @@ def error_exit(message, code)
   kafo.class.exit code
 end
 
-if module_enabled?('katello')
+if katello_enabled?
```
ekohl: This would also be `candlepin_enabled?`

katello/hooks/pre/20-certs_update.rb
```diff
@@ -18,7 +18,7 @@ def mark_for_update(cert_name, hostname = nil)
 cert_file = param('certs', 'server_cert').value
 key_file  = param('certs', 'server_key').value
 
-if app_value('certs_update_server_ca') && !module_enabled?('katello')
+if app_value('certs_update_server_ca') && !katello_enabled?
   fail_and_exit("--certs-update-server-ca needs to be used with katello", 101)
 end
```
ekohl, line 23: This part should really be a `pre_commit` hook IMHO. In `pre` the answers are already saved to disk. You should fail on invalid user input in `pre`.

It should also check if the `certs` module is enabled, rather than `katello`, right? It checks if the `certs` params are passed in and I think that `server_ca_cert` can also be used on a content proxy.

hooks/pre/10-reset_data.rb
ehelms: I thought about that with my prior PR but we do not have a clean way to detect that I could think of. So I kept it to a clearer representation of how things are.

katello/hooks/pre/20-certs_update.rb
ehelms: File an issue if you don't mind, touching our certs code requires a Friday :P

hooks/pre/10-reset_data.rb
ekohl: Well, for now we just have it equal to `katello_enabled?` but it prepares the rest of the codebase. Once we circle back to https://github.com/theforeman/foreman-installer/pull/421 it will be good.

ehelms:
Added candlepin_enabled? helper

katello/hooks/pre_commit/14-cdn_setting.rb
```diff
@@ -1,4 +1,4 @@
-if module_enabled?('katello') &&
+if katello_enabled? &&
```
ekohl: I think this doesn't make that much sense. This is very much into the actual config and answers. It's not about a high level concept.

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -58,7 +58,7 @@ def local_foreman_db?
   end
 
   def local_candlepin_db?
-    module_enabled?('katello') && param_value('katello', 'candlepin_manage_db')
+    katello_enabled? && param_value('katello', 'candlepin_manage_db')
```
ekohl: I think this only makes it less clear how this works. I'd suggest this only reduces the understanding of the flow.

If anything, this should be `candlepin_enabled?`

ehelms:
This could use another round of review

ekohl (commented):
Looks like I had comments that I failed to submit

katello/hooks/pre_commit/14-cdn_setting.rb
ehelms: I don't follow the request here.

katello/hooks/pre/15-check_java.rb
```diff
@@ -15,7 +15,7 @@ def error_exit(message, code)
   kafo.class.exit code
 end
 
-if module_enabled?('katello')
+if candlepin_enabled?
   java_version_string = `/usr/bin/java -version 2>&1`
   java_version = java_version_string.split("\n")[0].split('"')[1]
```
ekohl, line 20: On an unrelated note: I think this check itself should ensure JDK 1.8 is used since 1.7 is too old.

katello/hooks/boot/13-hiera.rb
```diff
@@ -1,4 +1,4 @@
-if module_enabled?('foreman')
+if foreman_server?
```
ekohl: I'm going to hold off on merging this in favor of merging https://github.com/theforeman/foreman-installer/pull/542 first. IMHO that solution is cleaner and you get to deal with the merge conflict.

ekohl (commented):
Overall this looks ok, but holding off until we merge tuning.

ekohl (approved):

PR #547, opened 2020-07-29
Refs #30506: Install crane-selinux if pulp enabled

ehelms:


ekohl (approved):
Just a thought on packaging: we moved katello-selinux into the main foreman repo since foreman-installer depended on it but since that dependency has now been dropped, should we move it back to the katello packages again?

ehelms:
Yes, we can re-do or un-do that dance of selinux packages.

PR #552, opened 2020-07-30
added foreman_datacenter plugin

Zenya:


ekohl (changes requested):
This fails because there is no datacenter plugin in the foreman module. This would first need an addition here: https://github.com/theforeman/puppet-foreman/tree/master/manifests/plugin

PR #553, opened 2020-08-03
Fixes #30553 - Add foreman_statistics plugin

ekohl:


config/katello.migrations/200803163936-add-statistics-plugin.rb
```diff
@@ -0,0 +1 @@
+answers['foreman::plugin::statistics'] ||= false
```
tbrisker, line 1: Not sure how much it matters, but you're missing a `20` in the file name

ekohl: It was added by the script, in katello they all skip the leading `20`.

ekohl:
Tests are failing because I didn't update the answers in the migration tests.

config/katello.migrations/200803163936-add-statistics-plugin.rb
tbrisker: ok, then we should be good until 2100 ;)

ehelms (approved):

PR #557, opened 2020-08-05
fixes #30569 - Added migration to change logging pattern

domitea:
Complementary PR for https://github.com/theforeman/puppet-foreman/pull/847

This PR contains migration for changing logging layout

Thanks @ekohl for mention this!

### DIFF
--- a/config/foreman.migrations/20200805102253_change_logging_layout.rb
+++ b/config/foreman.migrations/20200805102253_change_logging_layout.rb
@@ -1,0 +1,5 @@
+mod = answers['foreman']
+layout_key = 'logging_layout'
+if mod.is_a?(Hash) && mod[layout_key] == 'pattern'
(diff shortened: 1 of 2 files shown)


ekohl (changes requested):
Please copy the migration to `config/katello.migrations`. Note it uses a slightly different naming convention, but `bin/create-migration` can deal with that.

ekohl:
Nit: please use `Add` instead of `Added`. https://chris.beams.io/posts/git-commit/ explains why using the imperative mood is a good thing.

ekohl:
It's also better to reuse the Redmine issue because only together do they fix the issue. So here you'd use `Refs #29938`.

domitea:
@ekohl migration to Katello added and commit message reworded. Thanks for the link!

ehelms:
Close and re-opening to re-trigger tests

ehelms (approved):

PR #559, opened 2020-08-19
Fixes #30402: Set :warn to yellow

ehelms:


ekohl (commented):
This feels like something that also needs to be applied to develop or fixed in [Kafo's color scheme](https://github.com/theforeman/kafo/blob/master/lib/kafo/color_scheme.rb) so we don't have future regressions.

PR #562, opened 2020-08-20
Add module checkout scripts

ekohl:
This is mainly intended for the stable branches. It allows checking out all modules based on Puppetfile.lock. This is done by having a directory called workdir. Within that there's a directory 'bare' which contains all the modules as bare checkouts. Then in the modules a directory is created for the MAJOR.MINOR version. Within that directory there's a checkout for each module owned by theforeman/Katello.

The output of module-versions is changed to make it easier in the checkout-module script.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ modules/*
 .bundle/
 Gemfile.lock
 config/foreman.migrations/.applied
+workdir/
 .vendor/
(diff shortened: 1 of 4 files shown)


ekohl:
I just ran into an edge case I didn't anticipate: you can only have a single checkout of a branch. So that means you can't checkout master for both 2.1 and 2.3.

PR #564, opened 2020-08-20
Fixes #28928: Error if checkpoint_segments detected

ehelms:


hooks/pre_validations/10-postgresql_checkpoint_segments.rb
```diff
@@ -0,0 +1,17 @@
+custom_hiera = YAML.load_file('/etc/foreman-installer/custom-hiera.yaml')
+
+if custom_hiera.key?('postgresql::server::config_entries') &&
+     custom_hiera['postgresql::server::config_entries'].key?('checkpoint_segments')
+
+  message= <<~HEREDOC
+  ERROR:
+    checkpoint_segments tuning found in /etc/foreman-installer/custom-hiera.yaml
+    This tuning option is no longer valid in PostgreSQL 12+
```
ekohl: Technically it's no longer valid since 9.5.

hooks/pre_validations/10-postgresql_checkpoint_segments.rb
```diff
@@ -0,0 +1,17 @@
...
+    Please remove this from the following locations and then re-run the installer:
+      - /etc/foreman-installer/custom-hiera.yaml and re-run the installer
+      - /var/opt/rh/rh-postgresql12/lib/pgsql/data/postgresql.conf
```
ekohl: This path is only relevant for EL7. It's unlikely other users will hit this, but worth to consider.

Users who have not yet upgraded will not have this file either since this is at the pre_validations stage. That might confuse users. Perhaps we should check if the file exists before suggesting this?

hooks/pre_validations/10-postgresql_checkpoint_segments.rb
```diff
@@ -0,0 +1,17 @@
+custom_hiera = YAML.load_file('/etc/foreman-installer/custom-hiera.yaml')
```
ekohl: Given the path is repeated 3 times, should it be in a variable?

ekohl (commented):
This was roughly what I was thinking about as well.

ehelms:
Rubocop update should alleviate lint failures -- https://github.com/theforeman/foreman-installer/pull/565

ekohl (approved):
Code wise :+1: once we get Rubocop happy. Not sure if the phrasing of the message is the best but I'm not sure what to improve. In the Katello scenario, should we also hint at `--tuning`?

mccun934:
+1 to a note in the error:

We recommend tuning via the predefined templates available with the --tuning parameter. 
Consider reviewing your existing  #{HIERA_FILE} configuration values that can be replaced
with our built in tuning profiles.

ekohl:
> We recommend tuning via the predefined templates available with the --tuning parameter.

I always prefer the impersonal form. Maybe also a slightly different take on it:

The presence of checkpoint_segments in #{HIERA_FILE} indicates manual tuning. Manual tuning can override values provided by the --tuning parameter. Review #{HIERA_FILE} for values that are already provided by the built in tuning profiles. Built in tuning profiles also provide a supported upgrade path.

mccun934:
:+1:  to @ekohl wording. The impersonal is superior

ekohl (approved):
:+1: provided Travis is green. I've approved https://github.com/theforeman/foreman-installer/pull/565 with a small comment that you may want to ignore. Feel free to merge that, rebase this and merge it.

hooks/pre_validations/10-postgresql_checkpoint_segments.rb
```diff
@@ -0,0 +1,30 @@
...
+    message += "    - #{SCL_POSTGRESQL_CONF}"
+  end
+
+  message += <<~HEREDOC
+
+    The presence of checkpoint_segments in #{HIERA_FILE} indicates manual tuning.
+    Manual tuning can override values provided by the --tuning parameter.
+    Review #{HIERA_FILE} for values that are already provided by the built in tuning profiles.
+    Built in tuning profiles also provide a supported upgrade path.
+  HEREDOC
```
ekohl: One thing to note about this is that `--tuning` is only available on the Katello scenario. However, PostgreSQL can be tuned on any. Perhaps that needs some conditional.

hooks/pre_validations/10-postgresql_checkpoint_segments.rb
```diff
@@ -0,0 +1,30 @@
+HIERA_FILE = '/etc/foreman-installer/custom-hiera.yaml'
+SCL_POSTGRESQL_CONF = '/var/opt/rh/rh-postgresql12/lib/pgsql/data/postgresql.conf'
+
+custom_hiera = YAML.load_file(HIERA_FILE)
+
+if custom_hiera.key?('postgresql::server::config_entries') &&
```
ekohl: This also needs to check if it isn't `nil`. By default the file has nothing, not even a hash and you'll end up with `nil.key?`.

PR #566, opened 2020-08-27
Refs #29801 - Handle mongodb cache size separate from tuning

ehelms:

PR #568, opened 2020-08-27
Refs #29801: Fix broken mongo cache size setting

ehelms:
I goofed the concept and thus the logic here.


wbclark (approved):
This is consistent with the original commit that introduced it. Somehow the formatting got butchered in my editor when I was moving things around. Sorry for not catching it.

wbclark:
Oh, it happened in https://github.com/theforeman/foreman-installer/pull/566/files

Regardless, the logic looks consistent now with the original working implementation, even in the case where `!app_option?(:tuning)` so I will merge it.

PR #569, opened 2020-08-31
Refs #30523: Let puppet-candlepin handle DB migration workflow

ehelms:
Requires: https://github.com/theforeman/puppet-candlepin/pull/161

### DIFF
--- a/hooks/pre/10-reset_data.rb
+++ b/hooks/pre/10-reset_data.rb
@@ -52,7 +52,7 @@ def empty_db_in_postgresql(db)
 end
 
 def reset_candlepin
-  execute('rm -f /var/lib/candlepin/{cpdb_done,cpinit_done}')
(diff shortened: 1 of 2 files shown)


hooks/pre/10-reset_data.rb
```diff
@@ -52,7 +52,7 @@ def empty_db_in_postgresql(db)
 end
 
 def reset_candlepin
-  execute('rm -f /var/lib/candlepin/{cpdb_done,cpinit_done}')
+  execute('rm -f /var/lib/candlepin/.puppet-candlepin*')
```
ekohl: Let's be slightly stricter:
```suggestion
  execute('rm -f /var/lib/candlepin/.puppet-candlepin-cpdb-*')
```

Maybe we need to store more state in the future

ehelms: Fair.

ekohl (approved):

PR #570, opened 2020-09-01
Refs #29797 - Move interactive part to pre_validations

ekohl:
In the pre phase everything has been committed to disk. It should never call exit based on user input. By moving the interactive part of pre_validations, the user is asked before any changes to the system are made.


ekohl:
This should be cherry picked into 2.2.

ehelms (approved):

PR #574, opened 2020-09-08
Foreman 2.2.0-rc2 cherry picks

ekohl:
I didn't check the various modules, but these are at least the picks for the installer itself.


tbrisker (approved):
thanks @ekohl !

PR #575, opened 2020-09-08
Fixes #30802: Set proxy timeout to 15 minutes

ehelms:
This sets the Foreman HTTPs reverse proxy timeout to 15 minutes to
give a cushion for some requests (such as provisioning) that can
keep an open request for 10+ minutes. This sets a sane default and
users can then tune it to their own infrastructure needs.

### DIFF
--- a/config/foreman.hiera/common.yaml
+++ b/config/foreman.hiera/common.yaml
@@ -9,3 +9,7 @@ dhcp::config_comment: |
   foreman-installer. Use custom-hiera.yaml for multiple subnets.
 
 mongodb::server::set_parameter: "{diagnosticDataCollectionEnabled: false}"
+
+foreman::config::apache::proxy_params:
+  retry: '0'
+  timeout: '900'


config/foreman.hiera/common.yaml
```diff
@@ -9,3 +9,5 @@ dhcp::config_comment: |
   foreman-installer. Use custom-hiera.yaml for multiple subnets.
 
 mongodb::server::set_parameter: "{diagnosticDataCollectionEnabled: false}"
+
+foreman::config::apache::proxy_params: {'retry': '0', 'timeout': '900'}
```
ekohl: Why do you use this syntax? Doesn't this work as well?
```suggestion
foreman::config::apache::proxy_params:
  retry: '0'
  'timeout': '900'
```

Btw, not sure why we quoted the value in the Puppet manifest. Not sure that's really needed.

ehelms: YAMl vs Ruby brain. I updated to this cleaner style.

ekohl (approved):

PR #576, opened 2020-09-10
Refs #30489 - enable dhcp acls by default 

ezr-ondrej:
Both puppet-dhcp and puppet-foreman_proxy are ready to handle acls on both supported systems.
The best approach here is to leverage this to our favor instead of enforcing a user or mode on config files.

- [x] Needs https://github.com/theforeman/puppet-foreman_proxy/pull/614

### DIFF
--- a/config/foreman.migrations/20200909151007_manage_acls_on_debian.rb
+++ b/config/foreman.migrations/20200909151007_manage_acls_on_debian.rb
@@ -1,0 +1,3 @@
+if answers['foreman_proxy'].is_a?(Hash) && facts[:os][:family] == 'Debian'
+  answers['foreman_proxy']['dhcp_manage_acls'] ||= true if answers['foreman_proxy'].key?('dhcp_manage_acls')
+end


config/foreman.migrations/20200909151007_manage_acls_on_debian.rb
```diff
@@ -0,0 +1 @@
+answers['foreman_proxy::dhcp_manage_acls'] ||= true
```
ekohl: The structure is a bit different. `answers` is a mapping of class name to the answers of that class. The answers can be `false` (class disabled), `true` (class enabled, but no specific answers`) or a hash, mapping parameter to value.
```suggestion
if answers['foreman_proxy'].is_a?(Hash)
  answers['foreman_proxy']['dhcp_manage_acls'] ||= true
end
```

ekohl (changes requested):
Might be good to add a `Refs #30489` here as well since it's part of that.

config/foreman-answers.yaml
```diff
@@ -54,7 +54,7 @@ foreman::plugin::snapshot_management: false
 foreman::plugin::statistics: false
 foreman::plugin::tasks: false
 foreman::plugin::templates: false
-foreman_proxy: {}
+foreman_proxy: { dhcp_manage_acls: true }
```
ekohl: This shouldn't be needed if we change the module default (which we did).

config/foreman.migrations/20200909151007_manage_acls_on_debian.rb
```diff
@@ -0,0 +1,3 @@
+if answers['foreman_proxy'].is_a?(Hash)
```
ekohl: I think you can check for Debian here to limit the scope
```suggestion
if answers['foreman_proxy'].is_a?(Hash) && facts[:os][:family] == 'Debian'
```

Untested though.

config/foreman-answers.yaml
ezr-ondrej: Right, we just need to release, right?

ekohl: Nightly builds use git master

ekohl:
It looks like that breaks the tests. Perhaps I'm wondering what the cleanest solution is. Perhaps explicitly checking if the key is set in the hash?

Could you also use `Refs #30489` in the commit message?

ezr-ondrej:
The check for the key helped :+1:
Added `Refs ` to the title, it was in the commit message already.

ekohl (approved):

PR #577, opened 2020-09-14
Fix pulpcore assets directory permissions

ekohl:
Prior to Katello 3.18 assets were built by root. Katello 3.18 runs it as pulp and this corrects the permissions.

This goes together with https://github.com/theforeman/puppet-pulpcore/pull/121.

### DIFF
--- a/hooks/pre/33-pulpcore_assets_permissions.rb
+++ b/hooks/pre/33-pulpcore_assets_permissions.rb
@@ -1,0 +1,11 @@
+# Prior to Katello 3.18 assets were built by root. Katello 3.18 runs it as pulp
+# and this corrects the permissions.
+unless app_value(:noop)
+  DIRECTORY = '/var/lib/pulp/assets'.freeze
+  USER = 'pulp'.freeze
+
+  if File.directory?(DIRECTORY) && File.stat(DIRECTORY).uid == 0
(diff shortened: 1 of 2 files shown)


jlsherrill (approved):

jlsherrill:
tested this by hand and it worked well

hooks/pre/33-pulpcore_assets_permissions.rb
```diff
@@ -0,0 +1,9 @@
+# Prior to Katello 3.18 assets were built by root. Katello 3.18 runs it as pulp
+# and this corrects the permissions.
+unless app_value(:noop)
+  DIRECTORY = '/var/lib/pulp/assets'
+  if File.directory?(DIRECTORY) && File.stat(DIRECTORY).uid == 0
+    require 'fileutils'
+    FileUtils.chown_R('pulp', 'pulp', DIRECTORY)
```
ekohl: This does assume the pulp user exists. I'm wondering if there's a case where that's not true, but `/var/lib/pulp/assets` does exist and owned by root. Backups perhaps? I'll push an update that checks it. Need fix rubocop anyway.

ekohl:
Updated

jlsherrill (approved):
ack again

PR #580, opened 2020-09-21
Fixes #30891 - Create /var/log/redis with correct selinux label

wbclark:


wbclark:
I encountered this issue while working on https://github.com/theforeman/forklift/pull/1208

Happily enough, the same PR provides exactly what we need to test it (I'm running those now)

hooks/pre/32-install_selinux_packages.rb
```diff
@@ -1,5 +1,5 @@
 # Working around https://tickets.puppetlabs.com/browse/PUP-10548
-if facts.dig(:os, :selinux, :enabled)
+if facts[:os]['selinux']['enabled']
```
ekohl: I thought we symbolized all facts. I do think using `dig` is cleaner, but perhaps we should have some very early check (`pre_values`) that does roughly
```ruby
if facts.dig(:os, 'selinux', 'enabled').nil?
  fail('Unable to detect SELinux status')
end
```

Because facter should always set the boolean, even when it's disabled.

wbclark: In fact in just failed for me due to that reason. Trying with symbols now but I have a feeling it will also fail (as we believe it did with dig).

I might temporarily remove the condition entirely, just to prove to myself that this is where the issue lies, then come back and work out the correct syntax.

wbclark:
It still failed when I tried `if true` debugging, so the issue isn't the conditional. The package does get properly pre-installed by the hook, it's just that preinstallation doesn't prevent the selinux issue from occuring.

After convincing myself it is properly preinstalled, I still needed to do a restorecon to get the `rh-redis5-redis.service` up.

[root@centos7-foreman-nightly ~]# systemctl restart rh-redis5-redis.service
Job for rh-redis5-redis.service failed because the control process exited with error code. See "systemctl status rh-redis5-redis.service" and "journalctl -xe" for details.
[root@centos7-foreman-nightly ~]# restorecon -RvF /var/log/redis
restorecon reset /var/log/redis context unconfined_u:object_r:var_log_t:s0->system_u:object_r:redis_log_t:s0
[root@centos7-foreman-nightly ~]# systemctl restart rh-redis5-redis.service
[root@centos7-foreman-nightly ~]#

wbclark:
I was suspecting a root cause like https://tickets.puppetlabs.com/browse/PUP-7559 which could be worked around in puppet-redis similar to https://github.com/voxpupuli/puppet-redis/pull/202/files but I tested setting `selinux_ignore_defaults => true` for the file resource that creates /var/log/redis and still I faced the same issue.

In fact, even when I bypass puppet entirely and create the directory on a fresh centos7 box, I face the same issue:

[root@centos7 ~]# matchpathcon /var/log/redis
/var/log/redis	system_u:object_r:redis_log_t:s0
[root@centos7 ~]# mkdir /var/log/redis
[root@centos7 ~]# ll -aZ /var/log/redis/
drwxr-xr-x. root root unconfined_u:object_r:var_log_t:s0 .
drwxr-xr-x. root root system_u:object_r:var_log_t:s0   ..
[root@centos7 ~]# restorecon -RvF /var/log/redis
restorecon reset /var/log/redis context unconfined_u:object_r:var_log_t:s0->system_u:object_r:redis_log_t:s0
[root@centos7 ~]#

wbclark:
[root@centos7 ~]# mkdir -Z /var/log/redis
[root@centos7 ~]# ll -Zd /var/log/redis
drwxr-xr-x. root root unconfined_u:object_r:redis_log_t:s0 /var/log/redis
[root@centos7 ~]# restorecon -v /var/log/redis
[root@centos7 ~]# 

Now, I need to get puppet to do that or work around it in https://github.com/theforeman/forklift/pull/1208 by creating the directory and performing restorecon in advance

hooks/pre/32-install_selinux_packages.rb
```diff
@@ -14,4 +14,5 @@
   packages << 'crane-selinux' if pulp_enabled?
 
   ensure_packages(packages, 'installed')
+  execute('mkdir -Z /var/log/redis') if local_redis?
```
ekohl, line 17:
```console
# semanage fcontext --list | grep /var/log/redis
/var/log/redis(/.*)?                               all files          system_u:object_r:redis_log_t:s0 
```

I'm not sure how I can see which policy contains this, but I think it's in the base policy. Not sure why this is needed. Do you have a detailed log?

ekohl: Oh, and I doubt this is idempotent so it will break on upgrades.

wbclark: The problem is that the policy is not used when creating the directory. It does so only with the -Z flag, otherwise the context is computed by the kernel using transition rules, which doesn't always lead to the desired context. Compare the two scenarios below on a newly provisioned centos7 box:

```
[root@centos7 ~]# matchpathcon /var/log/redis
/var/log/redis system_u:object_r:redis_log_t:s0
[root@centos7 ~]# mkdir /var/log/redis
[root@centos7 ~]# ll -dZ /var/log/redis/
drwxr-xr-x. root root unconfined_u:object_r:var_log_t:s0 /var/log/redis/
[root@centos7 ~]# restorecon -v /var/log/redis
restorecon reset /var/log/redis context unconfined_u:object_r:var_log_t:s0->system_u:object_r:redis_log_t:s0
[root@centos7 ~]#
```

vs.

```
[root@centos7 ~]# mkdir -Z /var/log/redis
[root@centos7 ~]# ll -Zd /var/log/redis
drwxr-xr-x. root root unconfined_u:object_r:redis_log_t:s0 /var/log/redis
[root@centos7 ~]# restorecon -v /var/log/redis
[root@centos7 ~]#
```

Now observe the same issue when installing with foreman-installer:

```
    [ INFO 2020-09-30T11:34:01 verbose] Installer finished in 300.29119112 seconds
      Something went wrong! Check the log for ERROR-level output
      The full log is at ./_build/foreman.log
```

Checking the log, the issue was: `[ERROR 2020-09-30T11:30:57 main] Systemd start for rh-redis5-redis failed!`

Checking the journal to determine why that occurred:

```
[root@centos7-foreman-nightly ~]# journalctl | grep -v dynflow | grep -i redis
Sep 30 11:29:37 centos7-foreman-nightly.mercury.example.com yum[8014]: Installed: rh-redis5-runtime-3.3-1.el7.x86_64
Sep 30 11:29:37 centos7-foreman-nightly.mercury.example.com groupadd[8032]: group added to /etc/group: name=redis, GID=991
Sep 30 11:29:37 centos7-foreman-nightly.mercury.example.com groupadd[8032]: group added to /etc/gshadow: name=redis
Sep 30 11:29:37 centos7-foreman-nightly.mercury.example.com groupadd[8032]: new group: name=redis, GID=991
Sep 30 11:29:37 centos7-foreman-nightly.mercury.example.com useradd[8037]: new user: name=redis, UID=995, GID=991, home=/var/opt/rh/rh-redis5/lib/redis, shell=/sbin/nologin
Sep 30 11:29:37 centos7-foreman-nightly.mercury.example.com yum[8014]: Installed: rh-redis5-redis-5.0.5-1.el7.x86_64
Sep 30 11:30:21 centos7-foreman-nightly.mercury.example.com yum[8416]: Installed: tfm-rubygem-redis-4.1.2-2.el7.noarch
Sep 30 11:30:57 centos7-foreman-nightly.mercury.example.com systemd[1]: Starting Redis persistent key-value database...
Sep 30 11:30:57 centos7-foreman-nightly.mercury.example.com redis-server[8812]: *** FATAL CONFIG FILE ERROR ***
Sep 30 11:30:57 centos7-foreman-nightly.mercury.example.com redis-server[8812]: Reading the configuration file, at line 96
Sep 30 11:30:57 centos7-foreman-nightly.mercury.example.com redis-server[8812]: >>> 'logfile /var/log/redis/redis.log'
Sep 30 11:30:57 centos7-foreman-nightly.mercury.example.com redis-server[8812]: Can't open the log file: Permission denied
Sep 30 11:30:57 centos7-foreman-nightly.mercury.example.com systemd[1]: rh-redis5-redis.service: main process exited, code=exited, status=1/FAILURE
Sep 30 11:30:57 centos7-foreman-nightly.mercury.example.com systemd[1]: Failed to start Redis persistent key-value database.
Sep 30 11:30:57 centos7-foreman-nightly.mercury.example.com systemd[1]: Unit rh-redis5-redis.service entered failed state.
Sep 30 11:30:57 centos7-foreman-nightly.mercury.example.com systemd[1]: rh-redis5-redis.service failed.
[root@centos7-foreman-nightly ~]# 
```

The issue was Permission denied when trying to open /var/log/redis/redis.log. There is an AVC denial:

```
[root@centos7-foreman-nightly ~]# grep -i avc /var/log/audit/audit.log | grep redis
type=AVC msg=audit(1601465457.218:2113): avc:  denied  { write } for  pid=8812 comm="redis-server" name="redis" dev="vda1" ino=101239371 scontext=system_u:system_r:redis_t:s0 tcontext=unconfined_u:object_r:var_log_t:s0 tclass=dir permissive=0
```

The policy looks correct:

```
[root@centos7-foreman-nightly ~]# semanage fcontext -l | grep /var/log/redis
/var/log/redis(/.*)?                               all files          system_u:object_r:redis_log_t:s0 
[root@centos7-foreman-nightly ~]# matchpathcon /var/log/redis/
/var/log/redis	system_u:object_r:redis_log_t:s0
```

And yet:

```
[root@centos7-foreman-nightly ~]# ll -dZ /var/log/redis/
drwxr-x---. redis redis unconfined_u:object_r:var_log_t:s0 /var/log/redis/
[root@centos7-foreman-nightly ~]# restorecon -v /var/log/redis/
restorecon reset /var/log/redis context unconfined_u:object_r:var_log_t:s0->unconfined_u:object_r:redis_log_t:s0
```

After running the restorecon, suddenly everything is fine:

```
[root@centos7-foreman-nightly ~]# systemctl start rh-redis5-redis
[root@centos7-foreman-nightly ~]# systemctl status rh-redis5-redis
● rh-redis5-redis.service - Redis persistent key-value database
   Loaded: loaded (/usr/lib/systemd/system/rh-redis5-redis.service; disabled; vendor preset: disabled)
  Drop-In: /etc/systemd/system/rh-redis5-redis.service.d
           └─limit.conf
   Active: active (running) since Wed 2020-09-30 11:42:27 UTC; 7s ago
 Main PID: 12012 (redis-server)
    Tasks: 4
   CGroup: /system.slice/rh-redis5-redis.service
           └─12012 /opt/rh/rh-redis5/root/usr/bin/redis-server 127.0.0.1:6379

Sep 30 11:42:27 centos7-foreman-nightly.mercury.example.com systemd[1]: Starting Redis persistent key-value database...
Sep 30 11:42:27 centos7-foreman-nightly.mercury.example.com systemd[1]: Started Redis persistent key-value database.
```

Since the same behavior was observed at the baseos level when using `mkdir` instead of `mkdir -Z`, and the dir does get created by puppet-redis at https://github.com/voxpupuli/puppet-redis/blob/master/manifests/config.pp#L15-L20 my best guess is that Puppet's file provider should be using the `-Z` option when selinux is enabled but does not. So that issue needs to get fixed in Puppet, but in the meantime we also need a workaround here.

wbclark: With that said, idempotence is a problem and it does indeed break on upgrades so I'll need to find a different solution.

ehelms: When you run into this, has the system had a full 'yum update' before hand?

ekohl: Do you have the detailed Puppet log from the installer? It should log which matchcon rules it found. Normally it does look up the rules but perhaps the correct rules were not present at the start of the run?

wbclark: @ehelms The error still occurred when I performed a full `yum update` before installation.

wbclark: @ekohl the logs don't reference any kind of matchpathcon lookup that I can see:

```
[DEBUG 2020-10-06T19:07:29 main] Executing: '/bin/yum -d 0 -e 0 -y install foreman-dynflow-sidekiq'
[ WARN 2020-10-06T19:07:32 main] /Stage[main]/Foreman::Install/Package[foreman-dynflow-sidekiq]/ensure: created
[DEBUG 2020-10-06T19:07:32 main] /Package[foreman-dynflow-sidekiq]: The container Class[Foreman::Install] will propagate my refresh event
[DEBUG 2020-10-06T19:07:32 main] Class[Foreman::Install]: The container Stage[main] will propagate my refresh event
[ INFO 2020-10-06T19:07:32 main] Class[Foreman::Install]: Scheduling refresh of Class[Foreman::Config]
[ INFO 2020-10-06T19:07:32 main] Class[Foreman::Install]: Scheduling refresh of Class[Foreman::Service]
[ INFO 2020-10-06T19:07:32 main] Class[Foreman::Install]: Scheduling refresh of Class[Foreman::Service]
[ WARN 2020-10-06T19:07:32 main] /Stage[main]/Redis::Config/File[/etc/opt/rh/rh-redis5/redis]/ensure: created
[DEBUG 2020-10-06T19:07:32 main] /Stage[main]/Redis::Config/File[/etc/opt/rh/rh-redis5/redis]: The container Class[Redis::Config] will propagate my refresh event
[ WARN 2020-10-06T19:07:32 main] /Stage[main]/Redis::Config/File[/var/log/redis]/ensure: created
[DEBUG 2020-10-06T19:07:32 main] /Stage[main]/Redis::Config/File[/var/log/redis]: The container Class[Redis::Config] will propagate my refresh event
[ WARN 2020-10-06T19:07:32 main] /Stage[main]/Redis::Ulimit/File[/etc/systemd/system/rh-redis5-redis.service.d/limit.conf]/mode: mode changed '0644' to '0444'
[DEBUG 2020-10-06T19:07:32 main] /Stage[main]/Redis::Ulimit/File[/etc/systemd/system/rh-redis5-redis.service.d/limit.conf]: The container Class[Redis::Ulimit] will propagate my refresh event
[DEBUG 2020-10-06T19:07:32 main] Augeas[Systemd redis ulimit](provider=augeas): Opening augeas with root /, lens path , flags 64
```

(Excerpt from `/usr/share/foreman-installer/_build/foreman.log`)

wbclark: Interestingly, the behavior is different when using the foreman-installer RPM:

```
[root@centos7-foreman-nightly ~]# ll -dZ /var/log/redis/
drwxr-x---. redis redis system_u:object_r:redis_log_t:s0 /var/log/redis/
[root@centos7-foreman-nightly ~]# grep '/var/log/redis' /var/log/foreman-installer/foreman.log 
[DEBUG 2020-10-06T21:55:54 main] /File[/var/log/redis]/seluser: Found seluser default 'system_u' for /var/log/redis
[DEBUG 2020-10-06T21:55:54 main] /File[/var/log/redis]/selrole: Found selrole default 'object_r' for /var/log/redis
[DEBUG 2020-10-06T21:55:54 main] /File[/var/log/redis]/seltype: Found seltype default 'redis_log_t' for /var/log/redis
[DEBUG 2020-10-06T21:55:54 main] /File[/var/log/redis]/selrange: Found selrange default 's0' for /var/log/redis
[ WARN 2020-10-06T21:56:41 main] /Stage[main]/Redis::Config/File[/var/log/redis]/ensure: created
[DEBUG 2020-10-06T21:56:41 main] /Stage[main]/Redis::Config/File[/var/log/redis]: The container Class[Redis::Config] will propagate my refresh event
```

In that case I do see the proper lookup for seltype before the dir is created by puppet.

ekohl: The foreman-installer RPM depends on foreman-selinux which depends on a fairly recent selinux-policy. Perhaps that is a difference because it already loads various rules prior to running the installer itself? Just a theory.

wbclark: I'll investigate that possibility. Another thought is to check whether there are any changes in puppet-redis, which I'll check as well.

wbclark: Making note before I reprovision, when installed from installer RPM we get:

```
# grep '"version":' /usr/share/foreman-installer/modules/redis/metadata.json 
  "version": "6.1.0",
```

wbclark: puppet-redis version was the same when built from source:

```
[root@centos7-foreman-nightly ~]# grep '"version":' /usr/share/foreman-installer/_build/modules/redis/metadata.json 
  "version": "6.1.0",
```

I can also confirm that I saw the same incorrect behavior when I pre-installed the foreman-selinux RPM.

```
[root@centos7-foreman-nightly ~]# ll -dZ /var/log/redis/
drwxr-x---. redis redis unconfined_u:object_r:var_log_t:s0 /var/log/redis/
```
And no lookup in the installer logs:

```
[root@centos7-foreman-nightly ~]# grep -C1 '/var/log/redis' /usr/share/foreman-installer/_build/foreman.log 
[DEBUG 2020-10-07T14:41:15 main] /Stage[main]/Redis::Config/File[/etc/opt/rh/rh-redis5/redis]: The container Class[Redis::Config] will propagate my refresh event
[ WARN 2020-10-07T14:41:15 main] /Stage[main]/Redis::Config/File[/var/log/redis]/ensure: created
[DEBUG 2020-10-07T14:41:15 main] /Stage[main]/Redis::Config/File[/var/log/redis]: The container Class[Redis::Config] will propagate my refresh event
[ WARN 2020-10-07T14:41:15 main] /Stage[main]/Redis::Ulimit/File[/etc/systemd/system/rh-redis5-redis.service.d/limit.conf]/mode: mode changed '0644' to '0444'
```

ekohl:
In the Redis module's git master version we changed the logging path to the correct SCL path. I'd prefer to get that released and see if it fixes the problem.

wbclark:
> In the Redis module's git master version we changed the logging path to the correct SCL path. I'd prefer to get that released and see if it fixes the problem.

OK, makes sense to me.

ekohl:
Should we close this or is it still an issue?

ekohl:
Since https://github.com/voxpupuli/puppet-redis/pull/379 was merged, we should be using the correct log path.

PR #581, opened 2020-09-23
move migrate_cdn_setting to hooks/

wbclark:


ehelms:
There is a pre hook component associated with this hook as well.

ehelms:
@wbclark did you find out form @chris1984 or @jlsherrill if this is a hook we still need?

chris1984:
I think we still need it since this was a migration from an installer parameter to one in the UI and we call a rake command to update that in the DB.

chris1984:
I think @ekohl said it was safe to drop it at some point down the road but not sure when we should make that call?

ehelms:
In what version of Katello was this change made? @chris1984

chris1984:
@ehelms looks like 3.16.0 https://projects.theforeman.org/issues/28476

wbclark:
Updated the move the pre_commit hook as well

ehelms (approved):

PR #582, opened 2020-09-23
move set_umask to hooks/

wbclark:


ekohl:
I wonder if this file is still needed. IMHO it was always a workaround for failing to specify correct permissions in modules. I'd consider to remove it

ehelms:
@ekohl is there a way to verify we can remove it? or is it a drop and see what happens?

ekohl:
Back in the day we had a pipeline with umask 077 for the regular installer to verify. I think that would be a good test.

ehelms:
Can we test this at a more granular level or a simpler way without having to have an entire pipeline run dedicated to it?

ekohl:
We've been talking about idempotency tests. You could do a base run, run it again (and assert nothing changes) and then run it with a different umask (again, asserting nothing changes). However, I don't think that actually works. umask is used when creating resources.

A more granular approach would be to have acceptance tests in every module which run with a different umask.

wbclark:
I need to pick back up the work on https://github.com/theforeman/forklift/pull/1208 to make it work with the Katello scenario, but in that case it would provide a method to test whether modules have all necessary permissions configured correctly with this hook removed.

ehelms:
This file has been removed.

PR #583, opened 2020-10-05
Refs #30423 - Migrate Pulp media directory

ekohl:
The Pulp media root is being changed from /var/lib/pulp to /var/lib/pulp/media. This migration moves the existing content in /var/lib/pulp/artifact.

Pulp only really uses $MEDIA_ROOT/artifact but various tools that can do consistency checks use the entire MEDIA_ROOT.

These hooks are implemented in 4 stages, 3 of which in hooks.

* pre_validations - Checks if a migration would be a problem
* pre - Moves the directory and creates a symlink
* [Puppet - Updates the config and restarts services](https://github.com/theforeman/puppet-pulpcore/pull/115)
* post - Cleans up the old symlink

This should only be merged after the [Puppet change](https://github.com/theforeman/puppet-pulpcore/pull/115).

### DIFF
--- a/hooks/post/34-pulpcore_directory_layout.rb
+++ b/hooks/post/34-pulpcore_directory_layout.rb
@@ -1,0 +1,8 @@
+require 'pathname'
+
+LEGACY_DIR = Pathname.new('/var/lib/pulp/artifact')
+
+if LEGACY_DIR.symlink?
+  logger.debug("Removing legacy symlink #{LEGACY_DIR}")
+  LEGACY_DIR.unlink unless app_value(:noop)
+end
--- a/hooks/pre/34-pulpcore_directory_layout.rb
+++ b/hooks/pre/34-pulpcore_directory_layout.rb
@@ -1,0 +1,15 @@
+require 'pathname'
+
+PULP_ROOT = Pathname.new('/var/lib/pulp')
+LEGACY_DIR = PULP_ROOT / 'artifact'
+NEW_MEDIA_ROOT = PULP_ROOT / 'media'
+DESTINATION = NEW_MEDIA_ROOT / LEGACY_DIR.basename
+
+if LEGACY_DIR.directory? && !LEGACY_DIR.symlink?
+  logger.debug("Migrating #{LEGACY_DIR} to #{DESTINATION}")
+  unless app_value(:noop)
+    NEW_MEDIA_ROOT.mkpath
+    LEGACY_DIR.rename(DESTINATION)
+    LEGACY_DIR.make_symlink(DESTINATION)
+  end
+end
--- a/hooks/pre_validations/34-pulpcore_directory_layout.rb
+++ b/hooks/pre_validations/34-pulpcore_directory_layout.rb
@@ -1,0 +1,14 @@
+require 'pathname'
+
+PULP_ROOT = Pathname.new('/var/lib/pulp')
+LEGACY_DIR = PULP_ROOT / 'artifact'
+DESTINATION = PULP_ROOT / 'media' / LEGACY_DIR.basename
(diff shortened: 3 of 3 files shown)


hooks/post/34-pulpcore_directory_layout.rb
```diff
@@ -0,0 +1,8 @@
+unless app_value(:noop)
+  require 'pathname'
+
+  LEGACY_DIR = Pathname.new('/var/lib/pulp/artifact')
+
+  logger.debug("Removing legacy symlink #{LEGACY_DIR}")
+  LEGACY_DIR.unlink if LEGACY_DIR.symlink?
+end
```
ehelms, line 8: I am not asking for implementation here. But I think this provides a good opportunity to ask a question regrading noop. If we were aiming for true noop and reporting of what *would* change. Should we look in our hooks to present information about what would have been done? For example, detecting that it is a symlink and informing the user this would unlink it.

ekohl: I think that's a fair question and I would actually consider it in this PR. Abusing noop in tests and checking logs is actually often a decent way to test code paths.

hooks/pre_validations/34-pulpcore_directory_layout.rb
```diff
@@ -0,0 +1,9 @@
+require 'pathname'
+
+PULP_ROOT = Pathname.new('/var/lib/pulp')
+LEGACY_DIR = PULP_ROOT / 'artifact'
+DESTINATION = PULP_ROOT / 'media' / LEGACY_DIR.basename
+
+if LEGACY_DIR.directory? && !LEGACY_DIR.symlink? && DESTINATION.directory?
+  fail_and_exit("Target #{DESTINATION} already exists. Unable to migrate Pulp media root.")
```
ehelms: What action should a user take when getting this message?

hooks/pre/34-pulpcore_directory_layout.rb
```diff
@@ -0,0 +1,15 @@
+unless app_value(:noop)
+  require 'pathname'
+
+  PULP_ROOT = Pathname.new('/var/lib/pulp')
+  LEGACY_DIR = PULP_ROOT / 'artifact'
+  NEW_MEDIA_ROOT = PULP_ROOT / 'media'
+  DESTINATION = NEW_MEDIA_ROOT / LEGACY_DIR.basename
```
ehelms: Could we store these directories in a single place for reference?

ekohl: At this point I think it would only add confusion because you need to look at other files to see what's going on.

hooks/post/34-pulpcore_directory_layout.rb
ekohl: Updated to only skip the actual changes.

hooks/pre_validations/34-pulpcore_directory_layout.rb
ehelms: My last remaining question before merging.

ekohl:
https://github.com/theforeman/foreman-installer/pull/589 would be a way to avoid duplicating the constants. Please take a look and see if it makes it more or less understandable.

On the plus side, it's easier to follow the individual hooks but it may be harder to lose track of all hooks and the order they're executed.

hooks/pre_validations/34-pulpcore_directory_layout.rb
ekohl: How about now?

ehelms (approved):
Updated message looks good as it provides user with an action.

PR #584, opened 2020-10-07
Fixes #31022 - Use an isolated Puppet environment

ekohl:
When there's splay=true in /etc/puppetlabs/puppet/puppet.conf, puppet apply waits a random period (up to 30 minutes). Kafo deals with this by using --config=/tmp/.../puppet.conf which is generated in the ExecutionEnvironment. The boot hook apply_puppet_code does not use this which causes the host's puppet.conf to take effect. This can lead to much longer installer times where it just sleeps.

This patch reuses Kafo::ExecutionEnvironment for an isolated environment.

I have not tested this patch yet, but it's based on https://community.theforeman.org/t/puppet-with-splay-true/20777 as a report.

### DIFF
--- a/hooks/boot/01-kafo-hook-extensions.rb
+++ b/hooks/boot/01-kafo-hook-extensions.rb
@@ -1,4 +1,5 @@
 require 'English'
+require 'fileutils'
 require 'open3'
 
 module HookContextExtension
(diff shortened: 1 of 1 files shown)


ehelms:
Generally, this looks like a good change. Since I am working a lot in Kafo code would this be worth while to add to the hook context as a method?

ekohl:
Yes, that would be much cleaner. https://github.com/theforeman/kafo/pull/284 would do that. I'm not sure what would be best considering cherry picks so I'm leaving this open for now.

ekohl:
https://github.com/theforeman/foreman-installer/pull/585 for comparison.

ehelms:
@ekohl I am OK with this change as is, if you want to come back later and update it once the Kafo changes have been released.

PR #587, opened 2020-10-10
Move messages to exit hook

ehelms:
Moves message display to the new exit hook that will run for any success
or error.

Uses exit hook introduced by https://github.com/theforeman/kafo/pull/287 and updates the messaging of the failure based on discussions from https://github.com/theforeman/kafo/pull/278#issuecomment-705713775

Example when error occurs before puppet:

```
2020-10-10 19:51:05 [INFO ] [pre_commit] Executing hooks in group pre_commit
2020-10-10 19:51:05 [INFO ] [pre_commit] All hooks in group pre_commit finished
2020-10-10 19:51:05 [INFO ] [pre] Executing hooks in group pre
Puppet agent oauth failed to install
2020-10-10 19:51:05 [ERROR] [root] Puppet agent oauth failed to install

  There were errors detected during install.
  Please address the errors and re-run the installer to ensure the system is properly configured.
  Failing to do so is likely to result in broken functionality.

  The full log is at ./_build/foreman.log
```

Example after puppet:

```
2020-10-10 20:03:29 [ERROR] [configure] Could not find a suitable provider for augeas
2020-10-10 20:03:29 [ERROR] [configure] Could not find a suitable provider for foreman_smartproxy
2020-10-10 20:03:29 [WARN ] [configure] Applied catalog in 13.25 seconds
2020-10-10 20:03:30 [INFO ] [configure] Puppet has finished, bye!
2020-10-10 20:03:30 [INFO ] [post] Executing hooks in group post
Executing: foreman-rake upgrade:run
foreman-rake upgrade:run finished successfully!
2020-10-10 20:03:35 [INFO ] [post] All hooks in group post finished

  There were errors detected during install.
  Please address the errors and re-run the installer to ensure the system is properly configured.
  Failing to do so is likely to result in broken functionality.

  The full log is at ./_build/foreman.log
```
```

### CODE REVIEW

**Comment by ehelms** [comment]
Moved the messaging portion of this PR to -- https://github.com/theforeman/foreman-installer/pull/592

**Review by ekohl — COMMENTED** [review]
My initial reaction is :-1: on this. Post is only executed after Puppet was at least called. Exit is executed every time you exit. That means also if `fail_and_exit` is called. It doesn't look like an improvement to me.

**Comment by ekohl** [comment]
> Executing: foreman-rake upgrade:run
> foreman-rake upgrade:run finished successfully!

This part shouldn't run if Puppet didn't finish successfully. That in itself is a bug IMHO.

**Comment by ehelms** [comment]
> That means also if `fail_and_exit` is called.

Exactly one of the intents! If you call fail_and_exit and you have *defined* any exit handling, which in this case  is *outputting* a message to the user regarding that an error was hit and what you should do about it a user can and should see it anytime fail_and_exit is called.

File: hooks/boot/02-message-helpers.rb [hunk 1]
```diff
@@ -38,11 +38,13 @@ def dev_new_install_message(kafo)
   end
 
   def failure_message
-    say "  <%= color('Something went wrong!', :bad) %> Check the log for ERROR-level output"
+    say "\n  <%= color('There were errors detected during install.', :bad) %>"
+    say "  Please address the errors and re-run the installer to ensure the system is properly configured."
```
wbclark, line 42:
```suggestion
    say "\n  There were <%= color('ERROR', :bad) %>s detected during install."
    say "  Please address the <%= color('ERROR', :bad) %>s and re-run the installer to ensure the system is properly configured."
```

I want to make this scream "hey you! you need to do something about this!" as much as possible. What do you think @ehelms ? Capitalization is also consistent with how the log level is displayed.

wbclark:
I see that the message was already updated with #592 

So after the upcoming Kafo 6.0 release the remaining change to use the `pre_exit` hook here could be revisited.

PR #589, opened 2020-10-11
Demonstrate Kafo multi-hook

ekohl:
The Pulp media root is being changed from /var/lib/pulp to /var/lib/pulp/media. This migration moves the existing content in /var/lib/pulp/artifact.

Pulp only really uses $MEDIA_ROOT/artifact but various tools that can do consistency checks use the entire MEDIA_ROOT.

These hooks are implemented in 4 stages, 3 of which in hooks.

* pre_validations - Checks if a migration would be a problem
* pre - Moves the directory and creates a symlink
* Puppet - Updates the config and restarts services
* post - Cleans up the old symlink

It's an alternative to https://github.com/theforeman/foreman-installer/pull/583 to demonstrate [Kafo's new multi hook](https://github.com/theforeman/kafo/pull/288) approach to avoid duplicating constants.

### DIFF
--- a/hooks/multi/34-pulpcore_directory_layout.rb
+++ b/hooks/multi/34-pulpcore_directory_layout.rb
@@ -1,0 +1,30 @@
+require 'pathname'
+
+PULP_ROOT = Pathname.new('/tmp/pulp')
+LEGACY_DIR = PULP_ROOT / 'artifact'
+NEW_MEDIA_ROOT = PULP_ROOT / 'media'
+DESTINATION = NEW_MEDIA_ROOT / LEGACY_DIR.basename
+
+pre_validations do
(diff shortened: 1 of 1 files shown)


ehelms:
@ekohl Do we need this migration for Foreman 2.3?

ekohl:
Yes

Edit: spoke too soon - not really. I thought this was a different PR.

ehelms:
Thanks... I should have also asked, do we need this at all anymore? Not needing it for Foreman 2.3 sounds like a nice to have more than a needful thing?

ekohl:
The actual migration was merged in #583 and this is just a demo of the multi hook. Kafo first needs to support that, then we can rewrite hooks if we want.

ehelms:
Thanks, updating the PR title to reflect that for ease of looking through PR list.

PR #590, opened 2020-10-12
Fixes #31051 - Add verification to check if SAN entries match Subject CN on certificate

jpasqualetto:
Add verification to check if SAN entries match Subject CN on certificate

### DIFF
--- a/bin/katello-certs-check
+++ b/bin/katello-certs-check
@@ -175,12 +175,29 @@ function check-ca-bundle-size () {
 
 function check-cert-san () {
     printf "Checking Subject Alt Name on certificate "
-    CHECK=$(openssl x509 -noout -text -in $CERT_FILE | grep DNS:)
+    DNS_LIST=$(openssl x509 -noout -text -in $CERT_FILE | grep DNS:)
     if [[ $? == "0" ]]; then
         success
+        printf "Checking if any Subject Alt Name on certificate matches the Subject CN"
+        SUBJECT_CN=$(openssl x509 -in $CERT_FILE -noout -subject -nameopt multiline|grep commonName|cut -d '=' -f 2|tr -d ' ')
+        for DNS in ${DNS_LIST}
+          do
+          DNS_VALUE="$( echo ${DNS//DNS:/} | tr -d ',')"
+          if [ $SUBJECT_CN == $DNS_VALUE ]
+          then
+            success
+            return
+          fi
+        done
+        error
+        echo "The $CERT_FILE does not have a Subject Alt Name matching the Subject CN"
     else
-        warning
-        echo "The $CERT_FILE does not contain a Subject Alt Name."
+        error
+        cat <<Explanation
+$CERT_FILE does not contain a Subject Alt Name. Common Name is deprecated, use Subject Alt Name instead.
+See: https://tools.ietf.org/html/rfc2818#section-3.1
+
+Explanation
     fi
 }
 


bin/katello-certs-check
```diff
@@ -178,6 +178,20 @@ function check-cert-san () {
     CHECK=$(openssl x509 -noout -text -in $CERT_FILE | grep DNS:)
```
wbclark:
```suggestion
    DNS_LIST="$(openssl x509 -noout -text -in $CERT_FILE | grep DNS:)"
```

bin/katello-certs-check
```diff
@@ -178,6 +178,20 @@ function check-cert-san () {
...
     if [[ $? == "0" ]]; then
         success
+        printf "Checking if any of Subject Alt Names on certificate matches the Subject CN"
+        SUBJECT_CN=$(openssl x509 -in $CERT_FILE -noout -subject -nameopt multiline|grep commonName|cut -d '=' -f 2|tr -d ' ')
+        DNS_LIST="$(openssl x509 -noout -text -in $CERT_FILE | grep DNS:)"
```
wbclark:
```suggestion
```

We can already get this info from L#178 (see above suggestion)

bin/katello-certs-check
```diff
@@ -178,6 +178,20 @@ function check-cert-san () {
     CHECK=$(openssl x509 -noout -text -in $CERT_FILE | grep DNS:)
     if [[ $? == "0" ]]; then
         success
+        printf "Checking if any of Subject Alt Names on certificate matches the Subject CN"
```
wbclark:
```suggestion
        printf "Checking if any Subject Alt Name on certificate matches the Subject CN"
```

Nitpick: this wording is clearer I think

wbclark (commented):
Hi @jpasqualetto , thanks for filing the issue and opening this PR!

I left a few comments. Please take a look and let me know what you think.

bin/katello-certs-check
jpasqualetto: That's fine. I was not 100% comfortable with my initial wording.

bin/katello-certs-check
jpasqualetto: Thanks for the suggestions. I'll commit the changes.

wbclark:
I thought it was odd that we only give a warning when no SAN is present at all, but we should give an error when SANs are present but none matches the cname. If it is required to have at least one SAN matching the cname, shouldn't we also return an error when there is no SAN in the first place?

So I did some tests.

I. First I created a custom CA, cert, and key with the cname `wclark-custom-ssl.example.com` and no SAN. Then I ran `katello-certs-check` and saw the expected warning. Then I installed katello-3.14.0 (i.e. Satellite 6.7) with my custom cert and CA, and found that the installer completed successfully and hammer ping gives the expected output.

II. Then I reset the installation and created a new cert and key with the same cname `wclark-custom-ssl.example.com` but this time added a single typo'd SAN `wclark-customer-ssl.example.com`. Due to the presence of the SAN, katello-certs-check completed without any warning. But I was surprised to find that the installer also completed successfully in this case.

Confirming the CN and SAN I had in my second test:

```
# CERT_FILE=/etc/pki/mycerts/wclark-custom-ssl.example.com.crt
# openssl x509 -in $CERT_FILE -noout -subject -nameopt multiline|grep commonName|cut -d '=' -f 2|tr -d ' '
wclark-custom-ssl.example.com
# openssl x509 -noout -text -in $CERT_FILE | grep DNS:
                DNS:wclark-customer-ssl.example.com
#
```

Now after running the installer and confirming the application is working on a basic level (i.e. `hammer ping` returns success), I perform the same check against `/etc/pki/katello/certs/katello-apache.crt` and find that the SAN isn't present!

```
# CERT_FILE=/etc/pki/katello/certs/katello-apache.crt 
# openssl x509 -in $CERT_FILE -noout -subject -nameopt multiline|grep commonName|cut -d '=' -f 2|tr -d ' '
wclark-custom-ssl.example.com
# openssl x509 -noout -text -in $CERT_FILE | grep DNS:
# 
```

So I wasn't able to reproduce an installation error on this version, whether using no SAN present at all, or a single typo'd SAN that does not match the CN. And my typo'd SAN even appeared to get removed from the `katello-apache.crt` although I did not check all other certs used by the application.

Anyway, I thought you would find this interesting @jpasqualetto , and I wanted to ask if you could share some more information about how you reproduced the issue and what exactly were the errors that you saw. Further, do you see any reason why the behavior should be different in the case of no SAN on the cert vs. SAN on the cert but not matching cname (getting at this question was my ultimate goal in performing these tests, even though I couldn't reproduce an issue in either case).

Looking forward to hearing from you.

jpasqualetto:
Hi @wbclark 

> So I wasn't able to reproduce an installation error on this version, whether using no SAN present at all

A certificate with no SAN should not fail the installation. This behavior seems expected to me. The warning when checking the certificates is good enough in my opinion. See below snippet from RFC2818, about using CN as identity source is deprecated (but known to be used by many clients, yet).

> or a single typo'd SAN that does not match the CN. And my typo'd SAN even appeared to get removed from the katello-apache.crt although I did not check all other certs used by the application.

That's surprising. Here's the certificate I used to reproduce an installation error:

~~~
# openssl x509 -noout -text -in /root/EasyRSA-3.0.8/pki/issued/sat67.jpasqualetto.local.crt |grep -e DNS -e Subject
        Subject: CN=sat67.jpasqualetto.local                     <==== CN with my server's hostname
        Subject Public Key Info:
            X509v3 Subject Key Identifier: 
            X509v3 Subject Alternative Name: 
                DNS:sat67.jpasqualetto.local.with-typo            <==== SAN not matching the CN
~~~

This certificate passes the katello-cert-check with no errors or warnings.

When running satellite-installer to deploy it, it will fail in one step querying the smart proxies:

~~~
[DEBUG 2020-10-14T13:48:23 main]  Foreman_smartproxy[sat67.jpasqualetto.local](provider=rest_v3): Making get request to https://sat67.jpasqualetto.local/api/v2/smart_proxies?search=name=%22sat67.jpasqualetto.local%22
[ERROR 2020-10-14T13:48:23 main]  /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[sat67.jpasqualetto.local]: Failed to call refresh: Exception SSL_connect returned=1 errno=0 state=error: certificate verify failed in get request to: https://sat67.jpasqualetto.local/api/v2/smart_proxies?search=name=%22sat67.jpasqualetto.local%22
~~~

Using curl to query apache in the https port will also fail:

~~~
# curl https://$(hostname -f) -v
* About to connect() to sat67.jpasqualetto.local port 443 (#0)
*   Trying 192.168.250.100...
* Connected to sat67.jpasqualetto.local (192.168.250.100) port 443 (#0)
* Initializing NSS with certpath: sql:/etc/pki/nssdb
*   CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
* Server certificate:
* 	subject: CN=sat67.jpasqualetto.local
* 	start date: Oct 14 17:41:56 2020 GMT
* 	expire date: Jan 17 17:41:56 2023 GMT
* 	common name: sat67.jpasqualetto.local
* 	issuer: CN=Easy-RSA CA
* NSS error -12276 (SSL_ERROR_BAD_CERT_DOMAIN)
* Unable to communicate securely with peer: requested domain name does not match the server's certificate.
* Closing connection 0
curl: (51) Unable to communicate securely with peer: requested domain name does not match the server's certificate.
~~~

The reason for this fail is that, when a certificate contains any SAN, the SAN must be used as the source of the server's identify, ignoring the CN.  See this snippet from [RFC2818]

~~~
3.  Endpoint Identification

3.1.  Server Identity

   In general, HTTP/TLS requests are generated by dereferencing a URI.
   As a consequence, the hostname for the server is known to the client.
   If the hostname is available, the client MUST check it against the
   server's identity as presented in the server's Certificate message,
   in order to prevent man-in-the-middle attacks.

   If the client has external information as to the expected identity of
   the server, the hostname check MAY be omitted. (For instance, a
   client may be connecting to a machine whose address and hostname are
   dynamic but the client knows the certificate that the server will
   present.) In such cases, it is important to narrow the scope of
   acceptable certificates as much as possible in order to prevent man
   in the middle attacks.  In special cases, it may be appropriate for
   the client to simply ignore the server's identity, but it must be
   understood that this leaves the connection open to active attack.

   If a subjectAltName extension of type dNSName is present, that MUST
   be used as the identity. Otherwise, the (most specific) Common Name
   field in the Subject field of the certificate MUST be used. Although
   the use of the Common Name is existing practice, it is deprecated and
   Certification Authorities are encouraged to use the dNSName instead.
~~~

Let me know if you need anything else. 

[RFC2818]: https://tools.ietf.org/html/rfc2818

bin/katello-certs-check
```diff
@@ -175,9 +175,22 @@ function check-ca-bundle-size () {
...
+          then
+            success
+            return
+          fi
+        done
+        error
+        echo "The $CERT_FILE does not have a Subject Alt Name matching the Subject CN"
     else
         warning
         echo "The $CERT_FILE does not contain a Subject Alt Name."
```
wbclark:
```suggestion
        cat <<WarningMessage
$CERT_FILE does not contain a Subject Alt Name. Common Name is deprecated, use Subject Alt Name instead.
See: https://tools.ietf.org/html/rfc2818#section-3.1
WarningMessage
```

Thanks @jpasqualetto . I think the RFC also answers the other question that I had, which is whether we should get rid of the `warning` here and raise an `error` instead.

But based on your info, let's keep the `warning` and provide a better explanation about it to the user.

wbclark (approved):
@jpasqualetto overall it looks good to me.

Let me know if you'd like to update the warning message as well (see above comment) before merging, or I can do it in a separate PR later.

Thanks for the contribution :+1:

jpasqualetto:
I'm updating the warning message in this PR.

bin/katello-certs-check
```diff
@@ -175,12 +175,29 @@ function check-ca-bundle-size () {
...
+          if [ "$SUBJECT_CN" = "$DNS_VALUE" ]
+          then
+            success
+            return
+          fi
+        done
+        error
+        echo "The $CERT_FILE does not have a Subject Alt Name matching the Subject CN"
     else
         warning
```
ekohl: Should this be turned into an error? I think modern browsers will soon start to reject certs without a SAN. https://stackoverflow.com/questions/43665243/invalid-self-signed-ssl-cert-subject-alternative-name-missing suggests this is already the case for Chrome and Safari.

bin/katello-certs-check
```diff
@@ -175,12 +175,29 @@ function check-ca-bundle-size () {
...
     printf "Checking Subject Alt Name on certificate "
-    CHECK=$(openssl x509 -noout -text -in $CERT_FILE | grep DNS:)
+    DNS_LIST=$(openssl x509 -noout -text -in $CERT_FILE | grep DNS:)
     if [[ $? == "0" ]]; then
         success
+        printf "Checking if any Subject Alt Name on certificate matches the Subject CN"
+        SUBJECT_CN=$(openssl x509 -in $CERT_FILE -noout -subject -nameopt multiline|grep commonName|cut -d '=' -f 2|tr -d ' ')
+        for DNS in ${DNS_LIST}
+          do
+          DNS_VALUE="$( echo ${DNS//DNS:/} | tr -d ',')"
```
ekohl, line 185: If you want to be correct, it should ignore SANs with the `IP:` prefix and only deal with `DNS:`.

bin/katello-certs-check
```diff
@@ -175,12 +175,29 @@ function check-ca-bundle-size () {
...
+          if [ "$SUBJECT_CN" = "$DNS_VALUE" ]
```
ekohl: I always prefer bash comparisons when it comes to comparing strings. Not sure how difference there really is.
```suggestion
          if [[ $SUBJECT_CN == $DNS_VALUE ]]
```

bin/katello-certs-check
wbclark: This is already the case due to how `DNS_LIST=$(openssl x509 -noout -text -in $CERT_FILE | grep DNS:)` is defined

ehelms:
This fails because it actually exposes our current certificate fixtures do not have a SAN entry. To fix that, I have opened https://github.com/theforeman/foreman-installer/pull/605 and tested it alongside this PR.

wbclark:
Closing and re-opening to trigger tests to run again

wbclark:
Ah #605 isn't merged yet and that has its own test failures -_-

wbclark:
[test foreman-installer]

PR #593, opened 2020-10-14
Enable verbose output by default

ehelms:


ehelms:
This does not actually work. Currently, any option that is from the set of app options in Kafo such as verbose is not looked at from the config file and is only taken as input from the command line. I am looking at what would take to switch to that.

ehelms:
See https://github.com/theforeman/kafo/pull/291 for how this would have to be implemented

wbclark (commented):
shouldn't the migrations set the `verbose_log_level` as well?

wbclark (approved):

PR #594, opened 2020-10-14
Release 2.2.0-rc4

ehelms:


ehelms:
@tbrisker here ya go

tbrisker (changes requested):
Thanks, but this is missing a foreman_proxy release, we need 15.2.0+the two ACL changes after it (but not the puppetrun removal)

Puppetfile
```diff
@@ -10,7 +10,7 @@ mod 'katello/pulp',                  '>= 8.1.0 < 8.2.0'
 mod 'theforeman/pulpcore',           '>= 1.2.0 < 1.3.0'
 mod 'katello/qpid',                  '>= 6.1.0 < 6.2.0'
 mod 'theforeman/foreman',            '>= 15.1.0 < 15.2.0'
-mod 'theforeman/foreman_proxy',      '>= 15.0.0 < 15.1.0'
+mod 'theforeman/foreman_proxy',      '>= 15.2.0 < 15.3.0'
```
tbrisker, line 13: You'll need to also update this at the bottom of `Puppetfile.lock`

tbrisker:
Failures due to `spec/fixtures/katello-certs-check/foreman.example.com.crt` being expired, we should probably fix that

PR #595, opened 2020-10-20
Drop rackspace compute resource

ehelms:


ekohl (approved):
Failure due to the expired cert, solved in another PR.

@tbrisker this should also be picked into 2.2 (and probably 2.1). Do you prefer a Redmine for that?

tbrisker:
Usually that's good to make sure cherry picks aren't missed when needed, but in this case i just picked it right away:
2.2 - 1f48c65
2.1 - cb8d1ad

PR #600, opened 2020-10-23
Fixes #31135: Introduce execute! and execute for fatal and non-fatal commands

ehelms:


wbclark (changes requested):
`execute_command` and `execute` are used elsewhere in the code.

so rather than changing that behavior with possibly unknown consequences, why don't we change the logic here:

https://github.com/theforeman/foreman-installer/blob/5d22a4c38630c6ff090c32e0926cbd5b0f32f74c/hooks/boot/03-foreman-maintain-extensions.rb#L2-L5

or here:

https://github.com/theforeman/foreman-installer/blob/5d22a4c38630c6ff090c32e0926cbd5b0f32f74c/hooks/boot/09-version_locking.rb#L1-L3

In order to handle the return code of the `foreman-maintain packages -h` command correctly?

ehelms:
The behavior that changes is, execute_command will no longer attempt to make a choice on logging a message in response to the failure or success of the command but rather leave that up to the action that called it. In the long run, that feels better as error messages that are relative to the action itself are better than generic failure messages.

hooks/pre_commit/20-certs_update.rb
```diff
@@ -14,6 +14,6 @@
   key_file  = param('certs', 'server_key').value
 
   unless app_value(:certs_skip_check) || [cert_file, ca_file, key_file].all? { |v| v.to_s.empty? }
-    execute_command(%(katello-certs-check -c "#{cert_file}" -k "#{key_file}" -b "#{ca_file}"))
+    execute(%(katello-certs-check -c "#{cert_file}" -k "#{key_file}" -b "#{ca_file}"))
```
ekohl: Looks like a good change, but unrelated to #31135. Could you submit this with `Refs #31119` perhaps?

ekohl (changes requested):
> The behavior that changes is, execute_command will no longer attempt to make a choice on logging a message in response to the failure or success of the command but rather leave that up to the action that called it. In the long run, that feels better as error messages that are relative to the action itself are better than generic failure messages.

I think at this point that is a big change and hesitant to make it and share @wbclark's concerns. His suggestion sounds better.

hooks/pre_commit/20-certs_update.rb
ehelms: Here - https://github.com/theforeman/foreman-installer/pull/602

ehelms:
The only behavior changing here is how things are logged. And if you look at the use of `execute_command` it is used to run a command and then make a determination of the state based upon the result from running the command. Where as `execute` is intended to make a choice around what should be done if something fails (i.e. `exit 1`). Therefore, I have moved the error logging that existed into `execute` if a failure occurs, and kept the existing debug logging in `execute_command`. That way execute will flag and exit as it does today but also log/say what errored. And execute command will perform the operation, log debug output and return the status code in order for the hook code to make choices.

I am not sure how to fix this otherwise, given the bug itself revolves around a foreman maintain command that we expect to error in some cases but we do not treat it as an error because it indicates information we use. The alternative is to set no_log to true, and then we get no information in the logs at the debug level to be able to trace why something didn't occur.

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -132,19 +133,13 @@ def execute_command(command, do_say, do_log)
     begin
       stdout_stderr, status = Open3.capture2e(*Kafo::PuppetCommand.format_command(command))
     rescue Errno::ENOENT
-      log_and_say(:error, "Command #{command} not found", do_say, do_log)
```
ekohl, line 126: This is no longer implemented

ekohl (commented):
Perhaps it's time to start documenting these methods with comments. I do think it makes sense to have one method that is more like a "frontend" and one more like a "backend" method, but that wasn't immediately clear to me.

Please consider rebasing it on top of https://github.com/theforeman/foreman-installer/pull/601. I think that makes it much easier to restore the previous behavior.

ekohl:
Btw, I would also expect a different summary. When I read the current summary it makes me think it should only touch `foreman-maintain -h` but it's really changing the internal API. That allows you to achieve the foreman-maintain fix, but a good summary sets the right expectation for the reviewer. That explains why both @wbclark and I had objections.

ehelms:
I have instead taken this all the way to re-defining the available commands similar to the set of extensions defined for foreman-maintain commands. This logic can also simplify if we drop the use of the progress bar which is the reason for the `do_say` elements.

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -113,39 +113,40 @@ def log_and_say(level, message, do_say = true, do_log = true)
     Kafo::KafoConfigure.logger.send(level, message) if do_log
   end
 
-  def execute(commands, do_say = true, do_log = true)
-    commands = commands.is_a?(Array) ? commands : [commands]
-    results = []
+  def execute!(commands, do_say = true, do_log = true)
```
ekohl:
```suggestion
  def execute!(command, do_say = true, do_log = true)
```

hooks/boot/03-foreman-maintain-extensions.rb
```diff
@@ -1,7 +1,7 @@
 module ForemanMaintainHookContextExtension
   def foreman_maintain(command)
     command = "foreman-maintain #{command}"
-    execute_command(command, false, true)
+    execute(command, false, true)
   end
 
   def foreman_maintain!(command)
```
ekohl, line 7: This no longer prints what failed if it did.

hooks/pre/31-puppet_agent_oauth.rb
```diff
@@ -1,7 +1,9 @@
 unless app_value(:noop)
-  if File.exist?('/opt/puppetlabs/puppet/bin/ruby') && execute_command("rpm -q puppet-agent-oauth", false, false)
-    unless execute_command("/opt/puppetlabs/puppet/bin/ruby -e \"require 'oauth'\"", false, false)
-      execute_command("yum -y reinstall puppet-agent-oauth", false, true)
+  if File.exist?('/opt/puppetlabs/puppet/bin/ruby') && execute("rpm -q puppet-agent-oauth", false, false)
+    unless execute("/opt/puppetlabs/puppet/bin/ruby -e \"require 'oauth'\"", false, false)
+      success = execute("yum -y reinstall puppet-agent-oauth", false, true)
+
+      logger.error("Failed to install puppet-agent-oauth. Please check that the package is available from a repository.") unless success
```
ekohl:
```suggestion
      logger.error("Failed to reinstall puppet-agent-oauth. Please check that the package is available from a repository.") unless success
```

hooks/pre_commit/20-certs_update.rb
ekohl: Shouldn't this be fatal?
```suggestion
    execute!(%(katello-certs-check -c "#{cert_file}" -k "#{key_file}" -b "#{ca_file}"))
```

hooks/pre/25-remove_apache_from_foreman_group.rb
```diff
@@ -5,7 +5,7 @@ def remove_user_from_group(user, group)
     is_member = false
   end
   if is_member
-    execute("gpasswd -d '#{user}' '#{group}'")
+    execute!("gpasswd -d '#{user}' '#{group}'")
```
ekohl, line 8: I wonder if this is still needed. It was done in 1.23 to fix upgrades, but it's probably safe to drop by now.

ekohl (changes requested):

wbclark (approved):

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -123,20 +138,14 @@ def execute_command(command, do_say, do_log)
     begin
       stdout_stderr, status = Open3.capture2e(*Kafo::PuppetCommand.format_command(command))
     rescue Errno::ENOENT
-      log_and_say(:error, "Command #{command} not found", do_say, do_log)
-      return false
+      return [nil, false]
```
ekohl, line 141: Nit: I think `nil` is better than `false`. From type consistency it then becomes either an object or nil which feels more natural to me.

ekohl (approved):
Small nit, but overall I think this is a good improvement.

hooks/boot/01-kafo-hook-extensions.rb
ehelms: Sorry, I do not understand. The second argument is the status, and the command failed to execute. So the status of it having executed is false. That feels like type consistency.

ekohl: Currently the return type is (in yard style) `Array<(nil, Boolean)>` (command not found) or `Array<(String, Process::Status)>` (command ran). I'm arguing that `Array<(nil, nil)>` makes more sense to me in the former case because it allows cleaner code later on:
```ruby
false&.success? # raises NoMethodError: undefined method `success?' for false:FalseClass
nil&.success? # returns nil
```

It's a nit since this isn't actually used, but since you're creating an API this is something I think about.

An alternative is to return `nil` instead of `[nil, nil]` and capture it that way but I'm not sure if that's a much cleaner API.

ehelms: Ahh yea, thats a bug in the overall methods usage on my part and inconsistency. Now `execute_command` returns `Array(nil|string, Boolean)` consistently and is used as such by the `execute` and `execute!` methods which fixes a possible bug in those hooks using `execute` as well.

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -113,8 +113,23 @@ def log_and_say(level, message, do_say = true, do_log = true)
     Kafo::KafoConfigure.logger.send(level, message) if do_log
   end
 
-  def execute(command, do_say = true, do_log = true)
-    exit 1 unless execute_command(command, do_say, do_log)
+  def execute!(command, do_say = true, do_log = true)
+    stdout_stderr, status = execute_command(command, do_say, do_log)
```
ekohl, line 117: Would this make it clear that it isn't a `Process::Status` object?
```suggestion
    stdout_stderr, success = execute_command(command, do_say, do_log)
```

hooks/boot/01-kafo-hook-extensions.rb
ekohl: That's another way that I'm also fine with. All methods except `success?` are irrelevant anyway for the consumer of `execute_command` since the process has already completed, except perhaps `exitstatus`. Not sure if we may want that in the future.

hooks/boot/01-kafo-hook-extensions.rb
ehelms: You tell me, the current code is already clear to me :)

ehelms: @ekohl what's the final thought here?

ehelms:
Given this has two ACKs and it would benefit to be cherry picked to Foreman 2.3 I am going to merge it.

tbrisker:
2.3 - 858e01e

PR #601, opened 2020-10-23
Drop support for running an array of commands

ekohl:
676111fb58729216c3e06670af2d8e1a6bfe9075 removed the last use of execute with an array. This simplifies the code while remaining compatible.


wbclark (approved):
I checked all cases where we are currently using this method and indeed there are no cases where we are relying on support for an array of commands.

PR #602, opened 2020-10-23
Refs #29803: if katello-certs-check fails the installer should exit

ehelms:
This behavior got changed when the certs hook was moved to the
hooks/ directory and would now only log an error message while
continuing the installer execution to run.

### DIFF
--- a/hooks/pre_commit/20-certs_update.rb
+++ b/hooks/pre_commit/20-certs_update.rb
@@ -14,6 +14,6 @@ if module_enabled?('certs')
   key_file  = param('certs', 'server_key').value
 
   unless app_value(:certs_skip_check) || [cert_file, ca_file, key_file].all? { |v| v.to_s.empty? }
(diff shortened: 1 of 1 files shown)


wbclark:
Looking back at #514 , previously we used:

```
  check_cmd = %(#{CHECK_SCRIPT} -c "#{cert_file}" -k "#{key_file}" -b "#{ca_file}")
  output = `#{check_cmd} 2>&1`
  unless $CHILD_STATUS.success?
    fail_and_exit("Command '#{check_cmd}' exited with #{$CHILD_STATUS.exitstatus}:\n #{output}", 101)
  end
```

I need to look at `execute` vs. `execute_command` again to understand why moving to `execute` would restore the correct behavior.

ekohl:
> I need to look at `execute` vs. `execute_command` again to understand why moving to `execute` would restore the correct behavior.

https://github.com/theforeman/foreman-installer/pull/601 makes the implementation of `execute` a lot more straight forward.

ekohl (approved):

PR #534, opened 2020-06-30
Fixes #30251 - make it easier to brand katello-certs-check

evgeni:

PR #603, opened 2020-10-29
Refs #30251 - load a branding file if it's present

evgeni:


wbclark (approved):
I read over https://projects.theforeman.org/issues/30251 and https://github.com/theforeman/foreman-installer/pull/534 to get a sense of what this should accomplish and it looks good to me.

I suppose the branding file would be delivered in downstream packaging.

evgeni:
> I suppose the branding file would be delivered in downstream packaging.

Correct, that's the idea.

Quick ping to @ehelms if he's OK with the file path I've picked?

ehelms (approved):

ehelms:
[test foreman-installer]

ehelms:
There is a valid Travis run passing.

PR #609, opened 2020-11-04
Fixes #30569 - Add migration to change logging pattern

tbrisker:
(cherry picked from commit 832e1fe70d2400b291ea829c1d0f09ad6641ab61)

### DIFF
--- a/Puppetfile.lock
+++ b/Puppetfile.lock
@@ -95,7 +95,7 @@ FORGE
     theforeman-dns (8.0.0)
       puppetlabs-concat (>= 4.1.0, < 7.0.0)
       puppetlabs-stdlib (>= 4.13.1, < 7.0.0)
-    theforeman-foreman (15.1.1)
+    theforeman-foreman (15.1.2)
       camptocamp-systemd (>= 2.2.0, < 3.0.0)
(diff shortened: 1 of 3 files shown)


ekohl:
The change was only introduced in 16.0.0 but 2.2 includes 15.1.1.

tbrisker:
The new layout and the error page pointing to the rake task that leverages it are in 2.2 core, and i've already seen a few reports on discourse using it which needed to be replied to asking the user to change their layout manually to get the actual stacktrace. can we pick https://github.com/theforeman/puppet-foreman/pull/847 to 15.1?

ekohl:
@wbclark since you're on release rotation this week, mind taking this one?
@tbrisker https://github.com/theforeman/foreman/pull/8126 is IMHO also a good one to introduce.

tbrisker:
opened https://github.com/theforeman/puppet-foreman/pull/902 for the changes needed in puppet-foreman for this (as well as another small change i noticed was missing and probably needed)

tbrisker:
this will fail until 15.1.2 is released with the needed changes

tbrisker:
[test foreman-installer]

PR #611, opened 2020-11-13
Deploy Pulp 3 only Proxy only on EL8

ehelms:


config/foreman-hiera.yaml
```diff
@@ -6,11 +6,14 @@ defaults:
 hierarchy:
   - name: "Custom user answers"
     path: "custom.yaml"
+  - name: "Scenario OS overrides"
+    path: "scenario/%{facts.kafo.scenario.id}/family/%{facts.os.family}-%{facts.os.release.major}.yaml"
+  - name: "Scenario overrides"
+    path: "scenario/%{facts.kafo.scenario.id}.yaml"
```
ekohl: Scenarios should not override answers. It's a bad experience to have exposed parameters that you're not able to use. I think it's time to redesign the classes instead.

I've been thinking for a while about to define `class { 'foreman_proxy::plugin::plugin': ... }` inside the foreman-proxy-content class and not expose it in the installer.

config/foreman.hiera/scenario/foreman-proxy-content/family/RedHat-8.yaml
```diff
@@ -0,0 +1,6 @@
+foreman_proxy::plugin::pulp::pulpcore_enabled: true
+foreman_proxy::plugin::pulp::pulpcore_mirror: true
+foreman_proxy::plugin::pulp::pulpnode_enabled: false
+foreman_proxy::plugin::pulp::enabled: false
+foreman_proxy_content::proxy_pulp_isos_to_pulpcore: true
+foreman_proxy_content::proxy_pulp_yum_to_pulpcore: true
```
ekohl: Should we change the class to always proxy these if Pulp 2 is disabled? The same makes sense on EL7.

ekohl (changes requested):

config/foreman-hiera.yaml
ehelms: For a user, it will seem awkward if they are looking hard enough to try to configure something and it not have affect. Arguably though, there are plenty of parameters that exist today that a user *should not* touch. Whether they are things that will just break, or they are integral to the scenario itself. We have no good answers to that. 

Given this module already includes `foreman_proxy::plugin::pulp` (https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/manifests/init.pp#L176) couldn't we then drop it from the answers and let hiera do the declarations for us per scenario?

ekohl: > Given this module already includes `foreman_proxy::plugin::pulp` (https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/manifests/init.pp#L176) couldn't we then drop it from the answers and let hiera do the declarations for us per scenario?

So that is something I have been thinking about and was roughly referring to. I was thinking about exposing the options to `foreman_proxy_content` but in practice it's very unlikely someone modifies this and if they do, they can use `custom-hiera.yaml` as well. I think that leads to a cleaner setup, both from a code as well as a user perspective.

config/foreman.hiera/scenario/foreman-proxy-content/family/RedHat-8.yaml
ehelms: I think we are close to that being true or already there. Let me double check on that with Katello folks.

ehelms:
I think I hit the goal based on discussion. This drops the use of `foreman_proxy::plugin::pulp` in the answers file and relies on scenario configuration through hiera to configure the state of Pulp and Pulpcore based on scenario and OS.

config/foreman.hiera/scenario/foreman-proxy-content/family/RedHat-8.yaml
ekohl: I think that in the current approach these values end up not being used. I had hoped that we could override defaults this way but I never ended up having the time to implement it in Kafo.

ekohl (commented):
I do think this is the right approach. We just need to figure out the `proxy_pulp_*` params. It also gets rid of the useless `--[no-]enable-foreman-proxy-plugin-pulp` so that's a nice bonus.

config/foreman.hiera/scenario/foreman-proxy-content/family/RedHat-8.yaml
ehelms: In 3.18, these are still being used as they exist today. I think then if we want the potential to pull this proxy content for EL8 support back, we should leave this in and clean everything up for 4.0 later in the dev cycle. If we want to accept that number of changes for EL8 content proxy with EL8 is too high a cost and variance, then let's target nightly and do things right. Thoughts?

ekohl: I'm not entirely sure what you mean by that.

I'm saying that these 2 lines don't work and need to be dropped unless we modify Kafo. Right now they are just misleading because Kafo does **not** use Hiera when determining defaults. The answers file has a higher priority than this so it's never used.
```suggestion
```

ehelms:
This is designed to be cherry pickable back to 3.18. So the current state is:

 * Drops foreman_proxy::plugin::pulp from the answers file and lets hiera dictate per scenario, per OS
 * Contains a migration for smart proxy with content to turn on Pulp 2 content route proxying for EL8 only

config/katello-answers.yaml
```diff
@@ -85,12 +85,11 @@ foreman_proxy::plugin::discovery: false
 foreman_proxy::plugin::dns::infoblox: false
 foreman_proxy::plugin::monitoring: false
 foreman_proxy::plugin::openscap: false
-foreman_proxy::plugin::pulp:
-  pulpcore_enabled: true
 foreman_proxy::plugin::remote_execution::ssh: false
 foreman_proxy::plugin::salt: false
 foreman_proxy_content:
   proxy_pulp_yum_to_pulpcore: true
+  proxy_pulp_iso_to_pulpcore: true
```
ekohl, line 92: I don't see a migration for this. Is that intentional?

ehelms:
Here is the SELinux failure on a proxy (which has pulpcore-selinux installed):

```
/var/log/audit/audit.log:type=AVC msg=audit(1605819790.972:3427): avc:  denied  { write } for  pid=20200 comm="httpd" name="pulpcore-api.sock" dev="tmpfs" ino=161460 scontext=system_u:system_r:httpd_t:s0 tcontext=system_u:object_r:var_run_t:s0 tclass=sock_file permissive=0

# audit(1605819969.992:3439):
#  scontext="system_u:system_r:httpd_t:s0" tcontext="system_u:object_r:var_run_t:s0"
#  class="sock_file" perms="write"
#  comm="httpd" exe="" path=""
#  message="type=AVC msg=audit(1605819969.992:3439): avc:  denied  { write } for
#   pid=20202 comm="httpd" name="pulpcore-api.sock" dev="tmpfs" ino=161460
#   scontext=system_u:system_r:httpd_t:s0 tcontext=system_u:object_r:var_run_t:s0
#   tclass=sock_file permissive=1"
allow httpd_t var_run_t:sock_file write;
```

ekohl:
That should be fixed via pulpcore-selinux. It's probably not applying the right filesystem context rule on our location:
https://github.com/pulp/pulpcore-selinux/blob/3744dcdb6469675ba9ad31617f3e962b6bc45db7/pulpcore.fc#L25-L27

This is where pulpcore-selinux applied rules. Their Ansible installer places files in subdirectories of `/run`, but we place them at the top level.

I think you can use the following workaround locally:
```
chcon -t pulpcore_server_var_run_t /run/pulpcore-*.sock
```

Then it should work. If so, we should get that in the policy or change our socket locations to match (but I think our locations are good given our constraints).

ehelms:
@ekohl It occurred to me, I am pretty sure this is the Puppet SELinux bug. But not a new incarnation of it, rather the fact that this change does not trigger `pulpcore-selinux` to get installed. Why is that? Due to how we detect if Pulp or Pulpcore is present (https://github.com/theforeman/foreman-installer/blob/develop/hooks/boot/01-kafo-hook-extensions.rb#L80-L90). We could shift this check to simply seeing if `foreman_proxy_content` is enabled or if there was some way to peer into the hiera data at that stage of kafo?

ekohl:
Good call. I'll need to think about the best way. My intuition says that adding parameters still makes sense.

I'd say that we can simplify it to 3 parameters:
* Boolean pulp2
* Boolean pulp3
* Boolean pulp_is_mirror

This will also allow a standalone Pulp 3 non-mirror deployment outside of Katello itself (which is a long time desire I have).

ehelms:
Those three being parameters on `puppet-foreman_proxy_content` ?

ekohl:
That is indeed what I was thinking about.

evgeni:
This needs a rebase and the SELinux issue *should* be fixed by https://github.com/theforeman/pulpcore-packaging/pull/69

PR #613, opened 2020-11-16
use pulp2 for deb content on katello upgrade

jlsherrill:


wbclark (commented):
Default for foreman_proxy_content::proxy_pulp_deb_to_pulpcore is true in https://github.com/theforeman/puppet-foreman_proxy_content/pull/292

As this is currently written, new installs will initially get the default value true but the migration would immediately set it to false at an early stage of the installer run regardless. Setting foreman_proxy_content::proxy_pulp_deb_to_pulpcore to true in katello default answers here will prevent that issue (because the logic to check if the upgrade applies is `upgrade if answers[FPC] && answers[FPC][PROXY_DEB].nil?`)

regarding the other parameter katello::use_pulp_2_for_deb, that one defaults to false in puppet-katello. if the intention is that it should be false for new installs, and true for upgrades, you will run into a similar issue with the migration. since it's not explicitly defined in the default answers file, on a new install the migration will run. to prevent that the fix is the same: set the proper default for new installs in the katello answers file here.

it's likely also that some text fixtures will need updating too once the answers file is fixed (i didn't look at the failing tests but that's usually the case).

jlsherrill:
i think i addressed your comment @wbclark

spec/fixtures/pulpcore-migration-dont-proxy-yum/katello-answers-before.yaml
```diff
@@ -32,6 +32,7 @@ foreman_proxy::plugin::salt: false
 foreman_proxy_content:
   proxy_pulp_isos_to_pulpcore: false
   puppet: false
+  enable_deb: true
```
wbclark: My initial reaction was, the minimal change to get this test to work would be to omit this and add only `proxy_pulp_deb_to_pulpcore: false` to the katello-answers-after.yaml. My thinking was that adding `enable_deb: true` to the katello-answers-before.yaml fixture somewhat overloads the test as now it's also testing the new migration that you created.

However, thinking about it some more, I think that's actually fine, but let's also rename the test if we go that route. It should be sufficient to rename the test in `spec/migration_spec.rb` on L87 to `"pulpcore migration dont use content plugins on upgrades"` which is a better name I think.

Then also, `# git mv spec/fixtures/pulpcore-migration-{dont-proxy-yum,dont-use-content-plugins-on-upgrades}` and update the paths on L88 and L91 in `spec/migration_spec.rb` accordingly.

jlsherrill: updated!

config/katello.migrations/202011084145-dont-user-pulpcore-deb-on-upgrades.rb
```diff
@@ -0,0 +1,11 @@
...
+USE_PULP_2_FOR_DEB = 'use_pulp_2_for_deb'.freeze
+FOREMAN_PROXY_CONTENT = 'foreman_proxy_content'.freeze
+ENABLE_DEB = 'enable_deb'.freeze
+
+def upgrade
+  answers[KATELLO] = {} unless answers[KATELLO].is_a?(Hash)
+  answers[KATELLO][USE_PULP_2_FOR_DEB] = true
+end
+
+upgrade if answers[KATELLO] && answers[KATELLO][USE_PULP_2_FOR_DEB].nil? && answers[FOREMAN_PROXY_CONTENT][ENABLE_DEB]
```
ekohl: Why is it using `enable_deb` on `foreman_proxy_content` while `katello` also has the same parameter? Also, a separate method is not really needed. I'd also be fine with the simpler:

```suggestion
if answers[KATELLO].is_a?(Hash) && answers[KATELLO][USE_PULP_2_FOR_DEB].nil? && answers[KATELLO][ENABLE_DEB]
  answers[KATELLO][USE_PULP_2_FOR_DEB] = true
end
```

The same is true for the other migrations.

config/katello.migrations/202011084145-dont-user-pulpcore-deb-on-upgrades.rb
```diff
@@ -0,0 +1,11 @@
+KATELLO = 'katello'.freeze
+USE_PULP_2_FOR_DEB = 'use_pulp_2_for_deb'.freeze
+FOREMAN_PROXY_CONTENT = 'foreman_proxy_content'.freeze
+ENABLE_DEB = 'enable_deb'.freeze
+
+def upgrade
+  answers[KATELLO] = {} unless answers[KATELLO].is_a?(Hash)
```
ekohl: This line doesn't make sense in the current form. On line 11 it's already used as a hash so if it was `true` (which is technically a valid possibility), the migration would have crashed already.

ekohl (changes requested):

config/katello.migrations/202011084145-dont-user-pulpcore-deb-on-upgrades.rb
jlsherrill: i just mimicked what was done for the other migrations, should i update those too?

ekohl: I'm fine if you don't. Normally I don't like changing migrations unless there's bugs in them. However, I do like to keep new migrations short.

jlsherrill: thanks, updated!

config/katello.migrations/202011084145-dont-user-pulpcore-deb-on-upgrades.rb
```diff
@@ -0,0 +1,8 @@
+KATELLO = 'katello'.freeze
+USE_PULP_2_FOR_DEB = 'use_pulp_2_for_deb'.freeze
+FOREMAN_PROXY_CONTENT = 'foreman_proxy_content'.freeze
```
ekohl: This is now unused
```suggestion
```

jlsherrill: :+1:

jlsherrill:
@wbclark  @ekohl  i think i've addressed all your concerns

ekohl (approved):

PR #614, opened 2020-11-17
Add wildcard certificate test for katello-certs-check

ehelms:


ehelms:

bin/katello-certs-check
```diff
@@ -161,8 +161,8 @@ function check-ca-bundle () {
 }
 
 function check-ca-bundle-size () {
-    printf "Checking CA bundle size: "
     CHECK=$(grep -c "^--*BEGIN" $CA_BUNDLE_FILE)
+    printf "Checking CA bundle size: $CHECK"
```
ekohl: If there's anything on STDERR that will now be printed before the `Checking ...` part is printed. I think `printf` doesn't have a line ending so it might be better to call `printf "$CHECK"` separate.

ehelms: Fair point, updated

spec/katello_certs_check_spec.rb
```diff
@@ -52,4 +52,16 @@ def fixture(filename)
...
     end
   end
+
+  context 'with wildcard certificate' do
+    let(:key) { File.join(certs_directory, 'wildcard.key') }
+    let(:cert) { File.join(certs_directory, 'wildcard.crt') }
+
+    it 'completes correctly' do
+      command_with_certs = "#{command} -b #{ca} -k #{key} -c #{cert}"
+      _stdout, stderr, status = Open3.capture3(command_with_certs)
```
ekohl: Does it make sense to also check `stdout` to make sure `Checking CA bundle size: ` with all the extras is on a single line?

ekohl (approved):

PR #615, opened 2020-11-17
Fixes #30436 - add import-export paths to katello answers

jeremylenz:
This is the third PR from https://github.com/theforeman/puppet-pulpcore/pull/147#pullrequestreview-529534099

* override the default $allowed_{import,export}_path parameters for the Katello scenario only
* define a migration for the Katello scenario that sets the proper values of the parameters for users who already installed

(thanks to @chris1984 for some guidance)

### DIFF
--- a/config/katello-answers.yaml
+++ b/config/katello-answers.yaml
@@ -91,8 +91,6 @@ foreman_proxy::plugin::remote_execution::ssh: false
 foreman_proxy::plugin::salt: false
 foreman_proxy_content:
   proxy_pulp_yum_to_pulpcore: true
-  allowed_import_path: ['/var/lib/pulp/sync_imports', '/var/lib/pulp/imports']
-  allowed_export_path: ['/var/lib/pulp/exports']
 katello:
   use_pulp_2_for_yum: false
 puppet:
--- a/config/katello-answers.yaml
+++ b/config/katello-answers.yaml
@@ -91,6 +91,8 @@ foreman_proxy::plugin::remote_execution::ssh: false
 foreman_proxy::plugin::salt: false
 foreman_proxy_content:
   proxy_pulp_yum_to_pulpcore: true
+  allowed_import_path: ['/var/lib/pulp/sync_imports', '/var/lib/pulp/imports']
+  allowed_export_path: ['/var/lib/pulp/exports']
 katello:
   use_pulp_2_for_yum: false
 puppet:
--- a/config/katello.migrations/201117205851-add-import-export-params.rb
+++ b/config/katello.migrations/201117205851-add-import-export-params.rb
@@ -1,0 +1,8 @@
+if answers['pulpcore'].is_a?(Hash)
+  answers['pulpcore']['allowed_import_path'] = ['/var/lib/pulp/sync_imports', '/var/lib/pulp/imports']
+  answers['pulpcore']['allowed_export_path'] = ['/var/lib/pulp/exports']
+end
+if answers['foreman_proxy_content'].is_a?(Hash)
+  answers['foreman_proxy_content']['pulpcore']['allowed_import_path'] = ['/var/lib/pulp/sync_imports', '/var/lib/pulp/imports']
+  answers['foreman_proxy_content']['pulpcore']['allowed_export_path'] = ['/var/lib/pulp/exports']
+end
--- a/config/katello.migrations/201117205851-add-import-export-params.rb
+++ b/config/katello.migrations/201117205851-add-import-export-params.rb
@@ -1,7 +1,3 @@
-if answers['pulpcore'].is_a?(Hash)
-  answers['pulpcore']['allowed_import_path'] = ['/var/lib/pulp/sync_imports', '/var/lib/pulp/imports']
-  answers['pulpcore']['allowed_export_path'] = ['/var/lib/pulp/exports']
-end
 if answers['foreman_proxy_content'].is_a?(Hash)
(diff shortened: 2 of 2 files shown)


config/katello-answers.yaml
```diff
@@ -91,6 +91,8 @@ foreman_proxy::plugin::remote_execution::ssh: false
 foreman_proxy::plugin::salt: false
 foreman_proxy_content:
   proxy_pulp_yum_to_pulpcore: true
+  allowed_import_path: ['/var/lib/pulp/sync_imports', '/var/lib/pulp/imports']
+  allowed_export_path: ['/var/lib/pulp/exports']
```
ehelms: If these do not change the defaults as they are defined in puppet-foreman_proxy_content (which I suspect we would want that module to define the Katello defaults for those paths) you will not need to add them to the answers file here as those defaults will get picked up automatically.

config/katello.migrations/201117205851-add-import-export-params.rb
```diff
@@ -0,0 +1,8 @@
+if answers['pulpcore'].is_a?(Hash)
+  answers['pulpcore']['allowed_import_path'] = ['/var/lib/pulp/sync_imports', '/var/lib/pulp/imports']
+  answers['pulpcore']['allowed_export_path'] = ['/var/lib/pulp/exports']
```
ehelms: Just a general note to help with learning some of this, the `pulpcore` module is not exposed in the answers file and therefor changing them won't achieve anything. The pulpcore module is wrapped by the `puppet-foreman_proxy_content` module as I think you've encountered looking at your other PRs and thus all configuration has to flow through it or through hiera (an advanced method).

config/katello-answers.yaml
jeremylenz: @ehelms Ok, removing.

config/katello.migrations/201117205851-add-import-export-params.rb
jeremylenz: @ehelms So I guess that means I don't need the first section of this migration, and I'll leave only the second.

config/katello.migrations/201117205851-add-import-export-params.rb
```diff
@@ -0,0 +1,4 @@
+if answers['foreman_proxy_content'].is_a?(Hash)
+  answers['foreman_proxy_content']['pulpcore']['allowed_import_path'] = ['/var/lib/pulp/sync_imports', '/var/lib/pulp/imports']
+  answers['foreman_proxy_content']['pulpcore']['allowed_export_path'] = ['/var/lib/pulp/exports']
```
wbclark:
```suggestion
  answers['foreman_proxy_content']['pulpcore_allowed_import_path'] = ['/var/lib/pulp/sync_imports', '/var/lib/pulp/imports']
  answers['foreman_proxy_content']['pulpcore_allowed_export_path'] = ['/var/lib/pulp/exports']
```
This says that we will run a migration for the Katello scenario, that sets values for the parameters `pulpcore_allowed_{import,export}_path` in the `foreman_proxy_content` module (i.e., what you have defined at https://github.com/theforeman/puppet-foreman_proxy_content/pull/297 )

config/katello.migrations/201117205851-add-import-export-params.rb
```diff
@@ -0,0 +1,4 @@
+if answers['foreman_proxy_content'].is_a?(Hash)
```
wbclark, line 1: I wonder if we need to account for a possible case where `answers['foreman_proxy_content'] == true` rather than being a hash. I think some other migration would still run first and convert it to a hash but I'm not 100% positive.

ekohl:
FYI: if the defaults are correct you don't need a migration. So I think that in this case where they are new and in a very deployment specific scenario. I doubt you really need this PR at all and rather get them right in the puppet-foreman_proxy_content PR.

wbclark:
> FYI: if the defaults are correct you don't need a migration. So I think that in this case where they are new and in a very deployment specific scenario. I doubt you really need this PR at all and rather get them right in the puppet-foreman_proxy_content PR.

Katello and FPC scenarios should get different values, because FPC has no need for CV import/export.

So if we set correct defaults for FPC in puppet-FPC, then we need a migration in the Katello scenario to support CV import/export.

config/katello.migrations/201117205851-add-import-export-params.rb
jeremylenz: All the other migrations I see use the same check, `if answers['foreman_proxy_content'].is_a?(Hash)`.  The only exception is config/katello.migrations/200123161606-enable-pulpcore.rb:

```rb
  unless answers['foreman_proxy_content']
    answers['foreman_proxy_content'] = {} unless answers['foreman_proxy_content'].is_a?(Hash)
    answers['foreman_proxy_content']['proxy_pulp_isos_to_pulpcore'] = false
  end
```

This block would only run if it were `false` or `nil`, and would leave it unchanged in the event it were `true`.

Using `answers['foreman_proxy_content'].is_a?(Hash)` as a check seems reasonable to me, since it's also used by so many other migrations and I assume those don't break anything.

jeremylenz:
[test foreman-installer]

not sure how to address the testing situation here

jeremylenz:
@ekohl @ehelms  In light of the changes in https://github.com/theforeman/puppet-foreman_proxy_content/pull/297 and the [comment above](https://github.com/theforeman/foreman-installer/pull/615#issuecomment-731442094), should this PR be closed?

ekohl:
Yes

PR #616, opened 2020-11-17
Fixes #31326: Error in certs check if using only a shortname 

ehelms:
This currently includes https://github.com/theforeman/foreman-installer/pull/614 as it was easier for me to build on top of work done there in the testing script. I can split the two but I think its OK if the other goes in first as these are both improvements in certificate testing and reporting.


chris1984:
@ehelms do you want me to test this?

ehelms:
@chris1984 You can, my intent was to try to ensure new tests captured this. So if there are use cases, or examples of certificate set ups I could add that would also be helpful.

ehelms:

ehelms:
This now no longer depends on #618 and is a stand alone improvement that can be reviewed.

chris1984:
@ehelms the tests look fine, so will just go with those instead of doing manual testing.

bin/katello-certs-check
```diff
@@ -211,6 +211,38 @@ function check-cert-usage-key-encipherment () {
...
+    printf "Checking for use of shortname as CN"
+
+    SUBJECT_CN=$(openssl x509 -in $CERT_FILE -noout -subject -nameopt multiline|grep commonName|cut -d '=' -f 2|tr -d ' ')
+    if [[ $SUBJECT_CN != *"."* ]]; then
+        error 1 "The $(basename $CERT_FILE) is using a shortname for Common Name (CN) and cannot be used with $PROJECT.\n"
+    fi
+
+    DNS_LIST=$(openssl x509 -noout -text -in $CERT_FILE | grep DNS:)
+    if [[ $? == "0" ]]; then
+        if [[ "${#DNS_LIST[@]}" == 1 ]]; then
```
ekohl: Why is it needed to special case this (other than the error message)? Doesn't the `-gt 1` code handle this already?

ekohl (commented):
It's becoming more and more reasonable to rewrite the script in Ruby.

bin/katello-certs-check
```diff
@@ -215,6 +215,29 @@ function check-cert-usage-key-encipherment () {
...
 
+function check-shortname () {
+    printf "Checking for use of shortname as CN"
+
+    SUBJECT_CN=$(openssl x509 -in $CERT_FILE -noout -subject -nameopt multiline|grep commonName|cut -d '=' -f 2|tr -d ' ')
+    if [[ $SUBJECT_CN != *"."* ]]; then
+        error 1 "The $(basename $CERT_FILE) is using a shortname for Common Name (CN) and cannot be used with $PROJECT.\n"
+    fi
+
+    DNS_LIST=$(openssl x509 -noout -text -in $CERT_FILE | grep DNS:)
```
ekohl, line 226: `check-cert-san` ensures there is at least 1 SAN and it matches the subject. Since previously the subject was checked, we can safely assume that there is at least one SAN that is not a short name. That means this loop should be redundant. If it's not (and the test is hitting this) then is there a bug in `check-cert-san` or my logic here?

ehelms: I did consider that when building this. And I landed on doing this because this is a stand-alone check and has no expressed dependency on the ordering of the checks. If someone were to re-order them, we could lose out on a this check.

ekohl: Oh, there is also one potential flaw in my logic: `error` doesn't exit but rather it shows all the errors, right?

ehelms: Correct, `error` sets the error code and prints a message, but it does not exit immediately. So in theory, if you had 3 things wrong with your certificates, you would see all of them.

ekohl (approved):

PR #617, opened 2020-11-18
Fixes #27462 - Capture the openssl error correctly 

ehelms:
This is intended to replace https://github.com/theforeman/foreman-installer/pull/376 and add a test for the use case.


ehelms:

ehelms:
Rebased on latest develop for the new test structure

bin/katello-certs-check
```diff
@@ -152,11 +152,13 @@ function check-priv-key () {
 function check-ca-bundle () {
     printf "Checking CA bundle against the certificate file: "
     CHECK=$(openssl verify -CAfile $CA_BUNDLE_FILE -purpose sslserver -verbose $CERT_FILE 2>&1)
-    if [[ $? == "0" ]]; then
-        success
+
+    ERROR_PATTERN="error [0-9]+ at"
+    if [[ $CHECK =~ $ERROR_PATTERN || $? != "0" ]]; then
```
ekohl: I'm always cautious with using `$?` further from the actual execution since it's easy to break in refactoring. Could you define `ERROR_PATTERN` before `CHECK=...` or do it inline?

ekohl (changes requested):

bin/katello-certs-check
ehelms: I'm not sure if my change fully achieves what you are wanting.

ehelms:
Sample error output:

```
hecking to see if the private key matches the certificate: 
[OK]

Checking CA bundle against the certificate file: 
[FAIL]

The /home/ehelms/workspace/upstream/foreman-installer/spec/fixtures/katello-certs-check/certs/ca.crt does not verify the /home/ehelms/workspace/upstream/foreman-installer/spec/fixtures/katello-certs-check/certs/invalid.crt
CN = foreman.example.com
error 26 at 0 depth lookup: unsupported certificate purpose
error /home/ehelms/workspace/upstream/foreman-installer/spec/fixtures/katello-certs-check/certs/invalid.crt: verification failed

Checking CA bundle size: 
[OK]
```

bin/katello-certs-check
ekohl: I normally always keep the `$()` line and `$?` without any whitespace separation. My reasoning is that `$?` is a side effect of `$()` and side effects should always be minimized. If they're not together, someone might add in another command in between. I'd also move it to the front to keep it as close as possible.
```suggestion
    if [[ $? != "0" || $CHECK =~ $ERROR_PATTERN ]]; then
```

bin/katello-certs-check
```diff
@@ -151,12 +151,15 @@ function check-priv-key () {
 
 function check-ca-bundle () {
     printf "Checking CA bundle against the certificate file: "
+    ERROR_PATTERN="error [0-9]+ at"
     CHECK=$(openssl verify -CAfile $CA_BUNDLE_FILE -purpose sslserver -verbose $CERT_FILE 2>&1)
-    if [[ $? == "0" ]]; then
-        success
+    CHECK_STATUS=$?
```
ehelms, line 156: @ekohl Does this work to make it explicit and reduce the risk you are worried about?

ekohl: That's another way that reduces the risk.

ekohl (approved):

PR #618, opened 2020-11-18
Only regenerate certificate fixtures if they do not exist

ehelms:
This allows adding more certificate fixtures without having to
regenerate all certificates every time and leading to larger
diffs.

After doing a set of certificate testing PRs, I would like to get this re-factor in first to reduce the diffs.


spec/fixtures/katello-certs-check/certs/foreman.example.com.csr
```diff
@@ -0,0 +1,15 @@
+-----BEGIN CERTIFICATE REQUEST-----
```
ekohl: Can we keep this out? Perhaps the `.gitignore` rules need to be moved/updated?

spec/fixtures/katello-certs-check/certs/ca.srl
```diff
@@ -0,0 +1 @@
+7AD3BECEFC09E272CDAF17BD95CAD3E21D27A023
```
ekohl: Is the serial needed to be stored in git?

spec/katello_certs_check_spec.rb
```diff
@@ -5,35 +5,39 @@
...
-  it 'without parameters' do
-    stdout, stderr, status = Open3.capture3(command)
-    expect(stderr).to eq fixture('missing-parameter.txt')
-    expect(stdout).to eq ''
-    expect(status.exitstatus).to eq 1
-  end
+  let(:command) { File.join(__dir__, '..', 'bin', 'katello-certs-check') }
+  let(:directory) { File.join(FIXTURE_DIR, 'katello-certs-check') }
+  let(:certs_directory) { File.join(directory, 'certs') }
+  let(:ca) { File.join(certs_directory, 'ca.crt') }
```
ekohl, line 15: Is it needed to move these from where they were? Moving things around makes a diff harder to read and thus more work for a reviewer.

ekohl (changes requested):

spec/katello_certs_check_spec.rb
ehelms: Do you mean the certificates themselves or do you mean these lines in the test? In both cases,  For the former, yes. For the latter, strictly needed? No. Does it help to draw a clearer line for the context? Yes. I am also looking ahead at PRs that I have that will all need to make this structural change. And it being easier to do so here, and provide additional context with the change.

ehelms: I dunno if it helps, but Github provides a method to remove whitespace changes from the diff when reviewing, e.g.

https://github.com/theforeman/foreman-installer/pull/618/files?diff=unified&w=1

ekohl (approved):

PR #619, opened 2020-11-24
Fixes #31387 - Drop TLS 1.0 and TLS 1.1 from Apache

ekohl:
This tightens the defaults on Apache to only accept TLS 1.2+. The platforms that required this are going EOL and this makes it easier to comply with PCI-DSS and similar stricter policies.


ekohl:
I verified that CentOS 6.10 can connect over TLS 1.2 so that should not be an issue. Per https://bugzilla.redhat.com/show_bug.cgi?id=1289205 you need at least curl-7.19.7-50.el6. Per https://bugzilla.redhat.com/show_bug.cgi?id=1272504 installing nss-3.19.1-6.el6 will also default to TLS 1.2, but that should be redundant with this change. https://access.redhat.com/errata/RHBA-2016:0915 is the errata users should have installed prior to upgrading their server, which mentions that RHEL 6.8 fixed it.

config/foreman.hiera/family/RedHat-8.yaml
```diff
@@ -0,0 +1,7 @@
+---
+# EL8 doesn't have support for SSLv3 anymore and errors out on it. This
```
evgeni, line 2: wow, how annoying.

ekohl: I must admit I didn't test it myself on EL8, but I coped this from https://github.com/puppetlabs/puppetlabs-apache/blob/624330496a431dad03e430b8f2c9ac1914c3b8de/manifests/params.pp#L821-L825

ehelms:
[test foreman-installer]

ehelms (approved):

PR #620, opened 2020-11-26
reexec systemd after installing SELinux packages

ekohl:
Starting systemd 245 the daemon reloads SELinux contexts after a daemon-reload, but prior to that it needs to reexec. Both EL7 and EL8 are too old so this runs reexec. To limit the impact, it only does so if a package change was detected. This can fail if a metapackage (katello) already pulled in all packages but there's no reliable way to detect if a reexec is needed. Always running daemon-reexec can be unexpected.

This is untested and I'm not even sure if my logic is entirely correct.

### DIFF
--- a/hooks/boot/01-kafo-hook-extensions.rb
+++ b/hooks/boot/01-kafo-hook-extensions.rb
@@ -15,6 +15,14 @@ module HookContextExtension
     !File.exist?(success_file)
   end
 
+  # @summary Use Puppet's package resource to ensure package state
+  # @param [Array[String]] packages
+  #   A list of package names
+  # @param [String] state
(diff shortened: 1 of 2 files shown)


ekohl:
@ehelms this may also explain what you saw with Pulpcore's socket.

evgeni:
tests need updating.

I think this will solve the issue for fresh installs, but not for upgrades, yet I am unsure how to solve it there. %post of foreman-selinux seems rather harsh.

ekohl:
We would probably need to do the same thing in `%post` for `pulpcore-selinux` then. Still, doesn't sound too bad.

evgeni:
let's try https://github.com/theforeman/foreman-packaging/pull/6052 then

PR #621, opened 2020-11-30
Refs #28928 - Remove checkpoint_segments option

ekohl:
puppetlabs-postgresql 6.8.0 introduces support to remove config entries. This makes upgrades more reliable.

Currently untested, hence a draft.


ekohl:
At this point I think users have migrated so I'm not going to pursue this further.

PR #622, opened 2020-11-30
Foreman 2.3.0 GA picks

ekohl:
This includes a fix for qpid connection checking that doesn't generate a warning in the logs (very annoying for users). It also adds more checks to katello-certs-check.


ekohl:
@tbrisker I think this everything. It looks like a lot, but the tests take up a huge number of lines.

ekohl:
Not sure why the log isn't there. @ehelms is the job expected to work on the stable branch?

ehelms:
The job is, worked previously -- https://github.com/theforeman/foreman-installer/pull/612

ehelms:
[test foreman-installer]

ehelms:
Oh right, I think the test is no longer valid. We made the change in Kafo to not log if using `--help` which means that initial log file that was always empty does not get created if the first thing ran is `--help`.

tbrisker (approved):

PR #624, opened 2020-12-07
Fixes #31490 - Remove digital ocean plugin

snecklifter:
This plugin has not been maintained for a while now and causes
issues on installation. This commit removes the option from the
installer and can be easily reverted if the plugin is updated.

### DIFF
--- a/config/foreman-answers.yaml
+++ b/config/foreman-answers.yaml
@@ -32,7 +32,6 @@ foreman::plugin::chef: false
 foreman::plugin::column_view: false
 foreman::plugin::default_hostgroup: false
 foreman::plugin::dhcp_browser: false
-foreman::plugin::digitalocean: false
 foreman::plugin::discovery: false
 foreman::plugin::expire_hosts: false
 foreman::plugin::hooks: false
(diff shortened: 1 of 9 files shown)


ekohl (changes requested):
This needs a migration to also drop it from existing installations. In the bin directory there's a helper script to create the right files.

@timogoebel can you confirm that the plugin should be removed?

timogoebel:
> @timogoebel can you confirm that the plugin should be removed?

Yeah, I'm afraid it makes sense to remove it from the installer.

ekohl:
ok to test

ekohl (changes requested):
I think this would also be good to capture in a Redmine issue for the changelog. Other than that, it looks good, let's see what the tests say.

snecklifter:
> I think this would also be good to capture in a Redmine issue for the changelog. Other than that, it looks good, let's see what the tests say.

https://projects.theforeman.org/issues/31490

Please let me know if you need anything else.

ekohl:
If you use `Fixes #31490 - Remove ...` as the first line in the commit message then it will automatically be linked and closed. This allows our automation to work.

ekohl (approved):
Thanks!

PR #627, opened 2020-12-09
Fixes #31486: Remove postgresql-devel during postgresql upgrade

ehelms:
If a user has postgresql-devel installed, the removal of postgresql
will fail and prevent the upgrade from completing properly. All
subsequent runs will continually fail.

### DIFF
--- a/hooks/pre/30-el7_upgrade_postgresql.rb
+++ b/hooks/pre/30-el7_upgrade_postgresql.rb
@@ -11,6 +11,7 @@ def postgresql_12_upgrade
 
   execute!(%(scl enable rh-postgresql12 "PGSETUP_INITDB_OPTIONS='--lc-collate=#{collate} --lc-ctype=#{ctype} --locale=#{collate}' postgresql-setup --upgrade"))
   ensure_packages(['postgresql-server'], 'absent')
+  ensure_packages(['postgresql-devel'], 'absent')
   ensure_packages(['postgresql'], 'absent')
   execute!('rm -f /etc/systemd/system/postgresql.service')
   ensure_packages(['rh-postgresql12-syspaths'], 'installed')


tbrisker (approved):
makes sense, thanks @ehelms !

tbrisker:
2.3  - ba3a1c9
2.2 - 3a43f8a

hooks/pre/30-el7_upgrade_postgresql.rb
```diff
@@ -11,6 +11,7 @@ def postgresql_12_upgrade
 
   execute!(%(scl enable rh-postgresql12 "PGSETUP_INITDB_OPTIONS='--lc-collate=#{collate} --lc-ctype=#{ctype} --locale=#{collate}' postgresql-setup --upgrade"))
   ensure_packages(['postgresql-server'], 'absent')
+  ensure_packages(['postgresql-devel'], 'absent')
   ensure_packages(['postgresql'], 'absent')
```
evgeni, line 15: is there a reason those are three calls to `ensure_packages` (and thus 3 Puppet runs)?

hooks/pre/30-el7_upgrade_postgresql.rb
```diff
@@ -11,6 +11,7 @@ def postgresql_12_upgrade
 
   execute!(%(scl enable rh-postgresql12 "PGSETUP_INITDB_OPTIONS='--lc-collate=#{collate} --lc-ctype=#{ctype} --locale=#{collate}' postgresql-setup --upgrade"))
   ensure_packages(['postgresql-server'], 'absent')
+  ensure_packages(['postgresql-devel'], 'absent')
```
evgeni, line 14: should we install `rh-postgresql12-postgresql-devel` to replace it? like we do with `-contrib`?

hooks/pre/30-el7_upgrade_postgresql.rb
ekohl: Initially I wrote some Puppet code to only call Puppet once (and still provide the dependencies) but @ehelms didn't like that so now we're doing it the slow way.

PR #628, opened 2020-12-09
Add check for local pulpcore database to hooks

ehelms:


wbclark (approved):

wbclark:
Was this causing any issues and is there an associated Redmine ?

ehelms:
This is needed to allow Pulp 3 being installed on a content proxy (this check is what triggers things like rh-postgresql12 to get installed prior to the puppet run).

wbclark:
> This is needed to allow Pulp 3 being installed on a content proxy (this check is what triggers things like rh-postgresql12 to get installed prior to the puppet run).

OK, makes sense. I think it could also cause in issue in the (very edge) case where an user configures remote foreman and candlepin DBs but forgets to configure remote pulpcore DB or intentionally configures it to be local.

PR #632, opened 2020-12-16
Enable remote execution plugin since Katello requires it

ehelms:
Bit late to the party (https://github.com/theforeman/foreman-packaging/commit/af84d92162d701963edaffcd3d8833e27e405c59)


ekohl (approved):
If it's a hard dependency, it would be good to also include it [here](https://github.com/theforeman/puppet-katello/blob/2e15151a6d324149071c4e240bed78b5d977adf6/manifests/application.pp#L49).

ehelms:
Three places to have to declare something? Wooo!

PR #636, opened 2021-01-05
assume pulpcore is always enabled on katello and content_proxy

wbclark:
Requires https://github.com/theforeman/puppet-foreman_proxy_content/pull/306


wbclark:
This should also remove configuration such as

```
  proxy_pulp_yum_to_pulpcore: false
```

from default answers since these params will no longer do anything once puppet-FPC#306 is merged.

additionally there should be migrations to get all upgrades and new installs to the same status w.r.t. pulpcore.

ekohl:
a3d1da7605c642e85c03b1cfbb6303920b0b9e17 did this and more.

PR #638, opened 2021-01-12
Fixes #31614: Enable Pulp 3 by default on server and proxies

ehelms:
This is the companion PR to https://github.com/theforeman/puppet-foreman_proxy_content/pull/306 which aims to enable Pulp 3 everywhere and configure it in mirror mode on a content proxy. This moves enable and configure of the smart proxy pulp plugin into foreman_proxy_content module sense that has all the connection details and thus removes it from the answers file.


config/foreman-proxy-content.migrations/210112194603-enable-pulp3-content-proxy.rb
```diff
@@ -0,0 +1,7 @@
+answers.delete('foreman_proxy::plugin::pulp') if answers.key?('foreman_proxy::plugin::pulp')
```
ekohl: You can call delete safely even if it doesn't exist so this is safe:
```suggestion
answers.delete('foreman_proxy::plugin::pulp')
```

config/foreman-proxy-content.migrations/210112194603-enable-pulp3-content-proxy.rb
```diff
@@ -0,0 +1,7 @@
...
+
+if answers['foreman_proxy_content'].is_a?(Hash)
+  answers['foreman_proxy_content']['pulpcore_mirror'] = true
+  answers['foreman_proxy_content'].delete('proxy_pulp_isos_to_pulpcore')
+  answers['foreman_proxy_content'].delete('proxy_pulp_yum_to_pulpcore')
```
ekohl, line 11: In https://github.com/theforeman/puppet-foreman_proxy_content/pull/306 you've removed these parameters. Kafo should already do that automatically. There is a migration where I made the same mistake, but I don't think you need code for this.
```suggestion
```

config/katello.migrations/210112194551-enable-pulp3-content-proxy.rb
```diff
@@ -0,0 +1 @@
+answers.delete('foreman_proxy::plugin::pulp') if answers.key?('foreman_proxy::plugin::pulp')
```
ekohl: Same as above
```suggestion
answers.delete('foreman_proxy::plugin::pulp')
```

ekohl (commented):
Looks like a migration that adds the settings always runs and causes the tests to fail. I'm wondering if we should remove that migration. Alternative is to drop them here (as you had) with a comment that it's needed because `config/katello.migrations/200611220455-dont-proxy-pulp-yum-to-pulpcore-on-upgrades.rb`, `config/katello.migrations/200123161606-enable-pulpcore.rb` and `config/katello.migrations/202011084146-dont-proxy-pulp-deb-to-pulpcore-on-upgrades.rb` add them again.

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -90,7 +90,7 @@ def pulp_present?
   end
```
ekohl: This and `pulp_enabled?` can probably be removed now? They don't work anymore.

ekohl (changes requested):

hooks/boot/01-kafo-hook-extensions.rb
ehelms: Dropped alongside the hooks that relied on them

hooks/pre/10-reset_data.rb
```diff
@@ -72,18 +71,6 @@ def empty_mongo(config)
   execute!(cmd)
 end
```
ekohl: This function can also be removed since it's no longer used anywhere.

hooks/pre_validations/12-clear_pulp_content.rb
```diff
@@ -1,16 +0,0 @@
-if app_value(:clear_pulp_content)
```
ekohl, line 1: This app value is still mentioned in `KATELLO.md` under data reset. Looks like we also forgot to update that from `--reset` to `--reset-data`. There's also more things that are out of date (`--foreman-proxy-content-parent-fqdn` as also removed and `--foreman-proxy-content-certs-tar` as also renamed (by me) a long time ago) so I'm fine with leaving that for a future PR. It even still mentions katello-installer.git.

ekohl (changes requested):

hooks/pre_validations/12-clear_pulp_content.rb
ehelms: I think that whole Katello readme could use both some updates and some moving of things to the official Katello docs. I will take that on as a post merge task.

ekohl (approved):
Feel free to merge when all combined PRs have been ACKed.

PR #640, opened 2021-01-19
Fixes #31671: Ensure postgresql-contrib and -docs are removed during …

ehelms:
…upgrade


ekohl (approved):
Not a blocker, but for contrib we install the SCL equivalent if it's installed. Should we do the same for docs?

ekohl (approved):

PR #641, opened 2021-01-20
Fixing puppet-agent-oauth check to also work with debian based systems

timdeluxe:
This fixes the check for a correctly working puppet-agent-oauth package also for debian based systems.

Maybe this can be made smarter code-wise, but it should do the needful.

(There is no redmine ticket related to this so far)

### DIFF
--- a/hooks/pre/31-puppet_agent_oauth.rb
+++ b/hooks/pre/31-puppet_agent_oauth.rb
@@ -6,4 +6,11 @@ unless app_value(:noop)
       logger.error("Failed to reinstall puppet-agent-oauth. Please check that the package is available from a repository.") unless success
     end
   end
+  if File.exist?('/opt/puppetlabs/puppet/bin/ruby') && execute("dpkg-query --show --showformat='${db:Status-Status}' puppet-agent-oauth | grep -q '^installed'", false, false)
+    unless execute("/opt/puppetlabs/puppet/bin/ruby -e \"require 'oauth'\"", false, false)
+      success = execute("apt-get --reinstall install puppet-agent-oauth", false, true)
+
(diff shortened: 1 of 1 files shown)


hooks/pre/31-puppet_agent_oauth.rb
```diff
@@ -3,6 +3,13 @@
     unless execute("/opt/puppetlabs/puppet/bin/ruby -e \"require 'oauth'\"", false, false)
       success = execute("yum -y reinstall puppet-agent-oauth", false, true)
 
+      logger.error("Failed to reinstall puppet-agent-oauth. Please check that the package is available from a repository.") unless success
+    end
+  end
+  if File.exist?('/opt/puppetlabs/puppet/bin/ruby') && execute("dpkg-query --show --showformat='${db:Status-Status}' puppet-agent-oauth | grep -q '^installed'", false, false)
```
ekohl, line 9: Perhaps it can be formatted different. You are allowed to define functions in hooks so perhaps there should be something `def package_is_installed?(package)` that returns a boolean? Similar, a `reinstall_package(package)` can be good. You do have `facts` as a variable with the output of `facter` that can help you if needed.

ekohl (commented):
Thanks for this addition! It's certainly a good one now that Puppet 7 is out and users may see this more often.

ekohl:
In https://github.com/theforeman/foreman-installer/pull/706 I've expanded the implementation a bit. Would you mind having a look?

timdeluxe:
@ekohl Sorry that i have not worked on this so long. You basically did the work in #706 you requested me to do. Your implementation looks good. My proposal would be to close this PR in favor of yours?

ekohl:
That's what I thought. I wanted to get this issue fixed before we branch Foreman 3.0 next week.

PR #642, opened 2021-01-29
Fixes #31757 - Drop foreman proxy pulp plugin in foreman

ekohl:
In 01186b11dbc38ea45a570be758b681e1f5da7729 this was added, but it doesn't really make sense in the Foreman scenario. Only Katello can use it, but you need more things on the proxy which is only provided by the foreman-proxy-content scenario.


ehelms (approved):

ekohl:
Tests fail because the bump commit that's needed from https://github.com/theforeman/puppet-foreman_proxy_content/pull/317 hasn't been merged yet.

ekohl:
[test foreman-installer]

PR #646, opened 2021-02-08
Fixes #31834: Ensure PostgreSQL 12 is enabled on EL8

ehelms:
I was not sure if I should try to code up the steps to upgrade Foreman EL8 existing databases or leave that to the docs. The steps are effectively:

```
# ensure enough disk space on /var/lib/pgsql for twice the size of /var/lib/pgsql/data

systemctl stop postgresql foreman dynflow\*
mv /var/lib/pgsql/data /var/lib/pgsql/data-old
sed -i "s/\/var\/lib\/pgsql\/data/\/var\/lib\/pgsql\/data-old/g" data-old/postgresql.conf

dnf -y module remove postgresql-server
dnf -y module reset postgresql
dnf -y module enable postgresql:12
dnf -y install postgresql-server postgresql-upgrade

sudo -u postgres pg_upgrade --old-bindir /usr/lib64/pgsql/postgresql-10/bin --new-bindir /bin --old-datadir /var/lib/pgsql/data-old/ --new-datadir /var/lib/pgsql/data
```

### DIFF
--- a/hooks/pre/40-el8_modules.rb
+++ b/hooks/pre/40-el8_modules.rb
@@ -1,0 +1,7 @@
+if (module_enabled?('katello') || module_enabled?('foreman_proxy_content')) && el8?
+  ensure_dnf_module('pki-core', 'present')
+
+  if local_postgresql?
+    ensure_dnf_module('postgresql', '12')
+  end
+end
--- a/hooks/boot/01-kafo-hook-extensions.rb
+++ b/hooks/boot/01-kafo-hook-extensions.rb
@@ -31,6 +31,20 @@ module HookContextExtension
     end
   end
 
+  def ensure_dnf_module(dnf_module, version)
+    code = "package { ['#{dnf_module}']: ensure => #{version}, enable_only => true, provider => 'dnfmodule' }"
+    logger.info("Ensuring #{dnf_module}:#{version} is enabled")
+    stdout, stderr, status = apply_puppet_code(code)
+
+    unless [0, 2].include?(status.exitstatus)
+      log_and_say(:error, "Failed to ensure #{dnf_module}:#{version} is enabled")
+      log_and_say(:error, stderr.strip) if stderr && stderr.strip
+      logger.debug(stdout.strip) if stdout && stdout.strip
+      logger.debug("Exit status is #{status.exitstatus.inspect}")
+      exit(1)
+    end
+  end
+
   def apply_puppet_code(code)
     bin_path = Kafo::PuppetCommand.search_puppet_path('puppet')
     Open3.capture3(*Kafo::PuppetCommand.format_command("echo \"#{code}\" | #{bin_path} apply --detailed-exitcodes"))
@@ -91,6 +105,10 @@ module HookContextExtension
 
   def el7?
     facts[:os][:release][:major] == '7' && facts[:os][:family] == 'RedHat'
+  end
+
+  def el8?
+    facts[:os][:release][:major] == '8' && facts[:os][:family] == 'RedHat'
   end
 
   def log_and_say(level, message, do_say = true, do_log = true)


hooks/pre/40-el8_modules.rb
```diff
@@ -0,0 +1,14 @@
+if module_enabled?('katello') && el8?
+  execute!('dnf -y module enable pki-core')
+end
+
+if local_postgresql? && el8?
+  postgres_version, success = execute_command('postgres --version', false, true)
+
+  if success && !postgres_version.include?('12.')
```
ekohl: PG 13 is also out and I wouldn't be surprised if someone used it at some point. They may be annoyed that it forces them to downgrade.

Another thing to consider is users who use packages from postgresql.org. https://www.postgresql.org/download/linux/redhat/ suggests it installs as postgresql$VERSION-server (so postgesql12-server). Not sure if they'll show up and if so, how but we have seen community users who use them.

ehelms: Hrmm, sounds like you are saying I should just limit this to Katello and leave Foreman alone since it can be in a wide variety of states.

ekohl: I recall that user of postgresql.org was running Katello and ran into problems because of the evr package.

ehelms: Externally or local?

ekohl: https://community.theforeman.org/t/foreman-1-19-test-week/10533/4

ehelms: That was quite a bit ago. And I feel like Katello has traditionally supported PG that comes with the OS to reduce support burden and matrix. We did this on EL7 through the use of the SCL as well. I have simplified this PR to just do this for the Katello scenario.

hooks/pre/40-el8_modules.rb
```diff
@@ -0,0 +1,8 @@
+if (module_enabled?('katello') || module_enabled?('foreman_proxy_content')) && el8?
+  execute!('dnf -y module enable pki-core')
+
+  if local_postgresql?
+    execute!('dnf -y module reset postgresql', false, true)
+    execute!('dnf -y module enable postgresql:12', false, true)
```
ekohl: I think this fails if some postgres package was already installed. I think the katello meta package pulls that in, so I doubt this really works.

If the workflow is:
* install foreman-installer-katello
* run foreman-installer
* let foreman-installer install katello/foreman-proxy-content
Then it will probably work.

ekohl (changes requested):
I really doubt that this will work.

hooks/pre/40-el8_modules.rb
ehelms: From my testing, and I can do a little more. If you already have say postgresql 10 installed from that module, the reset clears the dnf metadata that prevents update. 

```
[root@42c3f74ab611 /]# rpm -q postgresql
postgresql-10.15-1.module_el8.3.0+619+dbc95fbc.x86_64
[root@42c3f74ab611 /]# ls /var/lib/               
alternatives/ dbus/         dhclient/     dnf/          games/        initramfs/    misc/         portables/    private/      rpm/          rpm-state/    selinux/      systemd/      
[root@42c3f74ab611 /]# rpm -q postgresql
postgresql-10.15-1.module_el8.3.0+619+dbc95fbc.x86_64
[root@42c3f74ab611 /]# dnf module reset postgresql
Failed to set locale, defaulting to C.UTF-8
Last metadata expiration check: 0:00:51 ago on Mon Feb 15 14:14:44 2021.
Dependencies resolved.
==========================================================================================================================================================================================================================================================================================
 Package                                                              Architecture                                                        Version                                                              Repository                                                            Size
==========================================================================================================================================================================================================================================================================================
Resetting modules:
 postgresql                                                                                                                                                                                                                                                                              

Transaction Summary
==========================================================================================================================================================================================================================================================================================

Is this ok [y/N]: y 
Complete!
[root@42c3f74ab611 /]# dnf module enable postgresql:12
Failed to set locale, defaulting to C.UTF-8
Last metadata expiration check: 0:00:57 ago on Mon Feb 15 14:14:44 2021.
Dependencies resolved.
==========================================================================================================================================================================================================================================================================================
 Package                                                              Architecture                                                        Version                                                              Repository                                                            Size
==========================================================================================================================================================================================================================================================================================
Enabling module streams:
 postgresql                                                                                                                               12                                                                                                                                             

Transaction Summary
==========================================================================================================================================================================================================================================================================================

Is this ok [y/N]: y
Complete!
[root@42c3f74ab611 /]# dnf install postgresql
Failed to set locale, defaulting to C.UTF-8
Last metadata expiration check: 0:01:03 ago on Mon Feb 15 14:14:44 2021.
Package postgresql-10.15-1.module_el8.3.0+619+dbc95fbc.x86_64 is already installed.
Dependencies resolved.
==========================================================================================================================================================================================================================================================================================
 Package                                                         Architecture                                                Version                                                                                 Repository                                                      Size
==========================================================================================================================================================================================================================================================================================
Upgrading:
 postgresql                                                      x86_64                                                      12.5-1.module_el8.3.0+620+f1abd701                                                      AppStream                                                      1.5 M

Transaction Summary
==========================================================================================================================================================================================================================================================================================
Upgrade  1 Package

Total download size: 1.5 M
Is this ok [y/N]: y
Downloading Packages:
postgresql-12.5-1.module_el8.3.0+620+f1abd701.x86_64.rpm                                                                                                                                                                                                  4.3 MB/s | 1.5 MB     00:00    
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Total                                                                                                                                                                                                                                                     3.1 MB/s | 1.5 MB     00:00     
Running transaction check
Transaction check succeeded.
Running transaction test
Transaction test succeeded.
Running transaction
  Preparing        :                                                                                                                                                                                                                                                                  1/1 
  Upgrading        : postgresql-12.5-1.module_el8.3.0+620+f1abd701.x86_64                                                                                                                                                                                                             1/2 
  Cleanup          : postgresql-10.15-1.module_el8.3.0+619+dbc95fbc.x86_64                                                                                                                                                                                                            2/2 
  Running scriptlet: postgresql-10.15-1.module_el8.3.0+619+dbc95fbc.x86_64                                                                                                                                                                                                            2/2 
  Verifying        : postgresql-12.5-1.module_el8.3.0+620+f1abd701.x86_64                                                                                                                                                                                                             1/2 
  Verifying        : postgresql-10.15-1.module_el8.3.0+619+dbc95fbc.x86_64                                                                                                                                                                                                            2/2 

Upgraded:
  postgresql-12.5-1.module_el8.3.0+620+f1abd701.x86_64                                                                                                                                                                                                                                    

Complete!
[root@42c3f74ab611 /]# 
```

ekohl: It looks like `rubygem-pg` depends on `libpq` which is **not** in the module. That means it's not version dependent, which is good for us.

It also appears nothing pulls in `psql` (which comes from the postgresql) pacakge even though `katello-debug` uses it. (Also triggered me to open https://github.com/theforeman/foreman-packaging/pull/6315).

That means it may work.

The only thing that concerns me is idempotency. On the next run (when we have installed postgresql), it will also reset and enable it again.

ehelms:
This now uses the dnfprovider for the package type through Puppet which is available as of 6.15.0 and 5.5.20 per https://tickets.puppetlabs.com/browse/PUP-10235

hooks/pre/40-el8_modules.rb
```diff
@@ -0,0 +1,7 @@
+if (module_enabled?('katello') || module_enabled?('foreman_proxy_content')) && el8?
+  ensure_dnf_module('pki-core', '10.6')
```
ekohl: I know we care about the postgresql version, but do we care here? Can we use `present`? Should we actually fix this in `puppet-candlepin` instead?

hooks/pre/40-el8_modules.rb
```diff
@@ -0,0 +1,7 @@
...
+
+  if local_postgresql?
+    ensure_dnf_module('postgresql', '12')
```
ekohl, line 5: I can take a look at if we can solve this in puppetlabs-postgresql as a long term solution.

hooks/pre/40-el8_modules.rb
ehelms: Present works for this.

hooks/pre/40-el8_modules.rb
ehelms: Could be nice there long term, I dont see modules going away.

hooks/pre/40-el8_modules.rb
```diff
@@ -0,0 +1,7 @@
+if (module_enabled?('katello') || module_enabled?('foreman_proxy_content')) && el8?
+  ensure_dnf_module('pki-core', 'present')
```
ekohl, line 2: I started with https://github.com/theforeman/puppet-candlepin/pull/175. We've observed that starting Puppet typically adds a few seconds so I'd prefer to solve it as part of the main loop.

ehelms: This blocks Katello so we need to get any changes and release ASAP.

ekohl (changes requested):
I've submitted https://github.com/theforeman/foreman-installer/pull/650 which I believe is a more correct approach. It doesn't deal with upgrades yet, but IMHO it's the right long term implementation.

ehelms:
#650 doesnt look like it will work for a release branch which we need this change on, so will we merge this change, cherry pick it and then pick up the long term fixes as they get released?

ekohl:
Summarizing an offline discussion:
* https://github.com/theforeman/puppet-candlepin/pull/175 should be turned into a release and included in 2.4
* PostgreSQL 12 is not a blocker for RC1 but is something we want to fix prior to GA so there are no Katello users on PostgreSQL 10.
* For PG 10 users (existing Foreman on EL8) we need to look into a strategy. It is likely we will make this an upgrade note that users need to upgrade prior to upgrading to Foreman 2.4 on EL8.

PR #647, opened 2021-02-09
Puppet module updates for 2.4

ehelms:


evgeni:
Can you also bump PostgreSQL to 6.10.1?

ekohl:
What @evgeni said. Also changed the topic slightly to be more precise. If you're updating it, mind doing the same to the commit message?

Puppetfile
```diff
@@ -1,17 +1,16 @@
 forge 'https://forgeapi.puppetlabs.com/'
 
-mod 'puppetlabs/postgresql',         '>= 5.6.0'
+mod 'puppetlabs/postgresql',         '>= 6.10.0'
```
ekohl: Might also be confusing that the stable requires a newer version than in develop. Would you mind doing the same in develop? It's a good change because it ensures it works with CentOS 8 stream.

ekohl (approved):

Puppetfile
evgeni: Technically, we don't *depend* on anything from that version. Just 6.9.0 and 6.10.0 are borked.

So I'd either bump the dep to `>= 6.10.1`, or leave it at 5.6.0 and just update the lockfile

evgeni (approved):

ekohl (approved):

PR #648, opened 2021-02-09
Drop puppet-pulp from Puppetfile

ehelms:


ekohl (approved):
This should be picked into 2.4 as well, right?

ehelms:
Given it looks different in a branch (because versions are included) I bundled it into https://github.com/theforeman/foreman-installer/pull/647/files#diff-2fdac63bae486868bc58af48f1236a663f64b59ebea96936a0d4bb5501d38ae9L9

PR #650, opened 2021-02-16
Fixes #31834 - Use PostgreSQL 12 on EL8

ekohl:
This aligns Red Hat 7 and Red Hat 8 on PostgreSQL 12. It uses the branch manage_dnf_module.

It may break existing installations on PostgreSQL 10 since there is no upgrade path. Those users are recommended to set version to 10 in their custom-hiera.yaml.

This is an alternative to https://github.com/theforeman/foreman-installer/pull/646.

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -1,7 +1,8 @@
 forge 'https://forgeapi.puppetlabs.com/'
 
 # Dependencies
-mod 'puppetlabs/postgresql',         '>= 6.10.1'
+# This uses a version with DNF module support but without dropping Puppet 5
+mod 'puppetlabs/postgresql',         :git => 'https://github.com/theforeman/puppetlabs-postgresql', :commit => '8d0efec9fb5a5df42ed64375c71a11d8b6f90b4c'
 mod 'theforeman/dhcp',               :git => 'https://github.com/theforeman/puppet-dhcp'
 mod 'theforeman/dns',                :git => 'https://github.com/theforeman/puppet-dns'
 mod 'theforeman/git',                :git => 'https://github.com/theforeman/puppet-git'
--- a/config/foreman.hiera/family/RedHat-8.yaml
+++ b/config/foreman.hiera/family/RedHat-8.yaml
@@ -5,3 +5,6 @@ apache::mod::ssl::ssl_protocol:
   - 'ALL'
   - '-TLSv1'
   - '-TLSv1.1'
+
+postgresql::globals::manage_dnf_module: true
+postgresql::globals::version: "12"


ehelms (approved):
I'd like to see this available in the next Foreman release so that before Katello 4.0 goes GA EL8 installs are on PG 12 to avoid unnecessary upgrades for users.

ekohl:
Now that my PR is merged I've updated this to point to a specific git commit. https://github.com/theforeman/theforeman.org/pull/1788 includes the upgrade instructions.

ehelms (approved):

ekohl:
For other reviewers: this is part of a larger change. I'd first like to get the upgrade instructions ready. We also don't want to break the upgrade pipelines so forklift needs to learn something too.

ehelms:
I've got a PR over in FOC to automate this process to expand its availability to both Forklift and users: https://github.com/theforeman/foreman-operations-collection/pull/21

ekohl:
I also found https://github.com/puppetlabs/puppetlabs-postgresql/pull/1253 which at least was an issue on my own server. Maybe it's not a problem. It also looks like they're working on a 7.0.0 release which includes my fix.

ekohl:
One complication is that this drops support for puppet 5 in metadata.json. Also looks like we'll need new module releases to allow version 7.x.

ehelms:
That sounds like a lot of churn and change to take on in Foreman 2.4? As this is needed for Katello 4.0 to go GA.

ekohl:
Honestly, given where Foreman 2.4 is and how much is involved here I'd say this is actually too late in the cycle. I'm starting to think we should push this to Katello 4.1.

ehelms:
> Honestly, given where Foreman 2.4 is and how much is involved here I'd say this is actually too late in the cycle. I'm starting to think we should push this to Katello 4.1.

Then we would need to *strongly* advertise PostgreSQL 12 everywhere and make it clear PG 10 is not supported. Users will still end up with PG 10 without enforcement and then we have to upgrade them. I do not think Katello 4.0 should GA without enforcement of PG 12.  Whether that is a version of this that works properly, or we use https://github.com/theforeman/foreman-installer/pull/646/files and update it in 4.1 to this method. This feels bad to start users out on the wrong foot who want to start with a clean EL8 or who start looking to migrate from EL7 to EL8.

ekohl:
I have switched it back to the specific commit ID we had previously. That allows us to not drop Puppet 5 and avoid the churn for 2.4. I do think we should soon change it to `>= 7.0.1` in develop to avoid being stuck.

ehelms:
> I do think we should soon change it to `>= 7.0.1` in develop to avoid being stuck.

:+1:

ehelms (approved):

ekohl:
Regardless of the approach we take in this PR, we still need to update forklift and our CI. Now that this has been merged, our upgrade pipelines will be broken.

PR #653, opened 2021-02-17
Fixes #31909 - Enable pki-core DNF module on EL8

ekohl:
On EL8 the pki-core module is needed by candlepin. candlepin-selinux pulls in candlepin. This means that in the installer the SELinux workaround hook needs to enable the module.

### DIFF
--- a/hooks/boot/01-kafo-hook-extensions.rb
+++ b/hooks/boot/01-kafo-hook-extensions.rb
@@ -93,6 +93,10 @@ module HookContextExtension
     facts[:os][:release][:major] == '7' && facts[:os][:family] == 'RedHat'
   end
 
+  def el8?
+    facts[:os][:release][:major] == '8' && facts[:os][:family] == 'RedHat'
+  end
+
   def log_and_say(level, message, do_say = true, do_log = true)
     style = case level
             when :error
--- a/hooks/pre/32-install_selinux_packages.rb
+++ b/hooks/pre/32-install_selinux_packages.rb
@@ -12,5 +12,10 @@ if facts.dig(:os, :selinux, :enabled)
   packages << 'candlepin-selinux' if katello_enabled?
   packages << 'pulpcore-selinux' if pulpcore_enabled?
 
+  if katello_enabled? && el8?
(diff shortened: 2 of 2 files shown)


hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -93,6 +93,10 @@ def el7?
     facts[:os][:release][:major] == '7' && facts[:os][:family] == 'RedHat'
   end
 
+  def el8?
+    facts[:os][:release][:major] == '7' && facts[:os][:family] == 'RedHat'
```
ehelms:
```suggestion
    facts[:os][:release][:major] == '8' && facts[:os][:family] == 'RedHat'
```

hooks/pre/32-install_selinux_packages.rb
```diff
@@ -12,5 +12,10 @@
   packages << 'candlepin-selinux' if katello_enabled?
   packages << 'pulpcore-selinux' if pulpcore_enabled?
 
+  if katello_enabled? && el8?
+    # candlepin-selinux pulls in candlepin which requires pki-core
+    execute!('dnf module enable pki-core --assumeyes')
```
ehelms:
```suggestion
    execute!('dnf module enable pki-core --assumeyes', false, true)
```

This avoids having the following output every time:

```
2021-02-17 20:51:27 [NOTICE] [root] Running validation checks
Executing: dnf module enable pki-core --assumeyes
Last metadata expiration check: 0:10:00 ago on Wed 17 Feb 2021 08:41:28 PM UTC.
Dependencies resolved.
================================================================================
 Package           Architecture     Version             Repository         Size
================================================================================
Enabling module streams:
 pki-core                           10.6                                        pki-deps                           10.6                                       
Transaction Summary
================================================================================

Complete!
```

hooks/boot/01-kafo-hook-extensions.rb
ekohl: I see I shouldn't push these things at the end of the day.

ehelms (approved):

PR #656, opened 2021-03-02
Automate unsetting variables

ekohl:
This PR is right now just to show the general concept. Completely untested.


hooks/boot/05-environment.rb
```diff
@@ -0,0 +1,5 @@
+%w[http_proxy https_proxy ssl_cert_file HTTP_PROXY HTTPS_PROXY SSL_CERT_FILE].each do |variable|
+  if ENV.delete(variable)
+    $stderr.puts "Unsetting environment variable #{variable}"
```
ehelms, line 3: I would think for standardization we should use the logger here to output this. And I'd consider doing it as a warning rather than a full on error.

ekohl: Yes, this was just the quickest way that I knew to show the concept.

ehelms:
I've taken the inspiration here and opened https://github.com/theforeman/foreman-installer/pull/665 with a Redmine issue.

PR #657, opened 2021-03-03
Fixes #32002: Use runuser for executing reset commands

ehelms:
Hides nonsense errors such as when a user cannot change directories
when running the reset-data command. Additionally, converts reset-data
execution into log output only.

### DIFF
--- a/hooks/boot/01-kafo-hook-extensions.rb
+++ b/hooks/boot/01-kafo-hook-extensions.rb
@@ -122,7 +122,7 @@ module HookContextExtension
     end
 
     unless status
-      log_and_say("#{command} failed! Check the output for error!", do_say, do_log)
+      log_and_say(:error, "#{command} failed! Check the output for error!", do_say, do_log)
       exit 1
     end
   end
--- a/hooks/pre/10-reset_data.rb
+++ b/hooks/pre/10-reset_data.rb
@@ -41,18 +41,18 @@ def load_db_config(db)
 end
 
 def empty_db_in_postgresql(db)
-  logger.info "Dropping #{db} database!"
+  logger.notice "Dropping #{db} database!"
 
   config = load_db_config(db)
   if remote_host?(config[:host])
     empty_database!(config)
   else
-    execute!("sudo -u postgres dropdb #{config[:database]}")
+    execute!("runuser -l postgres -c 'dropdb #{config[:database]}'", false, true)
   end
 end
 
 def reset_candlepin
-  execute!('rm -f /var/lib/candlepin/.puppet-candlepin-cpdb*')
+  execute!('rm -f /var/lib/candlepin/.puppet-candlepin-cpdb*', false, true)
   empty_db_in_postgresql('candlepin')
 end
 
@@ -72,14 +72,14 @@ def empty_database!(config)
         where schemaname = 'public';
       ))
   delete_statements = `#{generate_delete_statements}`
-  execute!(pg_sql_statement(config, delete_statements)) if delete_statements
+  execute!(pg_sql_statement(config, delete_statements), false, true) if delete_statements
 end
 
 def clear_pulpcore_content(content_dir)
   if File.directory?(content_dir)
-    logger.debug "Removing Pulpcore content from '#{content_dir}'"
+    logger.notice "Removing Pulpcore content from '#{content_dir}'"
     FileUtils.rm_rf(content_dir)
-    logger.info "Pulpcore content successfully removed from '#{content_dir}'"
+    logger.notice "Pulpcore content successfully removed from '#{content_dir}'"
   else
     logger.warn "Pulpcore content directory not present at '#{content_dir}'"
   end


ekohl (approved):
Using `runuser` is also better with various policies. Sudo should not be used by us in the installer. :+1: for that reason alone.

ehelms:
Example output:

```
2021-03-03 13:48:05 [NOTICE] [root] Loading default values from puppet modules...
2021-03-03 13:48:08 [NOTICE] [root] ... finished
Are you sure you want to continue? This will drop the databases, reset all configurations that you have made and bring all application data back to a fresh install. [y/n]
2021-03-03 13:48:10 [NOTICE] [root] Running validation checks
2021-03-03 13:49:14 [WARN  ] [pre] Pulpcore content directory not present at '/var/lib/pulp/docroot'
2021-03-03 13:49:18 [NOTICE] [configure] Starting system configuration.
  The total number of configuration tasks may increase during the run.
  Observe logs or specify --verbose-log-level to see individual configuration tasks.
2021-03-03 13:49:27 [NOTICE] [configure] 100 out of 1942 done.
2021-03-03 13:49:27 [NOTICE] [configure] 200 out of 1942 done.
2021-03-03 13:49:28 [NOTICE] [configure] 300 out of 1942 done.
```

wbclark (commented):
> Additionally, converts reset-data
> execution into log output only.

My gut reaction here is that `--reset-data` is something that users would never use lightly, so I would prefer to show more terminal output that immediately emphasizes the difference between doing this vs. running the installer normally.

ehelms:
> My gut reaction here is that `--reset-data` is something that users would never use lightly, so I would prefer to show more terminal output that immediately emphasizes the difference between doing this vs. running the installer normally.

So you want each of these operations to be output as a notice? What I want to avoid is printing naked output to the screen given we have been intentionally moving to log based output.

wbclark:
> > My gut reaction here is that `--reset-data` is something that users would never use lightly, so I would prefer to show more terminal output that immediately emphasizes the difference between doing this vs. running the installer normally.
> 
> So you want each of these operations to be output as a notice? What I want to avoid is printing naked output to the screen given we have been intentionally moving to log based output.

Dropping DBs is a pretty big deal, right? I think notice output is appropriate for that reason, and would still avoid the naked output.

ehelms:
I can update it such that you get the following output:

```
[root@pipe-katello-server-nightly-centos8 bats]# foreman-installer --reset-data
2021-03-03 19:33:00 [NOTICE] [root] Loading default values from puppet modules...
2021-03-03 19:33:03 [NOTICE] [root] ... finished
Are you sure you want to continue? This will drop the databases, reset all configurations that you have made and bring all application data back to a fresh install. [y/n]
y
2021-03-03 19:35:08 [NOTICE] [root] Running validation checks
2021-03-03 19:36:10 [NOTICE] [pre] Dropping foreman database!
2021-03-03 19:36:10 [NOTICE] [pre] Dropping candlepin database!
2021-03-03 19:36:11 [NOTICE] [pre] Dropping pulpcore database!
2021-03-03 19:36:11 [WARN  ] [pre] Pulpcore content directory not present at '/var/lib/pulp/docroot'
2021-03-03 19:36:15 [NOTICE] [configure] Starting system configuration.
```

ehelms:
@ekohl Does the content dir need an update? https://github.com/theforeman/foreman-installer/pull/657/files#diff-c6c2308a00cddbe7a02f7aff307d29e8630092354c3cbfb7ca4abfb524f80f4bR90

wbclark:
> I can update it such that you get the following output:

Assuming the full details (exact command executed) are still accessible at a more verbose level / in the logs, then the simple and immediate feedback in this approach looks ideal to me.

ehelms:
> Assuming the full details (exact command executed) are still accessible at a more verbose level / in the logs, then the simple and immediate feedback in this approach looks ideal to me.

They are indeed, from the logs:

```
2021-03-03 19:36:10 [NOTICE] [pre] Dropping foreman database!
2021-03-03 19:36:10 [DEBUG ] [root] Executing: runuser -l postgres -c 'dropdb foreman'
2021-03-03 19:36:10 [DEBUG ] [root] Executing: rm -f /var/lib/candlepin/.puppet-candlepin-cpdb*
2021-03-03 19:36:10 [NOTICE] [pre] Dropping candlepin database!
2021-03-03 19:36:10 [DEBUG ] [root] Executing: runuser -l postgres -c 'dropdb candlepin'
2021-03-03 19:36:11 [NOTICE] [pre] Dropping pulpcore database!
2021-03-03 19:36:11 [DEBUG ] [root] Executing: runuser -l postgres -c 'dropdb pulpcore'
2021-03-03 19:36:11 [WARN  ] [pre] Pulpcore content directory not present at '/var/lib/pulp/docroot'
```

wbclark (approved):

ekohl:
> @ekohl Does the content dir need an update? https://github.com/theforeman/foreman-installer/pull/657/files#diff-c6c2308a00cddbe7a02f7aff307d29e8630092354c3cbfb7ca4abfb524f80f4bR90

Yes. theforeman-pulpcore 2.0.0 changed the layout and that was shipped as part of 2.3. I think that deserves its own issue and appropriate cherry picks.

PR #659, opened 2021-03-04
Add webhooks plugins

adamruzicka:
requires:
- [x] https://github.com/theforeman/puppet-foreman/pull/920
- [x] https://github.com/theforeman/puppet-foreman_proxy/pull/651

### DIFF
--- a/config/katello.migrations/20210304102446_add_proxy_shellhooks_plugin.rb
+++ b/config/katello.migrations/20210304102446_add_proxy_shellhooks_plugin.rb
@@ -1,0 +1,1 @@
+answers['foreman_proxy::plugin::shellhooks'] ||= false
--- a/config/foreman-answers.yaml
+++ b/config/foreman-answers.yaml
@@ -67,5 +67,6 @@ foreman_proxy::plugin::omaha: false
 foreman_proxy::plugin::openscap: false
 foreman_proxy::plugin::remote_execution::ssh: false
 foreman_proxy::plugin::salt: false
+foreman_proxy::plugin::shellhooks: false
 puppet:
   server: true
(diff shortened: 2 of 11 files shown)


adamruzicka:
[test foreman-installer]

ekohl (approved):

config/katello.migrations/20210304102446_add_proxy_shellhooks_plugin.rb
```diff
@@ -0,0 +1 @@
+answers['foreman_proxy::plugin::shellhooks'] ||= false
```
ehelms, line 1: I think you also want to add a migration for foreman-proxy-content.migrations and add it to the answers file for that scenario as well.

adamruzicka: Good point, added

ehelms:
[test foreman-installer]

ehelms:
Some of the failures seen here are due to some job updates:

 1) See https://github.com/theforeman/jenkins-jobs/pull/32

 2) There are some modules that are not Puppet 7 compatible and cause a failure (See below). We can drop Puppet 7 from the job for now or ignore it when analyzing PRs. 

```
Using /usr/local/rvm/gems/ruby-2.7.0 with gemset foreman-installer-pr-test-139-2-7-7-0

2021-03-25 19:43:29 [ERROR ] [root] Puppet 7.5.0 does not meet requirements for herculesteam-augeasproviders_core (>= 5.0.0 < 7.0.0)

2021-03-25 19:43:29 [ERROR ] [root] Incompatible version of Puppet used, cannot continue

Puppet 7.5.0 does not meet requirements for herculesteam-augeasproviders_core (>= 5.0.0 < 7.0.0)

script returned exit code 30
```

ehelms:
@ekohl could you weigh in here on which direction we should proceed?

ekohl:
I'm leaning to ignoring for now. While it's not great, it may be a good motivation to support it.

ehelms:
[test foreman-installer]

ehelms:
Puppet 7 tests failing as expected

PR #661, opened 2021-03-12
correct theforeman/foreman_proxy pin to pull 17.1.0

evgeni:
the dependency was updated, but not the lockfile, resulting in Katello
failing to install with:

    Evaluation Error: Error while evaluating a Resource Statement,
    Could not find declared class foreman_proxy::plugin::container_gateway
    (file: /usr/share/foreman-installer/modules/foreman_proxy_content/manifests/init.pp, line: 245, column: 7)


ekohl (approved):
Oops

PR #662, opened 2021-03-18
Refs #32078 - Update theforeman/foreman_proxy to 17.1.1

ekohl:
This pulls in the explicit dynflow core service notifications to fix missing restarts.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ source 'https://rubygems.org'
 gem 'kafo', '~> 6.0'
 gem 'librarian-puppet'
 gem 'puppet', ENV.key?('PUPPET_VERSION') ? "~> #{ENV['PUPPET_VERSION']}" : '~> 6.0'
-gem 'facter', '>= 3.0'
+gem 'facter', '>= 3.0', '!= 4.0.52'
 
 gem 'puppet-strings'
 gem 'rake'
--- a/Puppetfile.lock
+++ b/Puppetfile.lock
@@ -82,7 +82,7 @@ FORGE
       puppetlabs-concat (>= 1.0.0, < 7.0.0)
       puppetlabs-postgresql (>= 6.5.0, < 7.0.0)
       puppetlabs-stdlib (>= 4.25.0, < 7.0.0)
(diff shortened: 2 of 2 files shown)


evgeni (approved):

ekohl:
[test foreman-installer]

ekohl:
The issue looks persistent but Jenkins doesn't archive the log file with the error.

ehelms:
[test foreman-installer]

ehelms:
I think this caused by the release of `kafo_parsers`:

```
Error: Evaluation Error: Error while evaluating a Function Call, 'downcase' parameter 'arg' expects a value of type Numeric, String, Array, or Hash, got Undef (file: /home/ehelms/workspace/upstream/foreman-installer/2.5-6.0/share/foreman-installer/modules/foreman/manifests/params.pp, line: 3, column: 17) on node 
```

The version of kafo_parsers for which this is broken is not released in this branch so we can pin the version here to get past this and match packaging. I'm looking into the breakage.

ehelms:
Ahh actually, this is related to facter. 4.0.52 appears to have broken this in my local testing.

ekohl:
I've pushed a commit to test this theory out. Not sure if we should merge it, but I want to know what CI says with it.

ehelms:
This error does not appear in the terminal or in the log files. The reason for this is we do not setup logging when running `--help` to prevent log file sprawl. The downside to this is that we are not at least setting up terminal logging at an error level to spit out error when running `--help`. I'll work on a fix for this if we agree that we should make this change due to an issue like this.

ekohl:
Perhaps it should still setup logging on the console but not the log file when running `--help`.

ekohl:
I've created https://github.com/theforeman/foreman-installer/pull/663 against develop so we can cherry pick it here.

ehelms:
Here is a follow up in Kafo to display this error in the future: https://github.com/theforeman/kafo/pull/328

ekohl:
Updated to use the cherry picked commit.

ehelms (approved):

PR #663, opened 2021-03-19
Don't use Facter 4.0.52

ekohl:
This has broken the FQDN fact when FFI is unavailable. It's fixed in git but no release has been made yet and this allows us to build again.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ source 'https://rubygems.org'
 gem 'kafo', '~> 6.0'
 gem 'librarian-puppet'
 gem 'puppet', ENV.key?('PUPPET_VERSION') ? "~> #{ENV['PUPPET_VERSION']}" : '~> 6.0'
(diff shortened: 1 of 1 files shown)


ehelms:
The expectation is that the next facter release will fix the issue?

ehelms (approved):
Should this get a Redmine issue to track it as part of the release? Could a Debian user end up broken with this? And to help track such that we dont accidentally package this version.

ekohl:
> The expectation is that the next facter release will fix the issue?

Yes, the git version already has a patch that resolves it so I think it's safe to assume the next release is fine.

> Should this get a Redmine issue to track it as part of the release? Could a Debian user end up broken with this? And to help track such that we dont accidentally package this version.

AFAIK this Gemfile is only used in development/CI. It can be added as a conflict in packaging to make sure they're never installed together but given that Puppet is loaded from AIO, the AIO Facter will also be used. I doubt we can really enforce it.

PR #664, opened 2021-03-23
Refs #32175: Set foreman-tasks backup to true on upgrades

ehelms:
For new installations, the intent is to default to task cleanup
to not store backups of the tasks deleted. Since the previous
behavior is to store backups on cleanup, when upgrading users
should retain the same behavior and intentionally disable backups
to prevent unexpected behavior changes.


ehelms:
Results from https://github.com/theforeman/puppet-foreman/pull/922

ekohl (approved):
:+1: pending merge of the puppet-foreman PR.

ehelms:
https://github.com/theforeman/puppet-foreman/pull/922 has been merged

PR #665, opened 2021-03-23
Fixes #32180: Unset system proxy and SSL environment variables for th…

ehelms:
…e duration of the install

Moves from checking for http proxy and SSL environment variables and
failing if they are present to a model of unsetting them during
the duration of the install run. This does not change them on
the system itself.

### DIFF
--- a/checks/env_variables.rb
+++ b/checks/env_variables.rb
@@ -1,9 +1,0 @@
-#!/usr/bin/env ruby
-
-variables = %w[http_proxy https_proxy ssl_cert_file
-               HTTP_PROXY HTTPS_PROXY SSL_CERT_FILE]
-
-if variables.map { |variable| ENV[variable] }.compact.any?
-  $stderr.puts "Please unset the following environment variables before running the installer: #{variables.join(', ')}"
-  exit 1
-end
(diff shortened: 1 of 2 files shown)


ekohl (approved):

khobbits:
_Posting here (instead of redmine), because this was the first result on google, when searching for the issue._

Out of curiosity, why do you need to unset these variables?  

My machine was unable to unable to pull down required packages via yum/dnf after unsetting these variables, as access to the internet was via http_proxy.

I worked around this, by adding a proxy definition into the /etc/yum.conf instead.

ekohl:
> I worked around this, by adding a proxy definition into the /etc/yum.conf instead.

This is indeed the intended solution. Within the installer itself we have various places where we do make some HTTP calls (like to register the Smart Proxy to Foreman) and they end up respecting the proxy settings, but most of the time it actually fails because you can't reach the Foreman via the HTTP proxy. Or it breaks client certificate based authentication.

Perhaps we should change those calls, but this is the what we decided would be best for most users.

khobbits:
Just a query...

Do you need to make any API calls, before doing the package install?

If not, it might make sense to pull down any external dependencies, like packages, before breaking internet access.

PR #666, opened 2021-03-23
Fixes #32186 - Install pulp-selinux when pulp2 is enabled

jturel:
/var/lib/pulp was getting its selinux context from selinux-pulpcore (var_lib_t) but it should be `httpd_sys_rw_content_t` when pulp2 is installed

### DIFF
--- a/hooks/pre/32-install_selinux_packages.rb
+++ b/hooks/pre/32-install_selinux_packages.rb
@@ -12,6 +12,7 @@ if facts.dig(:os, :selinux, :enabled)
   packages << 'candlepin-selinux' if katello_enabled?
   packages << 'pulpcore-selinux' if pulpcore_enabled?
   packages << 'crane-selinux' if pulp_enabled?
+  packages << 'pulp-selinux' if pulp_enabled?
 
   ensure_packages(packages, 'installed')
 end


jturel:
@ekohl this fixed the problem for me but I'm unable to discern if it's a deterministic fix or not. What do you think?

jturel:
2.2 is also affected but I suspect we aren't planning any releases there

ekohl (approved):
Context for others: https://github.com/theforeman/forklift/commit/c033e1560b394fd183e3e041ca98f8efee18184d stopped installing `katello` which stopped pulling in `pulp-selinux`. This makes the implicit dependency explicit.

It is not implemented in develop nor 2.4 since those stopped installing Pulp 2.

However, Katello 3.17 also suffers from this so we may need to implement a workaround in forklift for our pipelines to pass. Users are unlikely to hit this since the instructions still tell them to `yum install katello`.

cc @evgeni / @ehelms

ehelms:
We were curious when we discussed the change what might not work when we did. Looks like the answer was older releases. 

Tests are failing due to https://github.com/theforeman/foreman-installer/commit/b9049ef5f447e14ccbf62460ff745f27b4f65b7b which could be cherry picked into the branch to ensure tests do pass.

ehelms:
Oh and I'll look into a Forklift fix for older releases.

evgeni (approved):
Yeah, let's add a workaround in forklift to unblock pipes.

ehelms:
Forklift work around: https://github.com/theforeman/forklift/pull/1320

ekohl:
I'm going to merge this. Even if we need a workaround in Forklift because we won't release a 2.2 anymore, this is still safe to do and gives more guarantees to the end user.

PR #667, opened 2021-03-31
Raise the expected minimum Puppet version to 5.5.20

ekohl:
The Candlepin module expects at least Puppet 5.5.20 for DNF module support. This has been handled in packaging so it's safe to raise this.  Eventually we'll drop Puppet 5 since it's now EOL, but it's a larger effort that we're not ready for.

### DIFF
--- a/spec/puppet_version_spec.rb
+++ b/spec/puppet_version_spec.rb
@@ -2,7 +2,7 @@ require 'json'
 require 'semantic_puppet'
 
 describe 'Puppet module' do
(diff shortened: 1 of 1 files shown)


ehelms (approved):

ehelms:
This will hit the same test failures as -- https://github.com/theforeman/foreman-installer/pull/659#issuecomment-807376442 which need resolution

ehelms:
[test foreman-installer]

ehelms:
Puppet 7 failing is expected

PR #668, opened 2021-03-31
Handle server_puppetserver_metrics nil migration

ekohl:
In the Puppet module the parameter server_puppetserver_metrics was changed from Optional[Boolean] to Boolean. This breaks upgrades. The solution is to delete the entry if it's nil which causes Kafo to determine the new default automatically.

Needs https://github.com/theforeman/foreman-installer/pull/667 and other things to be green, but it should fix nightly.

### DIFF
--- a/config/foreman-proxy-content.migrations/210331121715-clear-puppetserver-nil-metrics.rb
+++ b/config/foreman-proxy-content.migrations/210331121715-clear-puppetserver-nil-metrics.rb
@@ -1,0 +1,3 @@
+if answers['puppet'].is_a?(Hash)
+  answers['puppet'].delete('server_puppetserver_metrics') if answers['puppet']['server_puppetserver_metrics'].nil?
+end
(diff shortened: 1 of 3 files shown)


ekohl:
Oh, for completeness: https://github.com/theforeman/puppet-puppet/commit/e799269abbe053933dadfdfc28db22cf80e86eda is the reason that this is needed.

wbclark (approved):
Checking the sosreport from the failed centos7 upgrade at foreman-nightly-rpm-pipeline #992, I observe the error:

```
$ grep '\[ERROR' var/log/foreman-installer/foreman.log 
2021-03-30 22:30:26 [ERROR ] [root] Parameter puppet-server-puppetserver-metrics invalid: nil is not a valid boolean
$
```

The migrations look good to me and should resolve this issue

ehelms:
[test foreman-installer]

ehelms:
[test foreman-installer]

PR #671, opened 2021-04-07
Fixes #32269 - Set ReservedCodeCache for Puppetserver

ekohl:
Puppetserver 6 really benefits from an increased ReservedCodeCache. Puppet recommends this in their tuning guide[1] and various deployments have shown to greatly benefit from it. 512m appears to be a good default since it works well even on small installs without real downsides.

Currently a draft since it's untested.

[1]: https://puppet.com/docs/puppet/6.21/server/tuning_guide.html

### DIFF
--- a/config/foreman-answers.yaml
+++ b/config/foreman-answers.yaml
@@ -70,3 +70,6 @@ foreman_proxy::plugin::salt: false
 foreman_proxy::plugin::shellhooks: false
 puppet:
   server: true
+  server_jvm_extra_args:
+    - "-Djruby.logger.class=com.puppetlabs.jruby_utils.jruby.Slf4jLogger"
+    - "-XX:ReservedCodeCacheSize=512m"
--- a/config/foreman-proxy-content.migrations/210407174237-add-puppet-reserved-code-cache.rb
+++ b/config/foreman-proxy-content.migrations/210407174237-add-puppet-reserved-code-cache.rb
@@ -1,0 +1,19 @@
+if answers['puppet'].is_a?(Hash)
+  reserved_code_cache_option = '-XX:ReservedCodeCacheSize'
+  reserved_code_cache_arg = "#{reserved_code_cache_option}=512m"
+  # Handle Optional[Variant[String,Array[String]]] $server_jvm_extra_args
+  if (answer = answers['puppet']['server_jvm_extra_args'])
+    if answer.is_a?(Array)
+      unless answer.any? { |arg| arg.include?(reserved_code_cache_option) }
+        answers['puppet']['server_jvm_extra_args'] << reserved_code_cache_arg
+      end
+    else
+      unless answer.include?(reserved_code_cache_option)
+        answers['puppet']['server_jvm_extra_args'] += " #{reserved_code_cache_arg}"
+      end
+    end
+  else
+    # The logger is silently added by the module if it's undef
+    answers['puppet']['server_jvm_extra_args'] = ['-Djruby.logger.class=com.puppetlabs.jruby_utils.jruby.Slf4jLogger', reserved_code_cache_arg]
+  end
+end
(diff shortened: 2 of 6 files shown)


ekohl:
I've verified all cases that I could think of:
* [x] Parameter not set
* Parameter set
  * Is array
    * [x] Option set already
    * [x] Is not set
  * Is string
    * [x] Option set already
    * [x] Is not set

Also checked on a real system that it does indeed show up in `/etc/sysconfig/puppetserver`.

config/foreman-answers.yaml
```diff
@@ -70,3 +70,6 @@ foreman_proxy::plugin::salt: false
 foreman_proxy::plugin::shellhooks: false
 puppet:
   server: true
+  server_jvm_extra_args:
```
ehelms, line 73: Are there any other commonly configured jvm settings that users would pass to this argument and potentially wipe out these defaults without realizing what they have done?

config/foreman-proxy-content.migrations/210407174237-add-puppet-reserved-code-cache.rb
```diff
@@ -0,0 +1,19 @@
+if answers['puppet'].is_a?(Hash)
+  reserved_code_cache_option = '-XX:ReservedCodeCacheSize'
+  reserved_code_cache_arg = "#{reserved_code_cache_option}=512m"
+  # Optional[Variant[String,Array[String]]]
```
ehelms: What does this comment mean in context here?

config/foreman-answers.yaml
ekohl: The migration should take care of that. These are just the defaults for fresh installs.

config/foreman-proxy-content.migrations/210407174237-add-puppet-reserved-code-cache.rb
ekohl: The data type of the variable. It should tell you enough to understand why there are 3 cases.

config/foreman-answers.yaml
ehelms: I was meaning more from a user passing arguments to the installer *after* a fresh install.

ekohl: None that I'm aware of. We already have dedicated parameters for the min and max heap as we do here:
https://github.com/theforeman/puppet-puppet/blob/c64b89694fb2a0815d1ade68d0cf98aa36524745/manifests/server/puppetserver.pp#L149-L150

That's also how I intend to make a dedicated parameter for it. Then well need to parse the existing answers to convert it in a migration, but I figured we have to do that anyway since today users may pass the option. This also allows for a safer cherry pick than a module change where you would have to deal with it anyway.

Of course, that I'm not aware of common options doesn't mean they don't exist.

config/foreman-proxy-content.migrations/210407174237-add-puppet-reserved-code-cache.rb
ehelms: Ahh, consider updating the comment to mention that cause I did not piece those two together :)

ehelms (approved):
Code looks good, consider making the comment a bit more verbose

config/foreman-proxy-content.migrations/210407174237-add-puppet-reserved-code-cache.rb
alexjfisher: Since I've been [thinking about `Optional` a lot recently](https://github.com/alexjfisher/puppet-lint-optional_default-check), I'm wondering what I think of `Optional[Variant[..` .

Give that `Optional[Foo]` is nothing more than a shorthand for `Variant[Undef, Foo]`...

Instead of `Optional[Variant[String,Array[String]]]`, is `Variant[Undef,String,Array[String]]` better style??

ekohl: I have followed your interest and had the same thought. However, I now just copied it from the definition: https://github.com/theforeman/puppet-puppet/blob/c64b89694fb2a0815d1ade68d0cf98aa36524745/manifests/init.pp#L698

alexjfisher: Yep, just something that caught my eye.  No need to change anything.

ekohl:
> Code looks good, consider making the comment a bit more verbose

Like this?

ehelms (approved):

PR #672, opened 2021-04-08
Fixes #32274 - Update default Apache ciphers

ekohl:
This takes the Mozilla intermediate SSL profile with EL7 software versions.


ehelms:
Failure is Puppet 7 and expected.

Does this need a manual test of an EL6 client?

wbclark (commented):
I borrowed an EL6 system to check. Picking the first cipher from the list here, I can see it is available:

```
# openssl ciphers -v 'TLSv1.2' | grep 'ECDHE-ECDSA-AES128-GCM-SHA256'
ECDHE-ECDSA-AES128-GCM-SHA256 TLSv1.2 Kx=ECDH     Au=ECDSA Enc=AESGCM(128) Mac=AEAD
# rpm -q openssl
openssl-1.0.1e-58.el6_10.x86_64
# cat /etc/redhat-release 
Red Hat Enterprise Linux Server release 6.10 (Santiago)
```

Is it necessary to support any older EL6 releases?

ekohl:
I think on RHEL curl may use gnutls instead of openssl. That might also be a good thing to verify.

I also haven't tried this out with Puppetserver (which uses Java). I'll see if i can try that out as well.

ehelms (approved):

PR #675, opened 2021-04-15
Fixes #32037 - Disable katello-agent on install, ensure enabled on up…

ehelms:
…grade

### DIFF
--- a/config/foreman-proxy-content.migrations/210415180240-enable-katello-agent.rb
+++ b/config/foreman-proxy-content.migrations/210415180240-enable-katello-agent.rb
@@ -1,0 +1,3 @@
+if answers['foreman_proxy_content'].is_a?(Hash) && answers['foreman_proxy_content'].key?('qpid_router')
+  answers['foreman_proxy_content']['enable_katello_agent'] = answers['foreman_proxy_content'].delete('qpid_router')
+end
--- a/config/katello.migrations/210415175000-enable-katello-agent.rb
+++ b/config/katello.migrations/210415175000-enable-katello-agent.rb
@@ -1,0 +1,3 @@
+if answers['foreman_proxy_content'].is_a?(Hash) && answers['foreman_proxy_content'].key?('qpid_router')
+  answers['foreman_proxy_content']['enable_katello_agent'] = answers['foreman_proxy_content'].delete('qpid_router')
+end
--- a/config/foreman.hiera/common.yaml
+++ b/config/foreman.hiera/common.yaml
@@ -12,6 +12,8 @@ foreman::config::apache::proxy_params:
   retry: '0'
   timeout: '900'
 
+katello::globals::enable_katello_agent: "%{alias('foreman_proxy_content::enable_katello_agent')}"
+
 katello::globals::enable_yum: "%{alias('foreman_proxy_content::enable_yum')}"
 katello::globals::enable_file: "%{alias('foreman_proxy_content::enable_file')}"
 katello::globals::enable_docker: "%{alias('foreman_proxy_content::enable_docker')}"


ehelms:
Needs:
  * https://github.com/theforeman/puppet-katello/pull/387
  * https://github.com/theforeman/puppet-foreman_proxy_content/pull/329

config/katello.migrations/210415175000-enable-katello-agent.rb
```diff
@@ -0,0 +1,3 @@
+if answers['foreman_proxy_content'] && answers['foreman_proxy_content']['enable_katello_agent'].nil?
+  answers['foreman_proxy_content']['enable_katello_agent'] = true
+end
```
wbclark, line 3: since this is the katello scenario migration, should this be `answers['katello']` throughout instead?

or does it work because the foreman.hiera for katello scenario includes the alias `katello::params::enable_katello_agent: "%{alias('foreman_proxy_content::enable_katello_agent')}"` ?

ehelms: > or does it work because the foreman.hiera for katello scenario includes the alias `katello::params::enable_katello_agent: "%{alias('foreman_proxy_content::enable_katello_agent')}"` ?

Correct. Everything is driven off of the user facing paramter `--foreman-proxy-content-enable-katello-agent`

config/foreman.hiera/scenario/katello.yaml
```diff
@@ -0,0 +1 @@
+katello::params::enable_katello_agent: "%{alias('foreman_proxy_content::enable_katello_agent')}"
```
ekohl: Should we put this in `common.yaml`? It doesn't hurt to have it there if the katello module isn't enabled and if any other scenario would have it, we wouldn't have to keep track of it. It also saves adding another layer to Hiera.

config/katello.migrations/210415175000-enable-katello-agent.rb
```diff
@@ -0,0 +1,3 @@
+if answers['foreman_proxy_content'] && answers['foreman_proxy_content']['enable_katello_agent'].nil?
```
ekohl: This only works if it's a hash. While unlikely, it could also be the boolean true.
```suggestion
if answers['foreman_proxy_content'].is_a?(Hash) && answers['foreman_proxy_content']['enable_katello_agent'].nil?
```

config/foreman-proxy-content.migrations/210415180240-enable-katello-agent.rb
```diff
@@ -0,0 +1,3 @@
+if answers['foreman_proxy_content'] && answers['foreman_proxy_content']['enable_katello_agent'].nil?
```
ekohl: This only works if it's a hash. While unlikely, it could also be the boolean true.
```suggestion
if answers['foreman_proxy_content'].is_a?(Hash) && answers['foreman_proxy_content']['enable_katello_agent'].nil?
```

ekohl (approved):
Small nit, otherwise :+1:

ekohl (approved):
:+1: merge when you feel it's ready.

ekohl (approved):

config/foreman-proxy-content.migrations/210415180240-enable-katello-agent.rb
```diff
@@ -0,0 +1,3 @@
+if answers['foreman_proxy_content'].is_a?(Hash) && answers['foreman_proxy_content']['enable_katello_agent'].nil?
+  answers['foreman_proxy_content']['enable_katello_agent'] = true
```
ekohl: We already had a parameter to enable/disable qpid dispatch. Should we instead "move" that answer?

ehelms: Yea, that's a good idea, as it allows a cleaner answers file that just lets the puppet module dictate the default for new installs and will handle if a user did actually disable qpid_router themselves.

config/foreman-proxy-content.migrations/210415180240-enable-katello-agent.rb
```diff
@@ -0,0 +1,3 @@
+if answers['foreman_proxy_content'].is_a?(Hash) && answers['foreman_proxy_content'].key?('qpid_router')
+  answers['foreman_proxy_content']['enable_katello_agent'] = answers['foreman_proxy_content']['qpid_router']
```
ekohl: This may be a bit more clear that it's really moving it:

```suggestion
  answers['foreman_proxy_content']['enable_katello_agent'] = answers['foreman_proxy_content'].delete('qpid_router')
```

ekohl (approved):
I like how small the change is now.

PR #679, opened 2021-05-05
Fixes #32532: Ensure a valid tuning option is supplied

ehelms:
Adds a validator to the tuning option to ensure that what is supplied
on the command line is in fact one of the valid options.

Requires: https://github.com/theforeman/kafo/pull/332

This has advantages over the current method which does not validate until the pre_commit stage thus running *more* of the installer than is necessary and provides a more Clamp native way to ensure input is valid.

Before:
```
# foreman-installer --disable-system-checks --tuning GIGANTIC
2021-05-05 20:52:05 [NOTICE] [root] Loading installer configuration. This will take some time.
2021-05-05 20:52:08 [NOTICE] [root] Running installer with log based terminal output at level NOTICE.
2021-05-05 20:52:08 [NOTICE] [root] Use -l to set the terminal output log level to ERROR, WARN, NOTICE, INFO, or DEBUG. See --full-help for definitions.
Invalid tuning profile
'GIGANTIC' is not one of default, medium, large, extra-large, extra-extra-large
```

After:
```
# foreman-installer --disable-system-checks --tuning GIGANTIC
Invalid option supplied for --tuning. Please choose from one of: default, medium, large, extra-large, extra-extra-large
See: 'foreman-installer --help'
```

This additionally solves the problem where if a user passed in `--tuning --medium` the default value for tuning would be selected *and* no error would be thrown for supplying the invalid `--medium`. Kafo strips out values passed to options that start with `-` (https://github.com/theforeman/kafo/blob/master/lib/kafo/kafo_configure.rb#L415).

### DIFF
--- a/hooks/boot/13-tuning.rb
+++ b/hooks/boot/13-tuning.rb
@@ -9,5 +9,11 @@ if has_custom_fact?('tuning')
     'INSTALLATION_SIZE',
     "Tune for an installation size. Choices: #{TUNING_SIZES.join(', ')}",
     :default => get_custom_fact(TUNING_FACT)
-  )
+  ) do |value|
+    unless TUNING_SIZES.include?(value)
+      signal_usage_error("Invalid option supplied for --tuning. Please choose from one of: #{TUNING_SIZES.join(', ')}")
+    end
+
+    value
+  end
 end
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'kafo', '~> 6.0'
+gem 'kafo', '~> 6.4'
 gem 'librarian-puppet'
 gem 'puppet', ENV.key?('PUPPET_VERSION') ? "~> #{ENV['PUPPET_VERSION']}" : '~> 6.0'
 gem 'facter', '>= 3.0', '!= 4.0.52'
--- a/hooks/pre_commit/13-tuning.rb
+++ b/hooks/pre_commit/13-tuning.rb
@@ -9,7 +9,6 @@ if app_option?(:tuning)
   }.freeze
   TUNING_FACT = 'tuning'.freeze
 
-  EXIT_INVALID_TUNING = 101
   EXIT_INSUFFICIENT_CPU_CORES = 102
   EXIT_INSUFFICIENT_MEMORY = 103
 
@@ -20,14 +19,8 @@ if app_option?(:tuning)
     new_tuning = current_tuning
   end
 
-  required = TUNING_SIZES[new_tuning]
-  if required.nil?
-    say "<%= color('Invalid tuning profile', :bad) %>"
-    say "'#{new_tuning}' is not one of #{TUNING_SIZES.keys.join(', ')}"
-    exit EXIT_INVALID_TUNING
-  end
-
   unless app_value(:disable_system_checks)
+    required = TUNING_SIZES[new_tuning]
     required_cores = required[:cpu_cores]
     required_memory = required[:memory]
 


hooks/boot/13-tuning.rb
```diff
@@ -4,10 +4,11 @@
   TUNING_SIZES = ['default'] + ['medium', 'large', 'extra-large', 'extra-extra-large']
   TUNING_FACT = 'tuning'.freeze
 
-  app_option(
-    '--tuning',
-    'INSTALLATION_SIZE',
-    "Tune for an installation size. Choices: #{TUNING_SIZES.join(', ')}",
-    :default => get_custom_fact(TUNING_FACT)
-  )
+  app_option('--tuning', 'INSTALLATION_SIZE', "Tune for an installation size. Choices: #{TUNING_SIZES.join(', ')}") do
```
ekohl: Don't you need to pass the value?
```suggestion
  app_option('--tuning', 'INSTALLATION_SIZE', "Tune for an installation size. Choices: #{TUNING_SIZES.join(', ')}") do |value|
```

hooks/boot/13-tuning.rb
```diff
@@ -4,10 +4,11 @@
...
+    unless TUNING_SIZES.include?(value)
+      puts "Invalid option supplied for --tuning. Please choose from one of: #{TUNING_SIZES.join(', ')}"
+      puts "See: 'foreman-installer --help'"
+      exit 101
+    end
```
ekohl, line 15: I think you need to return the value as well per https://github.com/mdub/clamp#validation
```suggestion
    end
  value
```

ekohl (changes requested):
I think that if we merge this without the Kafo change the block is silently ignored. So technically Kafo's change wouldn't be a blocker to merging.

hooks/boot/13-tuning.rb
```diff
@@ -4,10 +4,11 @@
...
-    '--tuning',
-    'INSTALLATION_SIZE',
-    "Tune for an installation size. Choices: #{TUNING_SIZES.join(', ')}",
-    :default => get_custom_fact(TUNING_FACT)
-  )
+  app_option('--tuning', 'INSTALLATION_SIZE', "Tune for an installation size. Choices: #{TUNING_SIZES.join(', ')}") do
+    unless TUNING_SIZES.include?(value)
+      puts "Invalid option supplied for --tuning. Please choose from one of: #{TUNING_SIZES.join(', ')}"
+      puts "See: 'foreman-installer --help'"
+      exit 101
```
wbclark: Where is this exit code defined ?

ehelms: Came from https://github.com/theforeman/foreman-installer/blob/develop/hooks/pre_commit/13-tuning.rb#L12

ehelms:
> I think that if we merge this without the Kafo change the block is silently ignored. So technically Kafo's change wouldn't be a blocker to merging.

I was leaning towards dropping this block of code: https://github.com/theforeman/foreman-installer/blob/develop/hooks/pre_commit/13-tuning.rb#L23-L28

Given it would be redundant and the option itself should handle validating it's inputs.

ekohl:
Then we do need a release of Kafo that's all packaged up. I do agree that that can be removed once it is.

ehelms:
Pushed up the cleanup and have opened a PR to release Kafo with this feature -- https://github.com/theforeman/kafo/pull/334

hooks/boot/13-tuning.rb
```diff
@@ -4,10 +4,13 @@
...
-  app_option(
-    '--tuning',
-    'INSTALLATION_SIZE',
-    "Tune for an installation size. Choices: #{TUNING_SIZES.join(', ')}",
-    :default => get_custom_fact(TUNING_FACT)
-  )
+  app_option('--tuning', 'INSTALLATION_SIZE', "Tune for an installation size. Choices: #{TUNING_SIZES.join(', ')}") do |value|
+    unless TUNING_SIZES.include?(value)
+      puts "Invalid option supplied for --tuning. Please choose from one of: #{TUNING_SIZES.join(', ')}"
+      puts "See: 'foreman-installer --help'"
```
evgeni: Can this be the binary name the user called? So ARGV[0], $0 or $PROGRAM_NAME or whatever Ruby likes today.

(Thinking branding ;))

ehelms: Good call, that led me to using a native method to handle this:

```
      raise Clamp::UsageError.new("Invalid option supplied for --tuning. Please choose from one of: #{TUNING_SIZES.join(', ')}", self)
```

Now this does directly reference Clamp which we try to avoid as far as I know. So, I can do one of two things or a combination of both:

 1) Introduce a Kafo::UsageError that inherits Clamp::UsageError
 2) Introduce a hook method `raise_usage_error` in Kafo that calls Clamp::UsageError

Number 2 feels like the best design while adhering to our creating a hook interface, but I can see an argument for the freedom to easily just reference native clamp exceptions. I'll hold https://github.com/theforeman/kafo/pull/334 until we decide since if I go the Kafo route that would be nice to include.

ehelms: Turns out that trying to call something in the Hook context won't exactly work because when the code is executed its executed within Clamps' environment. There is a method inside Clamps' environment named `signal_usage_error` that can be used. So I think the updated set of options are:

 1) Use Clamp::UsageError directly
 2) Use Clamp's `signal_usage_error`
 3) Define a Kafo::UsageError that inherits Clamp::UsageError and raise the Kafo version

ekohl: I think calling `signal_usage_error` is clean enough. It's an understandable API that doesn't mention clamp directly. I doubt they'll change it in a minor version and Kafo already depends on specific clamp versions so it's safe to expect it to remain unchanged. Perhaps we do need some tests in Kafo itself to ensure it works but that is more of a regression test and doesn't need a new release.

hooks/boot/13-tuning.rb
```diff
@@ -4,10 +4,11 @@
   TUNING_SIZES = ['default'] + ['medium', 'large', 'extra-large', 'extra-extra-large']
   TUNING_FACT = 'tuning'.freeze
 
-  app_option(
-    '--tuning',
-    'INSTALLATION_SIZE',
-    "Tune for an installation size. Choices: #{TUNING_SIZES.join(', ')}",
-    :default => get_custom_fact(TUNING_FACT)
```
ekohl, line 11: This default is lost in the refactor?

hooks/boot/13-tuning.rb
ehelms: Excellent, I have gone with signal_usage_error

ekohl (approved):
Perhaps this line should be updated too:
https://github.com/theforeman/foreman-installer/blob/91e12d9d1e331c90201d2eb976c74198da3c51a1/Gemfile#L3

It's only for CI, but it's the closest thing we have to specifying the required Kafo version somewhere other than in packaging itself.

ehelms:
> It's only for CI, but it's the closest thing we have to specifying the required Kafo version somewhere other than in packaging itself.

Agreed, I was thinking the same thing.

ekohl (approved):

PR #680, opened 2021-05-06
Restore DNF module management for PG on EL8

ekohl:
In fdc20de529ebcaba1708bf9afa43a23e96a7858a the version of puppetlabs/postgresql was pinned to an unreleased version which does have DNF module management but doesn't drop Puppet 5. Then in 157bd6a7165e11c87e53aa5338b57e58018ed8ee it was pinned to >= 6.10.1 which is too old and lacks DNF module management. That means it ends up using the default PostgreSQL module which is version 10.

Fixes: 157bd6a7165e11c87e53aa5338b57e58018ed8ee

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -1,6 +1,7 @@
 forge 'https://forgeapi.puppetlabs.com/'
 
-mod 'puppetlabs/postgresql',           '>= 6.10.1 < 7.0.0'
+# This uses a version with DNF module support but without dropping Puppet 5
+mod 'puppetlabs/postgresql',           :git => 'https://github.com/theforeman/puppetlabs-postgresql', :commit => '8d0efec9fb5a5df42ed64375c71a11d8b6f90b4c'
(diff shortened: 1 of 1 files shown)


evgeni (approved):

ekohl:
Failure is expected.

tbrisker:
Are we certain this will work now? the pr test failed with
```
Using /usr/local/rvm/gems/ruby-2.7.0 with gemset foreman-installer-pr-test-172-2-7-7-0
2021-05-06 12:16:19 [ERROR ] [root] Puppet 7.6.1 does not meet requirements for herculesteam-augeasproviders_core (>= 5.0.0 < 7.0.0)
2021-05-06 12:16:19 [ERROR ] [root] Incompatible version of Puppet used, cannot continue
Puppet 7.6.1 does not meet requirements for herculesteam-augeasproviders_core (>= 5.0.0 < 7.0.0)
script returned exit code 30
```
and i don't want to end up with another DOA rc

tbrisker:
ah, ignore me, now i see the comment that it's expected. ☕

PR #681, opened 2021-05-09
Update lockfile for 2.5.0-rc2

tbrisker:
@ekohl i re-ran librarian to make sure that the correct version gets pulled in to rc2 but i'm not sure that this is correct - please confirm before tagging rc2


ekohl (approved):
Oh yes, I'm glad I didn't tag on Friday. Clearly I wasn't paying attention 100%.

PR #682, opened 2021-05-10
Fixes #32011 - Fix logger typo in reset data hook

wbclark:


ekohl (changes requested):
Normally we use `git cherry-pick -x` so it links back to the original commit.

wbclark:
> Normally we use `git cherry-pick -x` so it links back to the original commit.

gotcha, will fix the commit message. thanks!

ekohl (approved):

PR #683, opened 2021-05-17
Fixes #32610: Drop testing for Puppet 5 compatability

ehelms:


spec/puppet_version_spec.rb
```diff
@@ -2,7 +2,7 @@
 require 'semantic_puppet'
 
 describe 'Puppet module' do
-  VERSIONS = ['5.5.20', '6.0.0'].map { |v| SemanticPuppet::Version.parse(v) }
+  VERSIONS = ['6.0.0'].map { |v| SemanticPuppet::Version.parse(v) }
```
ekohl: I know various modules will move to 6.1.0 as the minimum version (since it contains a nice fix to avoid the need for `systemctl daemon-reload` in many cases).
```suggestion
  VERSIONS = ['6.1.0'].map { |v| SemanticPuppet::Version.parse(v) }
```

ekohl (changes requested):

spec/puppet_version_spec.rb
ekohl: Oh, if https://github.com/theforeman/puppet-candlepin/pull/200 is merged it should really be testing for `6.15.0` as lower bound.

ehelms: I would like not to combine dropping 5 with updating the 6 minimum version. Additionally, if not all modules have a minimum of 6.1.0 would not changing to 6.1.0 then break this test?

ekohl: My observation is that many modules that drop Puppet 5 move their lower bound to 6.1.0. This is what Voxpupuli does.

> Additionally, if not all modules have a minimum of 6.1.0 would not changing to 6.1.0 then break this test?

I think you're misunderstanding the test. It states that the module should accept Puppet version x.y.z. Since it uses semantic versions in testing, anything that's `>= 6.0.0` should also accept `6.1.0`.

So in my suggestion I basically made it test for that it should accept 6.1.0. If it also accepts 6.0.0, that's fine but we don't care.

ehelms: Thank you for the explanation.

ekohl (approved):

PR #684, opened 2021-05-18
Fixes #32620: Unset NSS_DEFAULT_DB_TYPE during run

ehelms:
When NSS_DEFAULT_DB_TYPE is set to, for example, sql then qpidd
will fail to start.

### DIFF
--- a/hooks/boot/05-environment.rb
+++ b/hooks/boot/05-environment.rb
@@ -1,4 +1,4 @@
-%w[http_proxy https_proxy ssl_cert_file HTTP_PROXY HTTPS_PROXY SSL_CERT_FILE].each do |variable|
+%w[http_proxy https_proxy ssl_cert_file HTTP_PROXY HTTPS_PROXY SSL_CERT_FILE NSS_DEFAULT_DB_TYPE].each do |variable|
   if ::ENV.delete(variable)
     logger.warn "Unsetting environment variable '#{variable}' for the duration of the install."
   end


ekohl (commented):
Is this actually something we should do in our certs module instead? This feels very wrong to me.

evgeni:
Doesn't this mean qpid will fail to *re*start when the machine boots, or the services are restarted outside the installer?

ekohl:
https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/Reference/NSS_tools_:_certutil says you can use a prefix in the path. Either `sql:` or `dbm:`. I think we should be explicit on creation if dbm is the only one that works.

evgeni:
Now having read https://projects.theforeman.org/issues/32620 and Ewouds link, I tend to agree, let's make `certs` call `certutil` with `dbm:<PATH>` explicitly.

ehelms:
Interestingly, I have attempted to set this, delete the nssdb and then re-run the installer and hit no errors. Perhaps this is a bug in older qpidd versions?

ekohl:
Is there a difference in platform? EL7 vs EL8 (vs Fedora/EL9)?

ehelms:
From all my local testing, it appears something within the installer environment ignores this environment variable. I am wondering if the switch of the installer to the SCL and the cleaning of environment variables (https://github.com/theforeman/kafo/blob/master/lib/kafo/puppet_command.rb#L61) effectively solved this. Opting to close since I cannot reproduce the original issue.

evgeni:
@ehelms still feels like `katello/certs` could be more defensive here (when used outside the installer)?

ehelms:
I believe https://github.com/theforeman/puppet-certs/pull/327 will eliminate this being an issue. On EL7, both versions of the NSSDB are supported now.

PR #686, opened 2021-05-19
Update puppet-foreman_proxy_content and puppet-pulpcore

ehelms:


wbclark (approved):
LGTM

So that I understand the process, could you confirm that these changes were generated automatically with `librarian-puppet update foreman_proxy_content` (which would have been enough to update the pulpcore version as well)?

ehelms:
I updated Puppetfile and then ran `bundle exec rake build` to update the lockfile

PR #690, opened 2021-06-25
Fixes #32885: Remove puppet from foreman::user_groups if present

ehelms:
The change https://github.com/theforeman/puppet-foreman/pull/938
introduced dynamically determining if the puppet group needed to
be added to the foreman user. This no longer needs to be set
explicitly on the foreman::user_groups parameter thus the migration
cleans it up if present.

### DIFF
--- a/config/katello-answers.yaml
+++ b/config/katello-answers.yaml
@@ -22,7 +22,6 @@ foreman:
   server_ssl_chain: /etc/pki/katello/certs/katello-server-ca.crt
   server_ssl_crl: ""
   server_ssl_key: /etc/pki/katello/private/katello-apache.key
-  user_groups: []
 foreman::cli: true
 foreman::cli::ansible: false
 foreman::cli::azure: false
--- a/config/katello.migrations/190314163941-remove-foreman-user-groups.rb
+++ b/config/katello.migrations/190314163941-remove-foreman-user-groups.rb
@@ -1,5 +1,0 @@
-if answers['foreman'].is_a?(Hash)
-  answers['foreman']['user_groups'] = []
-elsif answers['foreman']
-  answers['foreman'] = { 'user_groups' => [] }
-end
--- a/config/katello.migrations/210625142712-dynamic-puppet-in-foreman-groups.rb
+++ b/config/katello.migrations/210625142712-dynamic-puppet-in-foreman-groups.rb
@@ -1,0 +1,5 @@
+if answers['foreman'].is_a?(Hash)
+  if answers['foreman']['user_groups'] && answers['foreman']['user_groups'].include?('puppet')
+    answers['foreman']['user_groups'].delete('puppet')
+  end
+end
--- a/spec/fixtures/cleanup-foreman-user-groups/foreman-answers-after.yaml
+++ b/spec/fixtures/cleanup-foreman-user-groups/foreman-answers-after.yaml
@@ -1,0 +1,2 @@
+foreman:
+  user_groups: []
--- a/config/foreman.migrations/20210625142707_dynamic_puppet_in_foreman_groups.rb
+++ b/config/foreman.migrations/20210625142707_dynamic_puppet_in_foreman_groups.rb
@@ -1,0 +1,5 @@
+if answers['foreman'].is_a?(Hash)
+  if answers['foreman']['user_groups'] && answers['foreman']['user_groups'].include?('puppet')
+    answers['foreman']['user_groups'].delete('puppet')
+  end
+end
--- a/spec/fixtures/cleanup-foreman-user-groups/foreman-answers-before.yaml
+++ b/spec/fixtures/cleanup-foreman-user-groups/foreman-answers-before.yaml
@@ -1,0 +1,2 @@
+foreman:
+  user_groups: ['puppet']
--- a/spec/fixtures/cleanup-foreman-user-groups/katello-answers-after.yaml
+++ b/spec/fixtures/cleanup-foreman-user-groups/katello-answers-after.yaml
@@ -1,0 +1,2 @@
+foreman:
+  user_groups: []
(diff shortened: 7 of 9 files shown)


config/katello.migrations/210625142712-dynamic-puppet-in-foreman-groups.rb
```diff
@@ -0,0 +1,3 @@
+if answers['foreman'].is_a?(Hash)
+  answers['foreman']['user_groups'].delete('puppet') if answers['foreman']['user_groups'].include?('puppet')
```
ekohl: Can we make this the same as the Foreman migration? That should be safer.

ekohl (changes requested):
We can also drop this line:
https://github.com/theforeman/foreman-installer/blob/3cbb7bd7201654a491a927cbb3d6867f50523c99/config/katello-answers.yaml#L26

config/katello-answers.yaml
```diff
@@ -23,7 +23,6 @@ foreman:
   server_ssl_chain: /etc/pki/katello/certs/katello-server-ca.crt
   server_ssl_crl: ""
   server_ssl_key: /etc/pki/katello/private/katello-apache.key
-  user_groups: []
```
ehelms, line 25: I figured this does not need to be here anymore.

config/katello.migrations/190314163941-remove-foreman-user-groups.rb
```diff
@@ -1,5 +1,3 @@
-if answers['foreman'].is_a?(Hash)
+if answers['foreman'].is_a?(Hash) && answers['foreman'].key?('user_groups')
   answers['foreman']['user_groups'] = []
-elsif answers['foreman']
-  answers['foreman'] = { 'user_groups' => [] }
```
ehelms, line 4: Cleaning up the answers file leads to this curiosity, where we have this migration that is no longer valid as far as I can tell. I wasn't sure if I should do the update like so, or if I should just basically remove the entire migration.

ekohl (approved):
:+1: pending the merge in puppet-foreman itself.

config/katello.migrations/190314163941-remove-foreman-user-groups.rb
wbclark: Looking at the history and timeline behind https://github.com/theforeman/foreman-installer/pull/331/commits/65039e54e1973c2893e900c62ba78e23b42132ad , I think it should be removed since you are also removing `user_groups: []` from the default answers

spec/fixtures/cleanup-foreman-user-groups/foreman-answers-after.yaml
```diff
@@ -0,0 +1,2 @@
+foreman:
+  user_groups: []
```
wbclark, line 2:
```suggestion
foreman:
```

Should we add a 2nd piece of logic to the migration to remove the key entirely if the value ends up `[]`?

ehelms: I prefer not to try to predict that situation and just do the targeted cleanup this is aimed at.

wbclark: Actually, I suppose it would come back anyway after the installer run, because `[]` is the default in puppet-foreman. So I think there is no point then.

ekohl (changes requested):
Would you mind fixing the merge conflict?

config/foreman.migrations/20210625142707_dynamic_puppet-in-foreman-groups.rb
```diff
@@ -0,0 +1,5 @@
+if answers['foreman'].is_a?(Hash)
```
ekohl: The file naming convention is to use underscores (except where we messed up)

config/katello.migrations/190314163941-remove-foreman-user-groups.rb
ekohl: Indeed, it should be taken care of by the new dynamic-puppet-in-foreman-groups. That said, I think we should remove the file entirely rather than keeping a dummy.

ekohl (changes requested):

config/foreman.migrations/20210625142707_dynamic_puppet-in-foreman-groups.rb
ehelms: This was what the script generated for me but I can change it.

ehelms:
Around whether to remove the migration file or not:

 a) I don't know what happens when you remove a migration that previously existed
 b) This creates a difference between a fresh install and an upgrade when looking at the what migrations have been applied that could lead to false positive issues being reported or folks going down a wrong path when investigating issues

This was my rationale for leaving the file and clearing the content.

ekohl:
> a) I don't know what happens when you remove a migration that previously existed

Luckily https://github.com/theforeman/kafo/blob/master/lib/kafo/migrations.rb is quite easy to follow. Most crucial is [this line](https://github.com/theforeman/kafo/blob/35a70342e2cc99d8ee9d58129e91efae66e98187/lib/kafo/migrations.rb#L22). It essentially is a list of (file)names that have already been applied and should be skipped. As such, it should be safe to remove the file. You should only be careful not to change/reuse migration names, but since a timestamp is included that shouldn't happen in practice.

>  b) This creates a difference between a fresh install and an upgrade when looking at the what migrations have been applied that could lead to false positive issues being reported or folks going down a wrong path when investigating issues

On the other hand, then you can know what they followed. The real migration or not. With this it'll be impossible to determine.

Given that, I'm in favor of removing the migration rather than keeping a dummy.

config/foreman.migrations/20210625142707_dynamic_puppet-in-foreman-groups.rb
ekohl: Please do change it.

ekohl (approved):
Thanks!

PR #692, opened 2021-06-30
Fixes #32914 - add puppet plugin to answer files

ezr-ondrej:


ekohl (commented):
Would you mind filing a Redmine issue for this? I think this may need to be cherry picked into 2.5.

ezr-ondrej:
> need to be cherry picked into 2.5

I'd not do that. For 2.5 the plugin was only for testing purposes, so it's ok there needed to be manual work done outside of installer. Given we would add that only in 2.5.2 I don't see a reason to, unless there are internal reasons :)

ekohl (approved):

ekohl (approved):
Test failure is expected.

ehelms:
I'd think we would want a migration as well to ensure the puppet plugin exists on upgraded machines.

ekohl:
Oh, that's a good point. Thanks for that.

PR #693, opened 2021-06-30
Fixes #32916 - add scenario overrides

evgeni:


evgeni:
```
Using /usr/local/rvm/gems/ruby-2.7.0 with gemset foreman-installer-pr-test-203-1-2-7-7-0

2021-06-30 14:50:43 [ERROR ] [root] Puppet 7.8.0 does not meet requirements for theforeman-puppet (>= 5.5.8 < 7.0.0)

2021-06-30 14:50:43 [ERROR ] [root] Incompatible version of Puppet used, cannot continue

Puppet 7.8.0 does not meet requirements for theforeman-puppet (>= 5.5.8 < 7.0.0)

script returned exit code 30
```

it wasn't me!

config/foreman-hiera.yaml
```diff
@@ -6,6 +6,8 @@ defaults:
 hierarchy:
   - name: "Custom user answers"
     path: "custom.yaml"
+  - name: "Scenario overrides"
+    path: "scenario/%{facts.kafo.scenario.id}.yaml"
```
ekohl: This may be a bit confusing since `scenario/%{facts.kafo.scenario.id}/family/%{facts.os.family}-%{facts.os.release.major}.yaml` is loaded at the lowest level and that's stored in the same directory.

Perhaps we can pick a name like `scenario/%{facts.kafo.scenario.id}/force-overrides.yaml` to make it explicit?

evgeni: Sure. The current code is what you did aeons ago for a workaround, so happy to make it cleaner.

ekohl: Another thing I'm wondering about is whether it should be above or below custom-hiera. Above it means we provide the user with no (supported) way to override them while below it means they still can. I'm slightly leaning to using above but maybe you have more thoughts.

evgeni: @ehelms @wbclark naming bikeshedding? I think I'd prefer `overrides.yaml` or `scenario-overrides.yaml` instead of `force`, but can live with it too ;)

ehelms: `these-overrides-are-law.yaml` ? `i-dont-care-what-the-answers-file-says.yaml` ? All kidding aside, I like `scenario-overrides.yaml`

ehelms:
> ```
> Using /usr/local/rvm/gems/ruby-2.7.0 with gemset foreman-installer-pr-test-203-1-2-7-7-0
> 
> 2021-06-30 14:50:43 [ERROR ] [root] Puppet 7.8.0 does not meet requirements for theforeman-puppet (>= 5.5.8 < 7.0.0)
> 
> 2021-06-30 14:50:43 [ERROR ] [root] Incompatible version of Puppet used, cannot continue
> 
> Puppet 7.8.0 does not meet requirements for theforeman-puppet (>= 5.5.8 < 7.0.0)
> 
> Cannot continue due to incompatible version of Puppet. Use --skip-puppet-version-check to disable this check.
> 
> script returned exit code 30
> ```
> 
> it wasn't me!

Yea, this is still a work in progress.

config/foreman-hiera.yaml
evgeni: > Another thing I'm wondering about is whether it should be above or below custom-hiera. Above it means we provide the user with no (supported) way to override them while below it means they still can. I'm slightly leaning to using above but maybe you have more thoughts.

Well, as so often with software design, "it depends"? ;-)

With the current "problem to solve", above custom-hiera sounds fine, as this is really something nobody should need to be able to override.
On the other hand, I like if software gives me a lever labeled "you shouldn't really be needing this, but if in hurry you can try it" as it allows to implement workarounds while a real solution is developed.
On yet another hand (Look behind you, a three handed monkey!) it's nice to be able to override the users overrides, as they get out of date (think migration of the olde custom-hiera tunings to tuning profiles).

That all said, I think for the "problem to solve" today it doesn't matter too much (logically I'd probably put it below, but nobody will be hurt if its above). But I can't tell what will be tomorrow.

ekohl: While I also like the `dont-tell-me-what-to-do.yaml` option perhaps we should have `enforced` for before `custom-hiera.yaml`. Then `overrides` for after `custom-hiera.yaml` but before `answers`.

evgeni: So, like this?

ekohl (approved):
We can bikeshed over if you need to include `scenario` in the filename again if it's already in the `scenario/$scenario` directory, but I'm ok with either.

evgeni:
> We can bikeshed over if you need to include `scenario` in the filename again if it's already in the `scenario/$scenario` directory, but I'm ok with either.

We can. I prefer the longer filename, as it's more descriptive, and you might not remember what folder you're in currently ;)

„What is this for?“ 
„It's a blue light.”
„What does it do?” 
„It turns blue.”

ehelms (approved):

PR #695, opened 2021-07-08
Refs #32914: Add migrations for foreman_puppet plugin

ehelms:
Currently, this is following existing behavior for new plugins. However, what is the expected behavior @ezr-ondrej  @ekohl ?  On upgrade, would we expect this plugin to be present and enabled?

Should these instead be written as:

```
unless answers.key?('foreman::plugin::puppet')
   answers['foreman::plugin::puppet'] = true
end
```

### DIFF
--- a/config/foreman.migrations/20210708144320_add_foreman_puppet.rb
+++ b/config/foreman.migrations/20210708144320_add_foreman_puppet.rb
@@ -1,0 +1,3 @@
+unless answers.key?('foreman::plugin::puppet')
+  answers['foreman::plugin::puppet'] = true
+end
--- a/config/foreman-answers.yaml
+++ b/config/foreman-answers.yaml
@@ -43,7 +43,7 @@ foreman::plugin::monitoring: false
 foreman::plugin::omaha: false
 foreman::plugin::openscap: false
 foreman::plugin::ovirt_provision: false
-foreman::plugin::puppet: false
+foreman::plugin::puppet: true
 foreman::plugin::puppetdb: false
 foreman::plugin::remote_execution: false
 foreman::plugin::remote_execution::cockpit: false
--- a/config/katello-answers.yaml
+++ b/config/katello-answers.yaml
@@ -53,7 +53,7 @@ foreman::plugin::leapp: false
 foreman::plugin::memcache: false
 foreman::plugin::monitoring: false
 foreman::plugin::openscap: false
-foreman::plugin::puppet: false
+foreman::plugin::puppet: true
 foreman::plugin::puppetdb: false
 foreman::plugin::remote_execution: true
 foreman::plugin::remote_execution::cockpit: false
--- a/config/katello.migrations/210708144422-add-foreman-puppet.rb
+++ b/config/katello.migrations/210708144422-add-foreman-puppet.rb
@@ -1,0 +1,3 @@
+unless answers.key?('foreman::plugin::puppet')
+  answers['foreman::plugin::puppet'] = true
+end
--- a/spec/fixtures/pulpcore-migration-dont-use-content-plugins-on-upgrades/katello-answers-after.yaml
+++ b/spec/fixtures/pulpcore-migration-dont-use-content-plugins-on-upgrades/katello-answers-after.yaml
@@ -18,6 +18,7 @@ foreman::plugin::kubevirt: false
 foreman::plugin::leapp: false
 foreman::plugin::memcache: false
 foreman::plugin::monitoring: false
+foreman::plugin::puppet: true
 foreman::plugin::remote_execution: true
 foreman::plugin::remote_execution::cockpit: false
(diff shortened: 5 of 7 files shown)


ekohl:
That's an excellent question. For users it would be smooth if we just update it that way but it does mean we should describe the removal process for upgrading users.

Perhaps @tbrisker can also weigh in here.

tbrisker:
I think it makes sense to default to not installing the plugin, as installing it after the upgrade if needed should be possible if i'm not mistaken. In any case, the release announcement for 3.0 should include clear upgrade instructions for both cases - i.e. what to do if you want to keep using the current puppet integration, as well as how to clean up the leftovers if you don't care about it.

ezr-ondrej:
As described in https://community.theforeman.org/t/puppet-in-3-0-release/24429 I believe we should default to plugin enabled for a transition period, as we still have manual reference including Puppet and we don't have alternative for Puppet certificates, apart of manually generating them, what is not the best default UX for users.

ezr-ondrej:
LGTM, I've got no idea what is the failure about tho :(

ekohl (changes requested):
I don't think this is right. We talked about enabling the plugin unconditionally so if you change this to true, it shouldn't need be modified:
https://github.com/theforeman/foreman-installer/blob/e7b5cdd7398ed2bedf567657190d2b6f70781861/config/foreman-answers.yaml#L46
https://github.com/theforeman/foreman-installer/blob/e7b5cdd7398ed2bedf567657190d2b6f70781861/config/katello-answers.yaml#L56

spec/migration_spec.rb
```diff
@@ -62,7 +62,7 @@
 
       it 'changes scenario answers' do
         _, after = migrator
-        expect(answers_after).to eq after
+        expect(after).to include answers_after
```
ekohl: Are these changes still needed?

ehelms: Yes. I either have to modify those fixtures which are targeting a specific chunk of the answers or make this change. This change feels more accurate.

config/foreman.migrations/20210708144320_add_foreman_puppet.rb
```diff
@@ -0,0 +1,3 @@
+unless answers.key?('foreman::plugin::puppet')
+  answers['foreman::plugin::puppet'] = true
```
ezr-ondrej, line 2: do we still need this? this plugin is new in the installer, so we should not need the migrations now.

ehelms: The migrations provide a way for upgraded systems to get the new module into the answers file. Without a migration, upgraded users would not have this entry in their answers file. You can think of the answers file like a database, from version to version we release we use migrations to update that schema. The default answers we deliver is as if we delivered the end state schema of the database.

ezr-ondrej (approved):
LGTM then

config/foreman.migrations/20210708144320_add_foreman_puppet.rb
ekohl: Essentially: if the answers file is missing, packaging will create it (with the then latest default). After that it's all migrations.

Also to be explicit, the migrations do run on the placed answers file. That's why we now use `unless answers.key?('foreman::plugin::puppet')`. If we just used `||= true` it would break in the future if we do want to flip the default to `false`. The use of `||= false` would be safe.

spec/migration_spec.rb
ekohl: I'm still torn between an exact match and this. If we go this route, why don't we write more precise tests? Let's take the remove-tuning-fact example:
```diff
--- foreman-proxy-content-before.yaml	2021-06-08 11:42:35.140954220 +0200
+++ foreman-proxy-content-after.yaml	2021-06-08 11:42:35.140954220 +0200
@@ -12,8 +12,9 @@
 :hook_dirs: ['./katello/hooks']
 :parser_cache_path: ./_build/parser_cache/foreman-proxy-content.yaml
 :hiera_config: ./config/foreman-hiera.yaml
-:facts:
-  tuning: 'default'
+
+:verbose: true
+:verbose_log_level: 'notice'
 
 :order:
   - certs
```

We could also stop including an after yaml file and simply state `expect(after).not_to contain(:facts)`. The reason I chose to include complete files is that I did like to see the complete effects of a change on various answer files.

Now I'll also say I'm generally not a fan of a lot of noise in the git log so I can be persuaded. I'm asking myself: how many issues can we catch with including the full files and how many will we miss if we don't.

However, having these partial matches with files that appear complete feels deceiving to me and a sort of worst of both worlds.

ares:
So just to be clear, are we ready to merge this one? From external observer point of view, it looks like it, but it's currently blocked by @ekohl requested changes.

ekohl (approved):

PR #696, opened 2021-07-12
2.5 puppet module updates for Pulp 3.14

ehelms:


ekohl (approved):
Linking back to another PR that should be merged together: https://github.com/theforeman/forklift/pull/1379

Then another thing we must not forget is to add a Conflict on pulpcore < 3.14 in the foreman-installer-katello RPM.

evgeni:
Given https://community.theforeman.org/t/katello-nightly-rpm-pipeline-1027-failed/24403, I am not sure that including puppet-pulpcore 4.0.0 is a good idea, as this will break proxy setups.

ehelms:
This is updated to include puppet-pulpcore 4.0.1 which fixes the Redis issue.

evgeni (approved):
ACK as this now contains `theforeman-pulpcore (4.0.1)` :)

PR #697, opened 2021-07-15
Fixes #32532: Ensure a valid tuning option is supplied

ehelms:
Adds a validator to the tuning option to ensure that what is supplied
on the command line is in fact one of the valid options.

(cherry picked from commit de9619fc9cc5dd0a964f8533f559aa026aebb2a9)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'kafo', '~> 6.0'
+gem 'kafo', '~> 6.4'
 gem 'librarian-puppet'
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'kafo', '~> 6.0'
+gem 'kafo', '~> 6.4'
```
ekohl, line 3: https://github.com/theforeman/foreman-packaging/pull/6890 is the RPM PR. Is there a matching Debian PR?

ekohl (approved):
:+1: to picking this, pending packaging picks of Kafo

Gemfile
ehelms: The Deb PR is -- https://github.com/theforeman/foreman-packaging/pull/6892

PR #698, opened 2021-07-17
Fixes #33046: Move scenarios to /var/lib/foreman-installer

ehelms:
This change moves scenarios from system config directory to the
shared state directory in /var/lib. The answers and config files
act as databases recording the current state of a scenario and
are not intended to be directly edited. Direct changes to the answers
file or configuration file can lead to inconsistencies or improper
values are that intended to be caught by validation. Moving to
the shared state directory draws this line better and emphasizes to
users that they should not be touching these files directly.

I wanted to see how much work this would be, and the impact and whether it was small enough to warrant consideration for Foreman 3.0 changes. I am however, completely OK with this being picked up for 3.1.

Projects that need checking:

- [ ] foreman-maintain -- https://github.com/theforeman/foreman_maintain/pull/508
- [ ] sosreport -- Needs update https://github.com/sosreport/sos/blob/master/sos/report/plugins/foreman_installer.py#L24
- [ ] satellite-clone
  * https://github.com/RedHatSatellite/satellite-clone/blob/0fd237703ea33ad98281a9fd07de1925799feb54/roles/satellite-clone/defaults/main.yml#L4
  * https://github.com/RedHatSatellite/satellite-clone/blob/f374e242911c5c41e9f6e65cb173530f0f948356/library/get_value_from_yaml_in_tarball.py
  * https://github.com/RedHatSatellite/satellite-clone/blob/bca1b16b404ade2e5f61841cdd020a002305d01b/roles/satellite-clone/tasks/backup_satellite_version_check.yml#L2
- [ ] change-hostname -- https://github.com/theforeman/foreman-packaging/pull/6910

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -77,7 +77,7 @@ SBINDIR = ENV['SBINDIR'] || "#{PREFIX}/sbin"
 INCLUDEDIR = ENV['INCLUDEDIR'] || "#{PREFIX}/include"
 SYSCONFDIR = ENV['SYSCONFDIR'] || "#{PREFIX}/etc"
 LOCALSTATEDIR = ENV['LOCALSTATEDIR'] || "#{PREFIX}/var"
-SHAREDSTAREDIR = ENV['SHAREDSTAREDIR'] || "#{LOCALSTATEDIR}/lib"
+SHAREDSTATEDIR = ENV['SHAREDSTATEDIR'] || "#{LOCALSTATEDIR}/lib"
 LOGDIR = ENV['LOGDIR'] || "#{LOCALSTATEDIR}/log"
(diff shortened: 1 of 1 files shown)


ekohl:
> I wanted to see how much work this would be, and the impact and whether it was small enough to warrant consideration for Foreman 3.0 changes. I am however, completely OK with this being picked up for 3.1.

While I'd like it to be in 3.0, I think given the time and other priorities it's more of a 3.1 thing.

> Projects that need checking:

I think there's also other tools in foreman-packaging.git. At least katello-change-hostname is one (though I think I fixed that by reading the scenario file).

ekohl:
For now I'm moving this back to a draft status, while I'd still love to see it happen.

PR #700, opened 2021-07-21
Use Kafo's migration DSL

ekohl:
This rewrites most Foreman migrations to a new migration DSL. The DSL decouples from the actual answer file structure and is more expressive.

Technically the DSL doesn't exist, but this at least makes sure that it's possible to express everything with the DSL I had in mind. As a reference to https://github.com/theforeman/kafo/pull/337.

### DIFF
--- a/config/foreman.migrations/20190111180118_delete_removed_settings.rb
+++ b/config/foreman.migrations/20190111180118_delete_removed_settings.rb
@@ -1,74 +1,10 @@
-DELETED = {
-  'foreman' => [
-    'custom_repo',
-    'dynflow_in_core',
-    'email_conf',
-    'email_config_method',
-    'email_source',
-    'puppet_home',
-    'puppet_ssldir',
-  ],
-  'foreman_proxy' => [
-    'custom_repo',
-    'puppetca_modular',
-    'realm_split_config_files',
-    'use_autosignfile',
-  ],
-  'foreman_proxy::plugin::dhcp::infoblox' => [
-    'use_ranges',
-  ],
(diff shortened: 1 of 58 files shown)


config/foreman.migrations/20190111180118_delete_removed_settings.rb
```diff
@@ -1,74 +1,10 @@
-DELETED = {
```
ekohl, line 1: I can be blamed for this. Kafo already removed unknown keys and this was never needed in the first place.

config/foreman.migrations/20190111180118_delete_removed_settings.rb
```diff
@@ -1,74 +1,10 @@
...
+  ['authentication', 'locations_enabled', 'organizations_enabled', 'repo'].each do |parameter|
+    mod.unset_answer(parameter)
   end
 end
 
-if (mod_answers = answers['foreman_proxy'])
-  mod_answers['dhcp_gateway'] = nil if mod_answers['dhcp_gateway'] == '192.168.100.1'
+migrate_module('foreman_proxy') do |mod|
+  mod.unset_answer('repo')
+  mod.unset_answer('dhcp_gateway') if mod['dhcp_gateway'] == '192.168.100.1'
```
ekohl, line 9: I thought about some more advanced DSL here:
```ruby
mod.migrate_answer('dhcp_gateway') do |param|
  param.unset_answer if param.value == '192.168.100.1'
end
```

However, I didn't want to overload things too soon.

PR #687, opened 2021-05-25
Fixes #32659 - Support Puppet 7

ekohl:
This is currently a draft, but it runs the tests to give a list of modules that need attention.

PR #703, opened 2021-07-26
Reflect changes for Puppetserver 7 support

ekohl:
This adds a migration to clear --puppet-server-ssl-chain-filepath if it's the old default. It also stops the puppetserver service if needed to migrate the CA directory.


ekohl:
Added to https://github.com/theforeman/foreman-installer/pull/687 instead

PR #704, opened 2021-07-28
Drop cdn_ssl_version hook

ehelms:
This data migration was introducedin Foreman 2.2, and is no longer
needed. There are no paths that would migrate directly from Foreman 2.2
to Foreman 3.0. This can safely be cleaned up.


ehelms:
[test foreman-installer]

ekohl (approved):

PR #705, opened 2021-07-29
Add ACD plugin

sbernhard:
* foreman ACD plugin
 * smart proxy ACD plugin

### DIFF
--- a/config/foreman-answers.yaml
+++ b/config/foreman-answers.yaml
@@ -25,6 +25,7 @@ foreman::compute::libvirt: false
 foreman::compute::openstack: false
 foreman::compute::ovirt: false
 foreman::compute::vmware: false
+foreman::plugin::acd: false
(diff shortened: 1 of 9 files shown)


sbernhard:
tests are failing because something else...

ekohl (changes requested):
You can call `./bin/create-migration` with multiple arguments and it'll create files with the same timestamps. This makes it clear that it's really the same migration.

You can also do both steps in a single file for `foreman.migrations` and `katello.migrations`.

The failure is resolved by https://github.com/theforeman/foreman-installer/pull/687 but there's one last thing I need to check with it. You can ignore it.

sbernhard:
Thanks @ekohl for your feedback. I hope I did all the requested changes.

ekohl (approved):
Thanks!

PR #706, opened 2021-07-30
Fixes #33191 - Debian puppet-agent-oauth reinstall support

ekohl:
This has been supported for a while on RPM-based distros. This adds it on Debian.

It also changes it to log an error on noop that it detected an incorrect situation.

Credits to @timdeluxe in https://github.com/theforeman/foreman-installer/pull/641 for the specific dpkg/apt commands. This PR takes a bit different implementation approach by also looking at facts. This should save some useless command executions on both platforms.


timdeluxe:
This implementation is much smarter than mine and has a smaller footprint on using system resources. Thumbs up!

ehelms (approved):

PR #709, opened 2021-08-03
Add hcf-puppet

amirfefer:


ekohl (changes requested):
You also need to add migrations. There's a `bin/create-migration` helper script which can be used.

config/foreman-answers.yaml
```diff
@@ -16,6 +16,7 @@ foreman::cli::azure: false
 foreman::cli::discovery: false
 foreman::cli::kubevirt: false
 foreman::cli::openscap: false
+foreman::cli:puppet: true
```
ekohl:
```suggestion
foreman::cli::puppet: true
```

config/foreman.migrations/20210803130619_add_hammer_puppet_plugin.rb
```diff
@@ -0,0 +1 @@
+answers['foreman::cli:puppet'] ||= true
```
ekohl:
```suggestion
answers['foreman::cli::puppet'] ||= true
```

ekohl (changes requested):
CI will fail until https://github.com/theforeman/puppet-foreman/pull/979 is merged.

ekohl:
[test foreman-installer]

config/foreman.migrations/20210803130619_add_hammer_puppet_plugin.rb
```diff
@@ -0,0 +1 @@
+answers['foreman::cli::puppet'] ||= true
```
ekohl, line 1: `||= false` would work well, but `||= true` doesn't. If we do decide to default to `false`, this migration will need modification. I'd prefer to do that ahead of time. From eb73fb7f4b2fe6470816a8707bae433e2b2cd71d:
```ruby
unless answers.key?('foreman::plugin::puppet')
  answers['foreman::plugin::puppet'] = true
end
```

So in this case
```suggestion
unless answers.key?('foreman::cli::puppet')
  answers['foreman::cli::puppet'] = true
end
```

ekohl (changes requested):
Please also have a look at the failing tests (they're related).

ekohl:
To move things along I submitted https://github.com/theforeman/foreman-installer/pull/710 with the changes I suggested.

PR #710, opened 2021-08-03
Add hammer-cli-foreman-puppet

ekohl:
This will break Debian nightly until hammer-cli-foreman 3.0.0 is out and https://github.com/theforeman/foreman-packaging/pull/6975 is merged. For now I'd like to accept that and continue with RPMs where we do have nightly RPMs.


ekohl:
<del>Hold off on merging, I want to sync the migration numbers.</del>

ehelms (approved):

PR #711, opened 2021-08-05
Update modules for 3.0 branching

tbrisker:


Puppetfile
```diff
@@ -1,26 +1,18 @@
...
-# Top-level modules
-mod 'theforeman/foreman',            :git => 'https://github.com/theforeman/puppet-foreman'
-mod 'theforeman/foreman_proxy',      :git => 'https://github.com/theforeman/puppet-foreman_proxy'
-mod 'theforeman/puppet',             :git => 'https://github.com/theforeman/puppet-puppet'
-
-# Top-level katello modules
-mod 'katello/foreman_proxy_content', :git => 'https://github.com/theforeman/puppet-foreman_proxy_content'
-mod 'katello/certs',                 :git => 'https://github.com/theforeman/puppet-certs'
-mod 'katello/katello',               :git => 'https://github.com/theforeman/puppet-katello'
+mod 'puppetlabs/xinetd',               '>= 3.3.0 < 3.4.0'
```
ekohl: This one must remain on git since there is no release that's Puppet 7 compatible. Yes, it is ironic that something in the puppetlabs namespace is the last to add Puppet 7 compatibility, but that's the way it is sadly.

ekohl (changes requested):

Puppetfile
tbrisker: 🤦🏼 updating

ekohl (approved):

PR #713, opened 2021-08-09
Refs #33214: Reset puma min threads to default from puppet module

ehelms:
If the default value for foreman_service_puma_threads_min is found,
which is zero, this deletes that value from the answers file. This
allows the new default to be populated from the puppet module. The
new default is to set min threads equal to max threads. If the minimum
is anything other than zero than we can assume the user has customized
the value.

### DIFF
--- a/config/foreman.migrations/20210809133823_reset_puma_min_threads.rb
+++ b/config/foreman.migrations/20210809133823_reset_puma_min_threads.rb
@@ -1,0 +1,6 @@
+if answers['foreman'].is_a?(Hash)
+  if answers['foreman']['foreman_service_puma_threads_min'] == 0 ||
+     answers['foreman']['foreman_service_puma_threads_min'] == answers['foreman']['foreman_service_puma_threads_max']
+    answers['foreman'].delete('foreman_service_puma_threads_min')
+  end
+end
--- a/config/katello.migrations/210809133829-reset-puma-min-threads.rb
+++ b/config/katello.migrations/210809133829-reset-puma-min-threads.rb
@@ -1,0 +1,6 @@
+if answers['foreman'].is_a?(Hash)
+  if answers['foreman']['foreman_service_puma_threads_min'] == 0 ||
+     answers['foreman']['foreman_service_puma_threads_min'] == answers['foreman']['foreman_service_puma_threads_max']
+    answers['foreman'].delete('foreman_service_puma_threads_min')
+  end
+end


wbclark (commented):
since there is no effective difference between setting foreman::foreman_service_puma_threads_min = 0 and foreman::foreman_service_puma_threads_min = foreman::foreman_service_puma_threads_max

should we therefore also clear it in the latter case to so that equivalent configurations actually look the same in answers?

the difference would be, that if we don't also clear it in that case, then an user who increases the max without also increasing the min, would run into memory leak issues again

ehelms:
Fair, I added a condition that if min = max to reset min to default of undef so allow min equal max as users tune the max.

wbclark (commented):
The more I think about this, the more I think we should simply reset the min for all users. This would only have a different impact for users who have previously supplied a custom configuration for the minimum number of threads. I see the tradeoff as follows:

1. Users who previously supplied a custom configuration are likely unaware of the memory leaks and other performance degradation associated with min < max, as it was not previously known/documented
2. Resetting to clear the min for all users means that if they pursue this configuration again, they are more likely to see the new parameter documentation mentioning the memory leaks and performance degradation
3. Users would still have the option of opting-in to that configuration if they really desire it even knowing this information
4. This could be slightly annoying to an user who really wants min < max, even knowing the downsides. But they would only have it reset once, since the migration won't run again.

For that reason, I think it's worth doing.

In the interest of being fully transparent, we could possibly log something here at WARN level, informing the user of the change and the reasoning behind it.

I think that offers the best balance between doing right by users, drawing attention to the new best practice, and still giving them total freedom in the end to opt back in to min < max if they really desire it.

ehelms:
> The more I think about this, the more I think we should simply reset the min for all users. This would only have a different impact for users who have previously supplied a custom configuration for the minimum number of threads. I see the tradeoff as follows:
> 
> 1. Users who previously supplied a custom configuration are likely unaware of the memory leaks and other performance degradation associated with min < max, as it was not previously known/documented
> 2. Resetting to clear the min for all users means that if they pursue this configuration again, they are more likely to see the new parameter documentation mentioning the memory leaks and performance degradation
> 3. Users would still have the option of opting-in to that configuration if they really desire it even knowing this information
> 4. This could be slightly annoying to an user who really wants min < max, even knowing the downsides. But they would only have it reset once, since the migration won't run again.
> 
> For that reason, I think it's worth doing.
> 
> In the interest of being fully transparent, we could possibly log something here at WARN level, informing the user of the change and the reasoning behind it.
> 
> I think that offers the best balance between doing right by users, drawing attention to the new best practice, and still giving them total freedom in the end to opt back in to min < max if they really desire it.

I understand the thinking behind this, I am hesitant to start changing user's configuration out from under them when we cannot make a clear reasoning about the state of things. Yes, users need to be aware of the tradeoffs between different configurations but this could be an unexpected change that goes undetected. Migrations are ran during RPM upgrade and thus less visible to users. Further, the current reports of memory leaks are only in Katello based scenarios. For all we know, pure Foreman installations are fine however a user has configured their environments.

wbclark (approved):
> I understand the thinking behind this, I am hesitant to start changing user's configuration out from under them when we cannot make a clear reasoning about the state of things. Yes, users need to be aware of the tradeoffs between different configurations but this could be an unexpected change that goes undetected. Migrations are ran during RPM upgrade and thus less visible to users. Further, the current reports of memory leaks are only in Katello based scenarios. For all we know, pure Foreman installations are fine however a user has configured their environments.

I find this reasoning pretty convincing regarding the design intent. The code itself looks good too.

ACK

PR #714, opened 2021-08-19
Refs #33277: Unset Puma defaults

ehelms:
I could be convinced to unset these values for everyone with a release note.

### DIFF
--- a/config/foreman.migrations/20210819164113_reset_puma_workers_and_max_threads.rb
+++ b/config/foreman.migrations/20210819164113_reset_puma_workers_and_max_threads.rb
@@ -1,0 +1,12 @@
+# https://github.com/theforeman/puppet-foreman/commit/533c1f3be8069139d7375841019afeb7c1102830
+if answers['foreman'].is_a?(Hash)
+  if answers['foreman']['foreman_service_puma_workers'] == 2
+    answers['foreman'].delete('foreman_service_puma_workers')
+  end
+end
+
+if answers['foreman'].is_a?(Hash)
+  if answers['foreman']['foreman_service_puma_threads_max'] == 16
+    answers['foreman'].delete('foreman_service_puma_threads_max')
+  end
+end
--- a/config/katello.migrations/210819164117-reset-puma-workers-and-max-threads.rb
+++ b/config/katello.migrations/210819164117-reset-puma-workers-and-max-threads.rb
@@ -1,0 +1,12 @@
+# https://github.com/theforeman/puppet-foreman/commit/533c1f3be8069139d7375841019afeb7c1102830
+if answers['foreman'].is_a?(Hash)
+  if answers['foreman']['foreman_service_puma_workers'] == 2
+    answers['foreman'].delete('foreman_service_puma_workers')
+  end
+end
+
+if answers['foreman'].is_a?(Hash)
+  if answers['foreman']['foreman_service_puma_threads_max'] == 16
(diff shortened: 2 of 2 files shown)


wbclark (commented):
Based on https://github.com/theforeman/puppet-foreman/pull/986#issuecomment-902190954 I'm going to recommend:

1. This migration should completely reset `foreman_service_puma_workers`, `foreman_service_puma_threads_min`, and `foreman_service_puma_threads_max`. A release note and Discourse post should be used to notify users of this change. They can still supply their preferred manual values after the migration has run with the one-time reset.

2. Tuning profiles should be updated based on the same calculations used to determine new defaults in puppet-foreman. Since the migration would reset manual configurations, users who don't make any further manual change would inherit configuration from their tuning profile.

3. Tuning profiles can be further updated based on new and improved performance data. As those updates take place, users who are now receiving their configuration from the tuning profile would automatically see the benefits of these further updates.

In my opinion, this provides the best experience for the majority of users. And for those who want to define their own values, who are already more likely to closely read the release notes and follow discussions related to Puma tuning, they will still have the ability to do so easily after the initial one-time reset.

config/foreman.migrations/20210819164456_reset_puma_max_threads.rb
```diff
@@ -0,0 +1,9 @@
+if answers['foreman'].is_a?(Hash)
+  if answers['foreman']['foreman_service_puma_threads_max'] == 16
+    answers['foreman'].delete('foreman_service_puma_threads_max')
+
+    if answers['foreman']['foreman_service_puma_threads_min'] == 16
```
ekohl: Why 16? I don't think we ever had that as default. I think it should be 0. I'm also inclined to keep it separate from the `threads_max` condition.

ekohl (commented):
One thing I didn't do for a while but you can find it in older migrations. That is to add a link to the PR/commit that actually requires the migration. I feel like that's a good practice.

ekohl (commented):
You dropped the `threads_min`? Maybe I should have given a code example, but I thought:
```ruby
if answers['foreman'].is_a?(Hash)
  if answers['foreman']['foreman_service_puma_threads_max'] == 16
    answers['foreman'].delete('foreman_service_puma_threads_max')
  end

  # https://github.com/theforeman/puppet-foreman/commit/172a16d0a0556b26c2c764f7339f292447cf62f2
  if answers['foreman']['foreman_service_puma_threads_min'] == 0
    answers['foreman'].delete('foreman_service_puma_threads_min')
  end
end
```

However, I now see that it should have a different Refs.

ehelms:
> You dropped the `threads_min`? Maybe I should have given a code example, but I thought:
> 
> ```ruby
> if answers['foreman'].is_a?(Hash)
>   if answers['foreman']['foreman_service_puma_threads_max'] == 16
>     answers['foreman'].delete('foreman_service_puma_threads_max')
>   end
> 
>   # https://github.com/theforeman/puppet-foreman/commit/172a16d0a0556b26c2c764f7339f292447cf62f2
>   if answers['foreman']['foreman_service_puma_threads_min'] == 0
>     answers['foreman'].delete('foreman_service_puma_threads_min')
>   end
> end
> ```
> 
> However, I now see that it should have a different Refs.

Already been handled here -- https://github.com/theforeman/foreman-installer/pull/713

ekohl (approved):
I'd have a slight preference to merge the two changes into a single file, given they both reflect the same commit in puppet-foreman.

PR #715, opened 2021-08-19
Fixes #33312 - Set Puma tuning values

ehelms:
This initially sets the Puma tuning values based on the same ideas
from https://github.com/theforeman/puppet-foreman/pull/986 statically
calculated for each of the CPU and memory values that exist for the
various profiles.

### DIFF
--- a/config/foreman.hiera/tuning/sizes/extra-extra-large.yaml
+++ b/config/foreman.hiera/tuning/sizes/extra-extra-large.yaml
@@ -11,3 +11,7 @@ postgresql::server::config_entries:
   work_mem: 8MB
   effective_cache_size: 32GB
   autovacuum_vacuum_cost_limit: 2000
+
+foreman::foreman_service_puma_workers: 72
+foreman::foreman_service_puma_threads_max: 5
+foreman::foreman_service_puma_threads_min: 5
--- a/config/foreman.hiera/tuning/sizes/extra-large.yaml
+++ b/config/foreman.hiera/tuning/sizes/extra-large.yaml
@@ -11,3 +11,7 @@ postgresql::server::config_entries:
   work_mem: 8MB
   effective_cache_size: 32GB
   autovacuum_vacuum_cost_limit: 2000
+
+foreman::foreman_service_puma_workers: 48
+foreman::foreman_service_puma_threads_max: 5
+foreman::foreman_service_puma_threads_min: 5
--- a/config/foreman.hiera/tuning/sizes/large.yaml
+++ b/config/foreman.hiera/tuning/sizes/large.yaml
@@ -8,3 +8,7 @@ postgresql::server::config_entries:
   shared_buffers: 8GB
   effective_cache_size: 16GB
   autovacuum_vacuum_cost_limit: 2000
+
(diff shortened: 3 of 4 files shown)


wbclark (commented):
Is there is a reason not to omit `foreman::foreman_service_puma_threads_min` ? If it is omitted, then the module will configure it equal to the maximum by default.

Arguably, including it here does make it easier for the user to understand what they are getting in each case?

ehelms:
> Is there is a reason not to omit `foreman::foreman_service_puma_threads_min` ? If it is omitted, then the module will configure it equal to the maximum by default.
> 
> Arguably, including it here does make it easier for the user to understand what they are getting in each case?

I looked at it as the tuning profiles are designed to be concrete bases of configuration. Users can vary values on top of that base but the bases are meant to be more tightly controlled. Thus being explicit with the settings seemed in line with that philosophy.

ekohl (changes requested):
This doesn't work. Answers are read before built ins (and tuning is considered a built in):
https://github.com/theforeman/foreman-installer/blob/dcd015ced528a36f5ad4534626e53d68a317ffc1/config/foreman-hiera.yaml#L13-L23

Hiera is also not used to determine defaults so it will first read the defaults from Puppet itself, store that in the answers and ignore these.

ehelms:
> This doesn't work. Answers are read before built ins (and tuning is considered a built in):
> https://github.com/theforeman/foreman-installer/blob/dcd015ced528a36f5ad4534626e53d68a317ffc1/config/foreman-hiera.yaml#L13-L23
> 
> Hiera is also not used to determine defaults so it will first read the defaults from Puppet itself, store that in the answers and ignore these.

How does it handle empty values? If the answers file has an empty value for the puma workers entry will that still take precedence over having a value set in the layer below such as this?

ekohl:
I think my information is outdated and I actually solved the limitation:
https://github.com/theforeman/kafo/commit/3b154b57768545d0a5044df36c5b1ccefe53fa65 was my start to making Puppet optional. I just never found the time to actually benefit from it.

It was originally my goal to separate out all scenario defaults so `katello-answers.yaml` contains nothing but a list of modules with true/false as value. All defaults should be in a separate layer. That way you can run `--enable-puppet` and benefit from those defaults. So roughly what https://github.com/theforeman/foreman-installer/pull/717 aims to achieve in the end.

ekohl:
So to clarify: right now I think this does work for fresh installations. However, for runtime switching (i.e., from existing medium to large) it won't change the answers. It'll be needed to write a hook that clears the answer, probably in `pre_values` but maybe earlier.

ehelms:
Could we skip writing empty values to the answer files? Would that solve this problem? Would it create other problems?

ekohl:
I think I found it. It's in this bit here:
https://github.com/theforeman/foreman-installer/blob/9e4d03cb917fde319bf3414c36a036d4c3ffdcad/config/foreman-proxy-content.migrations/200616155948-disable-pulp-3-proxying.rb#L11-L13

ekohl:
Oh, and I also thought about dropping a bunch of migrations instead but that's also a big task that I didn't want to start until I had time to finish it.

PR #717, opened 2021-08-24
Fixes #33337 - Do not enable Puppet infrastructure by default for Katello scenarios

ehelms:
This simplifies Katello's initial deployment to have Puppet off by default and matches what Foreman 3.1 will aim for. New installations will not get Puppet and users will need to make that choice. Upgrades will remain in tact. Users who wish to turn puppet on will need to run:

```
foreman-installer --scenario katello --enable-foreman-plugin-puppet --foreman-proxy-puppet true --foreman-proxy-puppetca true --foreman-proxy-content-puppet true --enable-puppet --puppet-server true --puppet-server-environments-owner apache --puppet-server-foreman-ssl-ca /etc/pki/katello/puppet/puppet_client_ca.crt --puppet-server-foreman-ssl-cert /etc/pki/katello/puppet/puppet_client.crt --puppet-server-foreman-ssl-key /etc/pki/katello/puppet/puppet_client.key
```

Upgrading users who wish to disable would run (though this needs to be tested):

```
foreman-installer --scenario katello --no-enable-foreman-plugin-puppet --foreman-proxy-puppet false --foreman-proxy-puppetca false --foreman-proxy-content-puppet false --no-enable-puppet
```

As I am not sure what disabling the plugin will od here as it would leave data. I will look to add some bats tests.

### DIFF
--- a/config/foreman-proxy-content-answers.yaml
+++ b/config/foreman-proxy-content-answers.yaml
@@ -13,6 +13,7 @@ certs:
   generate: false
 foreman_proxy_content:
   pulpcore_mirror: true
+  puppet: false
 foreman_proxy:
   foreman_ssl_ca: /etc/foreman-proxy/foreman_ssl_ca.pem
   foreman_ssl_cert: /etc/foreman-proxy/foreman_ssl_cert.pem
@@ -24,6 +25,8 @@ foreman_proxy:
   ssl_key: /etc/foreman-proxy/ssl_key.pem
   ssl_port: '9090'
   templates: true
+  puppet: false
+  puppetca: false
 foreman_proxy::plugin::acd: false
 foreman_proxy::plugin::ansible: false
 foreman_proxy::plugin::dhcp::infoblox: false
@@ -33,11 +36,4 @@ foreman_proxy::plugin::dns::infoblox: false
 foreman_proxy::plugin::openscap: false
 foreman_proxy::plugin::remote_execution::ssh: false
 foreman_proxy::plugin::shellhooks: false
-puppet:
-  server: true
-  server_foreman_ssl_ca: /etc/pki/katello/puppet/puppet_client_ca.crt
-  server_foreman_ssl_cert: /etc/pki/katello/puppet/puppet_client.crt
-  server_foreman_ssl_key: /etc/pki/katello/puppet/puppet_client.key
-  server_jvm_extra_args:
-    - "-Djruby.logger.class=com.puppetlabs.jruby_utils.jruby.Slf4jLogger"
-    - "-XX:ReservedCodeCacheSize=512m"
+puppet: false
--- a/katello_certs/hooks/boot/02-message-helpers.rb
+++ b/katello_certs/hooks/boot/02-message-helpers.rb
@@ -62,8 +62,7 @@ module KatelloCertsMessageHookContextExtension
                     --foreman-proxy-trusted-hosts                 "#{fqdn}"\\
                     --foreman-proxy-trusted-hosts                 "#{foreman_proxy_fqdn}"\\
                     --foreman-proxy-oauth-consumer-key            "#{foreman_oauth_key}"\\
-                    --foreman-proxy-oauth-consumer-secret         "#{foreman_oauth_secret}"\\
-                    --puppet-server-foreman-url                   "#{foreman_url}"
+                    --foreman-proxy-oauth-consumer-secret         "#{foreman_oauth_secret}"
 MSG
   end
 end
--- a/config/katello-answers.yaml
+++ b/config/katello-answers.yaml
@@ -29,7 +29,7 @@ foreman::cli::discovery: false
 foreman::cli::katello: true
 foreman::cli::kubevirt: false
 foreman::cli::openscap: false
-foreman::cli::puppet: true
+foreman::cli::puppet: false
 foreman::cli::remote_execution: true
 foreman::cli::tasks: false
 foreman::cli::templates: false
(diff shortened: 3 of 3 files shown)


ekohl:
> As I am not sure what disabling the plugin will od here as it would leave data. I will look to add some bats tests.

It would leave data and the Foreman plugin installed. So it would be safe and upgrades would probably continue to work but you're not removing it that way.

ekohl:
Some more thoughts: with this we need some instructions on how to enable Puppet. That should probably become a separate procedure in the docs.

ekohl (commented):
https://github.com/theforeman/foreman-installer/pull/715#issuecomment-905625031 reminded that my original goal was to introduce another scenario layer. In https://github.com/theforeman/foreman-installer/pull/718 I've roughly outlined this. Then it suddenly becomes a lot easier to flip defaults and is what my originally intended way to achieve this was.

config/foreman-proxy-content-answers.yaml
```diff
@@ -13,6 +13,7 @@ certs:
   generate: false
 foreman_proxy_content:
   pulpcore_mirror: true
+  puppet: false
```
ekohl, line 16: I've been thinking about this. It's this piece:
https://github.com/theforeman/puppet-foreman_proxy_content/blob/37a6e9eaa058fefd1788625ad5def7243cb351e5/manifests/init.pp#L319-L330

Perhaps we can make it `undef` by default and do:
```puppet
if pick($puppet, 'puppet' in lookup('classes', Array, undef, []) {
}
```

This would rely on Kafo presenting `classes` in Hiera to determine it dynamically. For users outside of Kafo it's still possible to explicit set it.

ehelms: I wondered if there was a way to do roughly the following in a hiera layer:

```
certs::puppet::hostname: "%{lookup($certs::foreman_proxy::hostname)}"
certs::puppet::generate: "%{lookup($puppet::server)}"
certs::puppet::deploy: "%{lookup($puppet::server)}"
```

And then use the order configuration to have:

```
:order:
  - certs
  - foreman
  - katello
  - foreman_proxy
  - foreman_proxy::plugin::pulp
  - foreman_proxy_content
  - certs::puppet
  - puppet
```

ekohl: https://github.com/theforeman/puppet-foreman_proxy_content/pull/383

As for your suggestion, I don't know if that would work. The integration is REALLY hacky now. Perhaps we should move it here instead:
https://github.com/theforeman/puppet-puppet/blob/7534cbe648d6576509303da7be4f7724ee2d1fa4/manifests/server/config.pp#L299-L315

That could get a `katello` boolean and include the class if needed. Let's see about that.

ehelms: I think including in puppet-puppet might get weird, in light of https://github.com/theforeman/puppet-certs/pull/370 and the desire to move certificate ownership into modules. 

I think we may generally face a problem with how all this works by properly splitting generation of certificates and deployment of certificates. If I am user providing my own certificates for Puppet, for Foreman then I have to generate those and hand them to the module parameters. Where we want to have a *default* certificate setup, where we default to tool X generating those certificates and then handing them to the right module parameters I have started to ask myself what should that look like. Today, we include our default certificate generation alongside deployment and work to ensure Puppet ordering helps things end up in the right places. But, perhaps, certificate generation should be either a separate step, separate command, or built into a hook that runs prior to the handoff to `puppet apply`.

ehelms:
Bats tests updates to go with this change https://github.com/theforeman/forklift/pull/1430

ehelms:
Related change: https://github.com/theforeman/foreman-packaging/pull/7181

ehelms:
Note: We will need this change available before we can merge this and the bats update --https://github.com/Katello/hammer-cli-katello/pull/818

katello_certs/hooks/boot/02-message-helpers.rb
```diff
@@ -62,8 +62,7 @@ def proxy_instructions_message(kafo)
                     --foreman-proxy-trusted-hosts                 "#{fqdn}"\\
                     --foreman-proxy-trusted-hosts                 "#{foreman_proxy_fqdn}"\\
                     --foreman-proxy-oauth-consumer-key            "#{foreman_oauth_key}"\\
-                    --foreman-proxy-oauth-consumer-secret         "#{foreman_oauth_secret}"\\
-                    --puppet-server-foreman-url                   "#{foreman_url}"
```
ekohl, line 66: Removing this does break it for Puppet. Should we keep a note for those users similar to what we've done in the hostname rename tool?

config/foreman-proxy-content-answers.yaml
ekohl: Marking it as resolved. https://github.com/theforeman/puppet-puppet/pull/812 is IMHO the path forward but for now we should accept the status quo.

scroogie:
It would be interesting to include in the Docs how to disable Puppet on Upgrade with Katello. I think that might be a goal for many existing users.

ehelms:
> It would be interesting to include in the Docs how to disable Puppet on Upgrade with Katello. I think that might be a goal for many existing users.

That is indeed part of the overall plan.

katello_certs/hooks/boot/02-message-helpers.rb
ehelms: The reason I chose to remove it is because this output is aimed at installing a brand new proxy. And in that scenario, where Puppet is disabled by default, a user needs to provide a lot more options than this option to enable Puppet properly. So I figured for the default case, we should drop this instruction and users will follow the docs on enabling Puppet to set it up properly on a smart proxy.

ekohl: Fair enough.

ehelms:
The hammer-cli-katello change is in now so this should be good to go alongside https://github.com/theforeman/forklift/pull/1430 where I think the bats change can go in stand-alone first.

ekohl (approved):
:+1: pending the pipeline merge. How are we going to deal with documentation? We need that if people are going to verify functionality with our release candidates.

PR #718, opened 2021-08-25
Externalize answer defaults

ekohl:
This moves all non-default options into the Hiera layer. Since Kafo 3.0 this works and is supported. It also means --reset-$option loads the installer default rather than the Puppet default.

Tests need modifications so this is still a draft. It's also untested but it's relevant to https://github.com/theforeman/foreman-installer/pull/717 and https://github.com/theforeman/kafo/pull/339.

### DIFF
--- a/config/foreman-hiera.yaml
+++ b/config/foreman-hiera.yaml
@@ -15,6 +15,7 @@ hierarchy:
   - name: "Built in"
     paths:
       - "scenario/%{facts.kafo.scenario.id}/family/%{facts.os.family}-%{facts.os.release.major}.yaml"
+      - "scenario/%{facts.kafo.scenario.id}/defaults.yaml"
       - "family/%{facts.os.family}-%{facts.os.release.major}.yaml"
       - "family/%{facts.os.family}.yaml"
       - "security.yaml"
--- a/config/foreman-proxy-content-answers.yaml
+++ b/config/foreman-proxy-content-answers.yaml
@@ -9,21 +9,9 @@
 #
 # See params.pp in each class for what options are available
 ---
-certs:
-  generate: false
-foreman_proxy_content:
-  pulpcore_mirror: true
-foreman_proxy:
-  foreman_ssl_ca: /etc/foreman-proxy/foreman_ssl_ca.pem
-  foreman_ssl_cert: /etc/foreman-proxy/foreman_ssl_cert.pem
-  foreman_ssl_key: /etc/foreman-proxy/foreman_ssl_key.pem
-  http: true
-  manage_puppet_group: false
-  ssl_ca: /etc/foreman-proxy/ssl_ca.pem
-  ssl_cert: /etc/foreman-proxy/ssl_cert.pem
-  ssl_key: /etc/foreman-proxy/ssl_key.pem
-  ssl_port: '9090'
-  templates: true
+certs: true
+foreman_proxy_content: true
+foreman_proxy: true
 foreman_proxy::plugin::acd: false
 foreman_proxy::plugin::ansible: false
 foreman_proxy::plugin::dhcp::infoblox: false
@@ -33,11 +21,4 @@ foreman_proxy::plugin::dns::infoblox: false
 foreman_proxy::plugin::openscap: false
 foreman_proxy::plugin::remote_execution::ssh: false
 foreman_proxy::plugin::shellhooks: false
-puppet:
-  server: true
-  server_foreman_ssl_ca: /etc/pki/katello/puppet/puppet_client_ca.crt
-  server_foreman_ssl_cert: /etc/pki/katello/puppet/puppet_client.crt
-  server_foreman_ssl_key: /etc/pki/katello/puppet/puppet_client.key
-  server_jvm_extra_args:
-    - "-Djruby.logger.class=com.puppetlabs.jruby_utils.jruby.Slf4jLogger"
-    - "-XX:ReservedCodeCacheSize=512m"
+puppet: true
--- a/config/katello-answers.yaml
+++ b/config/katello-answers.yaml
@@ -9,19 +9,8 @@
 #
 # See params.pp in each class for what options are available
 ---
-certs:
-  group: foreman
-foreman:
-  client_ssl_ca: /etc/foreman/proxy_ca.pem
-  client_ssl_cert: /etc/foreman/client_cert.pem
-  client_ssl_key: /etc/foreman/client_key.pem
-  initial_location: Default Location
-  initial_organization: Default Organization
-  server_ssl_ca: /etc/pki/katello/certs/katello-default-ca.crt
-  server_ssl_cert: /etc/pki/katello/certs/katello-apache.crt
-  server_ssl_chain: /etc/pki/katello/certs/katello-server-ca.crt
-  server_ssl_crl: ""
-  server_ssl_key: /etc/pki/katello/private/katello-apache.key
+certs: true
+foreman: true
 foreman::cli: true
 foreman::cli::ansible: false
 foreman::cli::azure: false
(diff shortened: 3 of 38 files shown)


config/foreman-hiera.yaml
```diff
@@ -15,6 +15,7 @@ hierarchy:
   - name: "Built in"
     paths:
       - "scenario/%{facts.kafo.scenario.id}/family/%{facts.os.family}-%{facts.os.release.major}.yaml"
+      - "scenario/%{facts.kafo.scenario.id}/defaults.yaml"
```
ehelms, line 18: Since there is duplication in some cases, that makes me wonder if we should have:

```
      - "scenario/%{facts.kafo.scenario.id}/defaults.yaml"
      - "scenario/defaults.yaml"
```

ehelms:
Seems like with this change, and combined with https://github.com/theforeman/kafo/pull/338 handling enable/disable support we could get rid of the need for the answers file all together and rely on pure hiera constructs + a config file.

config/foreman-hiera.yaml
ekohl: While the duplication is a good point, I think I'd prefer to add it to `common.yaml` instead.

ehelms:
I think there is one fundamental I am missing as to how this works, and whether its a feature of Kafo or a feature of Hiera. What keeps the value from the answer file from overriding the default? The example I could think of was the puppet server args. In the defaults layer we'd have:

```
puppet::server_jvm_extra_args:
    - "-Djruby.logger.class=com.puppetlabs.jruby_utils.jruby.Slf4jLogger"
    - "-XX:ReservedCodeCacheSize=512m"
```

Then, in the layer that is higher up and represented by the answer file the following would exist:

```
puppet::server_jvm_extra_args:
```

What prevents the empty value higher up coming from the answer file (due to the default from the puppet module) from clobbering the default values set for the scenario?

ekohl:
Kafo still works in the same way: defaults are copied to the answers file. It's just that since Kafo 3, Hiera is considered while determining defaults. I thought I wrote this down in the other PR, but looks like I didn't since the commit message I linked explained it already.

The key is that the answers layer is never written when loading defaults. Compare [defaults](https://github.com/theforeman/kafo/blob/35a70342e2cc99d8ee9d58129e91efae66e98187/lib/kafo/configuration.rb#L188-L191) and [actual](https://github.com/theforeman/kafo/blob/35a70342e2cc99d8ee9d58129e91efae66e98187/lib/kafo/kafo_configure.rb#L486-L496); note that `store_answers` is only in the latter.

Now you can modify [merge behavior](https://puppet.com/docs/puppet/7/hiera_merging.html) in Hiera. We do it here:
https://github.com/theforeman/foreman-installer/blob/4c129efe5a35d8e87da7483e3ad3bf53c110ed76/config/foreman.hiera/tuning/common.yaml#L2-L4
However, I'm not sure how useful that would be (except for hashes).

ehelms:
I think you probably did explain it previously, that doesn't mean I understand all the nuances. So I am still confused a bit.

Hiera is considered when calculating the defaults. Is Hiera also considered when running the actual `puppet apply` ? Does that mean on an initial installation some parameters from Hiera are 'elevated' to the answer file as default values and permanently stored and others that are above the answers file in the hierarchy just simply take precedence when the apply is run?

ekohl:
https://github.com/theforeman/kafo#default-values does go over this, but it's a bit brief on implementation details.

The first step it talks about (a static value) is done using puppet-strings. We have a JSON cache that's shipped in our installer with this info. If Kafo needs to determine a default value and it's static, this cache is sufficient.

In the second case where there is no default, it extracts it by dumping values. https://github.com/theforeman/kafo/blob/master/modules/kafo_configure/manifests/dump_values.pp does that and Kafo then parses the output of that. It runs Puppet in noop mode by including all classes that it think it needs. As you can imagine, this is a lot slower than loading values from a JSON cache (which is why that step in the installer always takes a while). Since now it's real Puppet with Hiera configured (just no answers written down), our custom hierarchy is considered.

At least, this is the theory. I must admit it's been a while since I worked on it and I'm not 100% sure this all just works. Perhaps it may still see `$module::params::var` and optimize that to only including `$module::params` and dumping that (because only loading params.pp is a **lot** faster than the entire class).

Previously I've looked at alternatives. There is `puppet catalog compile` which can just compile a catalog (without applying it) and return it in JSON but I'm not sure if that contains the expanded variables.

ekohl:
I've split off preparation in https://github.com/theforeman/foreman-installer/pull/719.

config/katello-answers.yaml
```diff
@@ -69,15 +58,7 @@ foreman::plugin::tasks: true
...
-foreman_proxy:
-  foreman_ssl_ca: /etc/foreman-proxy/foreman_ssl_ca.pem
-  foreman_ssl_cert: /etc/foreman-proxy/foreman_ssl_cert.pem
-  foreman_ssl_key: /etc/foreman-proxy/foreman_ssl_key.pem
-  manage_puppet_group: false
-  ssl_ca: /etc/foreman-proxy/ssl_ca.pem
-  ssl_cert: /etc/foreman-proxy/ssl_cert.pem
-  ssl_key: /etc/foreman-proxy/ssl_key.pem
-  ssl_port: '9090'
+foreman_proxy: {}
```
ehelms: I think we should aim to be consistent with either using true or an empty hash with this overall change.

ekohl: I think so too, but I didn't want to go back and change all migrations to deal with true instead of assuming a hash. Though it looks like I may have missed some already.

ekohl:
I don't quite understand why the tests pass on my machine but don't here. (I was actually surprised they passed)

ekohl:
Turns out I had a `./config/foreman.migrations/.applied` file. That's why the migrations didn't run for me.

spec/migration_spec.rb
```diff
@@ -138,7 +137,7 @@
 
       it 'changes scenario answers' do
         _, after = migrator
-        expect(after).to include answers_after
+        expect(after['foreman']['user_groups']).to eq([])
```
ekohl, line 140: This was needed because without it, it also found an additional entry for the email method. However, the test is about the user groups being cleaned so this is still testing what it's supposed to.

config/foreman-proxy-content-answers.yaml
```diff
@@ -9,21 +9,9 @@
...
-  foreman_ssl_key: /etc/foreman-proxy/foreman_ssl_key.pem
-  http: true
-  manage_puppet_group: false
-  ssl_ca: /etc/foreman-proxy/ssl_ca.pem
-  ssl_cert: /etc/foreman-proxy/ssl_cert.pem
-  ssl_key: /etc/foreman-proxy/ssl_key.pem
-  ssl_port: '9090'
-  templates: true
+certs: true
+foreman_proxy_content: {}
```
ekohl: When I switch this to true the tests don't pass anymore. Some Pulp 2 migration changes it to a hash (with entries) and then a later one removes the params again. so it becomes `true != {}`

ekohl:
I tried this out and ran into this issue:

```console
# foreman-installer --scenario katello --enable-puppet --foreman-proxy-puppet true --foreman-proxy-puppetca true --foreman-proxy-content-puppet true -s --disable-system-checks
2021-08-31 18:07:55 [NOTICE] [root] Loading installer configuration. This will take some time.
2021-08-31 18:08:00 [NOTICE] [root] Running installer with log based terminal output at level NOTICE.
2021-08-31 18:08:00 [NOTICE] [root] Use -l to set the terminal output log level to ERROR, WARN, NOTICE, INFO, or DEBUG. See --full-help for definitions.
2021-08-31 18:08:06 [WARN  ] [pre] Skipping system checks.
2021-08-31 18:08:06 [WARN  ] [pre] Skipping system checks.
2021-08-31 18:08:11 [NOTICE] [configure] Starting system configuration.
2021-08-31 18:08:25 [NOTICE] [configure] 250 configuration steps out of 1712 steps complete.
2021-08-31 18:08:25 [ERROR ] [configure] Could not set groups on user[foreman-proxy]: Execution of '/sbin/usermod -G puppet foreman-proxy' returned 6: usermod: group 'puppet' does not exist
2021-08-31 18:08:25 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Config/User[foreman-proxy]/groups: change from  to 'puppet' failed: Could not set groups on user[foreman-proxy]: Execution of '/sbin/usermod -G puppet foreman-proxy' returned 6: usermod: group 'puppet' does not exist
2021-08-31 18:08:27 [NOTICE] [configure] 500 configuration steps out of 1712 steps complete.
2021-08-31 18:08:28 [NOTICE] [configure] 750 configuration steps out of 1716 steps complete.
2021-08-31 18:08:31 [NOTICE] [configure] 1000 configuration steps out of 1723 steps complete.
2021-08-31 18:08:33 [NOTICE] [configure] 1250 configuration steps out of 1724 steps complete.
2021-08-31 18:09:18 [NOTICE] [configure] 1500 configuration steps out of 1724 steps complete.
2021-08-31 18:09:28 [NOTICE] [configure] System configuration has finished.

  There were errors detected during install.
  Please address the errors and re-run the installer to ensure the system is properly configured.
  Failing to do so is likely to result in broken functionality.

```

This is because the server isn't getting enabled. So my theory of the loading doesn't appear to be correct. I think it continues to read it the cached JSON if available and falls back to Hiera only when it can't.

So my whole theory about having fixed it was incorrect. Perhaps that's why I completely forgot about it: I never tested it and only did the preparation work with building an `ExecutionEnvironment` which allowed it.

Another theory is that previously it may have worked when we used `$puppet::params::server_ca` but now that a concrete value is available in puppet-strings' JSON cache that's preferred.

PR #722, opened 2021-09-07
Move default ssl-build directory to /var/lib/foreman-installer

ehelms:
The idea is to move away from using `/root/ssl-build` which users often do not like, and has a tendency to get deleted accidentally to a more appropriate location in `/var/lib/foreman-installer`. This should be less prone to accidental removal and a better location for backup predictability, for a mounted volume and generally a better design. The tricky bit is what to do about existing `/root/ssl-build` directories. 

For upgrades, this attempts to move them as part of the boot hook if they are found to the new location. The alternative approach would be to change it for new installs only and leave it for upgrades. This, however, creates a split experience for our users base and makes debugging harder.

### DIFF
--- a/config/foreman-proxy-content-answers.yaml
+++ b/config/foreman-proxy-content-answers.yaml
@@ -11,6 +11,7 @@
 ---
 certs:
   generate: false
+  ssl_build_dir: '/var/lib/foreman-installer/ssl-build'
 foreman_proxy_content:
   pulpcore_mirror: true
 foreman_proxy:
--- a/config/katello-answers.yaml
+++ b/config/katello-answers.yaml
@@ -11,6 +11,7 @@
 ---
 certs:
   group: foreman
+  ssl_build_dir: '/var/lib/foreman-installer/ssl-build'
 foreman:
   client_ssl_ca: /etc/foreman/proxy_ca.pem
   client_ssl_cert: /etc/foreman/client_cert.pem
--- a/hooks/pre/35-move_ssl_build_dir.rb
+++ b/hooks/pre/35-move_ssl_build_dir.rb
@@ -1,0 +1,12 @@
+old_ssl_build_dir = '/root/ssl-build'
+new_ssl_build_dir = '/var/lib/foreman-installer/ssl-build'
+
+if module_enabled?('certs')
+  if param('certs', 'ssl_build_dir') == old_ssl_build_dir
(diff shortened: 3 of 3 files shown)


hooks/boot/30-move_ssl_build_dir.rb
```diff
@@ -0,0 +1,12 @@
+old_ssl_build_dir = '/root/ssl-build'
+new_ssl_build_dir = '/var/lib/foreman-installer/ssl-build'
+
+if module_enabled?('certs')
+  if param('certs', 'ssl_build_dir') == old_ssl_build_dir
+    if File.exist?(old_ssl_build_dir) && !File.exist?(new_ssl_build_dir)
```
ekohl: What if the new build dir does exist?

ekohl (commented):
Why is this part of the boot hook? That means it ends up being moved during `yum upgrade` already since that runs migrations. It also means it doesn't respect `--noop` and I suspect you can get some weird interactions where the directory is moved but the answers file isn't saved.

We should also make sure that sosreport doesn't need an update.

ehelms:
> Why is this part of the boot hook?

Boot happens only after migrations, and that is why I chose it rather than actual migrations (https://github.com/theforeman/kafo/blob/master/lib/kafo/kafo_configure.rb#L140). I did not want this to happen during package upgrade as that felt prone to problems. And boot seemed the best hook to do it in, as it would happen before user input. 

I can make updates to handle `--noop`.

I first wanted to float this general idea out there to gather an idea if others thought this was a good change to make.

ekohl:
> Boot happens only after migrations, and that is why I chose it rather than actual migrations (https://github.com/theforeman/kafo/blob/master/lib/kafo/kafo_configure.rb#L140). I did not want this to happen during package upgrade as that felt prone to problems. And boot seemed the best hook to do it in, as it would happen before user input.

Looking at https://raw.githubusercontent.com/theforeman/kafo/master/doc/kafo_run.png a `pre_migrations` (if you want to run it via migrations) or `pre` (otherwise) hook feels better than a `boot` hook. Boot hooks are really intended for setting up the application, not doing any actual work.

> I first wanted to float this general idea out there to gather an idea if others thought this was a good change to make.

Overall I'd be in favor. `/root/ssl-build` always felt weird to me. We can debate `/var/lib/foreman-installer` vs `/var/lib/ssl-build` or something else, but `/var/lib` does feel like a better home.

PR #723, opened 2021-09-28
Fixes #33577 - Only enable Foreman Puppet plugin with Foreman

ekohl:
If the user passed in `--no-enable-foreman` then the Foreman plugin should not be enabled.

### DIFF
--- a/config/katello.migrations/170331152302-add-cli.rb
+++ b/config/katello.migrations/170331152302-add-cli.rb
@@ -1,1 +1,3 @@
-answers['foreman::cli'] ||= true
+unless answers.key?('foreman::cli')
+  answers['foreman::cli'] = true
+end
--- a/spec/migration_spec.rb
+++ b/spec/migration_spec.rb
@@ -142,4 +142,38 @@ describe 'migrations' do
       end
     end
   end
+
+  context 'disable puppet if needed' do
+    %w[foreman katello].each do |scenario_name|
+      context "on #{scenario_name}" do
+        let(:answers) do
+          {
+            'foreman' => false,
+            'foreman::cli' => false,
+          }
+        end
+        let(:scenario_name) { scenario_name }
+        let(:config) { load_config_yaml("#{scenario_name}.yaml") }
+        let(:migrations) { config_path("#{scenario_name}.migrations") }
+        let(:migrator) { Kafo::Migrations.new(migrations).run(config, answers) }
+        subject(:migrated_answers) { migrator[1] }
+
+        it 'keeps foreman::cli disabled' do
+          expect(migrated_answers['foreman::cli']).to be false
+        end
+
+        it 'adds foreman::cli::puppet disabled' do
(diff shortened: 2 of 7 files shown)


ekohl:
I chose to change the existing migration so we don't end up in a bad state, but I wonder if we should instead create another migration as well that disables it if foreman itself is disabled. Users who are affected couldn't have upgraded to 3.0.0 (https://community.theforeman.org/t/foreman-proxy-3-0-installation-fails-with-parameter-config-file-group-expects-a-string-value/25438) so perhaps it's ok, but it may not be the cleanest.

wbclark (commented):
I would personally vote for a new migration that disables the plugin if foreman itself is disabled, for the reason that it would reduce the number of accidentally broken configurations in the wild (vs. the solution here which stops adding to them but doesn't correct those which already exist)

Perhaps the difference is minimal or low impact but I like the practice of trying to reduce that drift wherever possible, as that can prevent future problems.

ekohl:
Updated. Now also contains the CLI part of it (as reported by the user) and tests. I've started a new pattern for the tests and want to look at using for all other tests. The reporting of what failed is much clearer to me. We can also pull the `let` blocks to a higher level to share them between tests.

spec/migration_spec.rb
```diff
@@ -142,4 +142,38 @@
...
+          expect(migrated_answers['foreman::cli::puppet']).to be false
+        end
+
+        it 'keeps foreman disabled' do
+          expect(migrated_answers['foreman']).to be false
+        end
+
+        it 'adds foreman::plugin::puppet disabled' do
+          expect(migrated_answers['foreman::plugin::puppet']).to be false
+        end
```
wbclark, line 175: :+1: :+1: :+1: 

Fully agreed about using this approach to make test output more readable.

config/katello.migrations/170331152302-add-cli.rb
```diff
@@ -1 +1,3 @@
-answers['foreman::cli'] ||= true
+unless answers.key?('foreman::cli')
+  answers['foreman::cli'] = true
+end
```
wbclark, line 3: I am guessing there is some subtle difference here. Can you elaborate on the change for this older migration?

ekohl: What `||=` does is that if the value evaluates `false` (so `false` or `nil`), it sets the value to `true`.

So previously this happened:
```ruby
answers['foreman::cli'] = false # from the answers
answers['foreman::cli'] ||= true # from the migration
```

So the result was that it was always enabled while the intended migration was to add it with a default value of `true`.

In https://github.com/theforeman/foreman-installer/pull/700 I played around with implementing a hypothetical DSL for migrations that would convey the intention of the migration much better.

wbclark: Ah, I understood the behavior of `||=` with `nil` but not `false`. Now I see why this was broken. TY

wbclark (approved):

PR #727, opened 2021-10-21
Fixes #33760 - Add host_reports plugins

ofedoren:
Requires:
 - [x] https://github.com/theforeman/puppet-foreman/pull/1000
 - [x] https://github.com/theforeman/puppet-foreman_proxy/pull/707


ofedoren:
@ekohl, can we make it to be available for 3.1?

ekohl (approved):
The code looks good. This will continue failing CI until the Puppet modules are finished (which in turn depends on packaging as well). The Redmine issue is aligned to 3.1.0 so it shouldn't be missed before we tag RC1.

ekohl:
[test foreman-installer]

PR #728, opened 2021-10-26
Fixes #33109 - Update katello-certs-check date format

BirkirFreyr:
Fixes #33109 - Date used by katello-certs-check should be formatted correctly by program
Unset LC_ALL and set LC_TIME to C before running date command to get correct output for subsequent commands.

### DIFF
--- a/bin/katello-certs-check
+++ b/bin/katello-certs-check
@@ -95,7 +95,7 @@ function check-server-cert-encoding () {
 }
 
 function check-expiration () {
-    DATE=$(date -u +"%b %-d %R:%S %Y")
+    DATE=$(LC_ALL= LC_TIME=C date -u +"%b %-d %R:%S %Y")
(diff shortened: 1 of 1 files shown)


bin/katello-certs-check
```diff
@@ -95,7 +95,7 @@ function check-server-cert-encoding () {
 }
 
 function check-expiration () {
-    DATE=$(date -u +"%b %-d %R:%S %Y")
+    DATE=$(LC_ALL= LC_TIME=C date -u +"%b %-d %R:%S %Y")
```
ekohl: On EL8 there is no `C` locale, just `C.UTF-8`.

Having said that, I really don't understand this method at all. It's only used in `DATE_TODAY` so can't we just optimize that to `DATE_TODAY=$(date +%Y%m%d%H%M%S)` and get rid of all locales? Am I missing something?

BirkirFreyr: If i'm being perfectly honest, i didn't look into this that much, just got it to run without outputting a format error on my system. But you have a valid point, i'll look into that change and double-check that DATE isn't used elsewhere and update accordingly

PR #731, opened 2021-11-08
Refs #33864 - Disable registration on upgrades

ekohl:
In Foreman 3.1 the registration module requires templates where it previously worked standalone.

In a Foreman or Katello scenario the registration module doesn't make sense since the clients can connect directly to a Foreman, which makes the indirection redundant. The Smart Proxy may also not have plain HTTP enabled which then breaks things.

It is explicitly enabled by default in the foreman-proxy-content scenario because there templates is enabled by default, which makes it safe to enable registration.

### DIFF
--- a/config/foreman-proxy-content-answers.yaml
+++ b/config/foreman-proxy-content-answers.yaml
@@ -24,6 +24,7 @@ foreman_proxy:
   ssl_cert: /etc/foreman-proxy/ssl_cert.pem
   ssl_key: /etc/foreman-proxy/ssl_key.pem
   ssl_port: '9090'
+  registration: true
   templates: true
   puppet: false
   puppetca: false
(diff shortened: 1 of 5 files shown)


evgeni:
Rubocop is unhappy, but otherwise this looks like what we discussed!

@ehelms you raised concerns earlier, mind having a look?

ehelms:
Copying what I said on the puppet-foreman_proxy PR, as this does probably apply more to here since it determines the default installation:

I wanted to note that this does create a bit of a pain in the horse for tooling. If I create an Ansible module or CLI tool to perform registration, I would have to ensure I have code branches to call the API directly to Foreman or the Smart Proxy end point depending on some condition that determines who or what I am talking to. For me, that breaks this idea conceptually a bit that I talk to a smart-proxy for registration and it handles the rest no matter where its deployed.

ehelms:
[test foreman-installer]

ehelms:
Can you fix up the linting issue?

ekohl:
Updated

ekohl:
:green_heart: now

ehelms (approved):

PR #733, opened 2021-11-11
Pin modules for 3.1 branching

upadhyeammit:


Puppetfile.lock
```diff
@@ -0,0 +1,122 @@
...
+      puppetlabs-xinetd (>= 1.1.0, < 4.0.0)
+
+DEPENDENCIES
+  katello-candlepin (>= 12.1.0, < 12.2.0)
+  katello-certs (>= 14.0.0, < 14.1.0)
+  katello-foreman_proxy_content (>= 20.1.0, < 20.2.0)
+  katello-katello (>= 21.2.0, < 21.3.0)
+  katello-qpid (>= 9.1.0, < 9.2.0)
+  puppetlabs-postgresql (>= 7.0.0)
+  puppetlabs-xinetd (>= 3.3.0, < 3.4.0)
```
upadhyeammit: It seems we will need to bump `puppetlabs-xinetd` which means bump also required for `theforeman-tftp`
~~~
Puppet 7.12.1 does not meet requirements for puppetlabs-xinetd (>= 4.7.1 < 7.0.0)
~~~

upadhyeammit:
@ehelms / @ekohl request to review

Puppetfile
```diff
@@ -1,26 +1,18 @@
...
-# Top-level modules
-mod 'theforeman/foreman',            :git => 'https://github.com/theforeman/puppet-foreman'
-mod 'theforeman/foreman_proxy',      :git => 'https://github.com/theforeman/puppet-foreman_proxy'
-mod 'theforeman/puppet',             :git => 'https://github.com/theforeman/puppet-puppet'
-
-# Top-level katello modules
-mod 'katello/foreman_proxy_content', :git => 'https://github.com/theforeman/puppet-foreman_proxy_content'
-mod 'katello/certs',                 :git => 'https://github.com/theforeman/puppet-certs'
-mod 'katello/katello',               :git => 'https://github.com/theforeman/puppet-katello'
+mod 'puppetlabs/xinetd',               '>= 3.3.0 < 3.4.0'
```
ehelms: This should remain the same as it is in `develop`:

```
mod 'puppetlabs/xinetd',             :git => 'https://github.com/puppetlabs/puppetlabs-xinetd', :commit => 'e742608dccdf42236144acf9f05e483b47c576f1'
```

Puppetfile.lock
ehelms: See https://github.com/theforeman/foreman-installer/pull/733/files#r747798250

Puppetfile
```diff
@@ -1,26 +1,18 @@
 forge 'https://forgeapi.puppetlabs.com/'
 
-# Dependencies
-# For Puppet 7 support
```
evgeni, line 4: can you keep the "puppet 7" comment please? it explains why the commit is pinned (see https://github.com/theforeman/foreman-installer/commit/51d21833ae04aa01d6f1fca7480deb74a41fc0e4)

evgeni:
can you remove `Puppetfile.lock` from `.gitignore` please?

evgeni:
And I think you need to re-run `bundle exec librarian-puppet install` after you added the git url for xinetd back, as right now the lockfile points at the released version.

Which make me think: seems CI still will use the non-locked version? While build will use the lock? Seems bad to me?

evgeni (approved):

PR #735, opened 2021-12-24
Use forgeapi-cdn.puppet.com as forge

ekohl:
These days puppet.com is the preferred domain. forgeapi-cdn has the benefit that it supports IPv6.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,9 @@
 source 'https://rubygems.org'
 
+# rdoc 6.4 pulls in psych 4 and Puppet is incompatible with that.
+# We don't want to list psych since that updates the bundled version
+gem 'rdoc', '< 6.4'
+
 gem 'kafo', '~> 6.4'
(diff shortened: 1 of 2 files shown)


evgeni:
forgeapi-cdn sounds like an internal name, which I'd prefer not to depend on?

Or is that documented somewhere as an alternative api hostname?

ekohl:
https://forgeapi.puppet.com/#section/Features states:
> The API is accessed over HTTPS via either the forgeapi.puppet.com (IPv4) or forgeapi-cdn.puppet.com (IPv4 or IPv6) domain. All data is returned in JSON format.

In https://forgeapi.puppet.com/#section/Hostname-Configuration all examples use forgeapi-cdn as well.

ekohl:
@evgeni can I take the :+1: as an approval? Though we currently have some other breakages (Psych 4), I'd like to know if I can merge this once green.

evgeni (approved):
Take this!

PR #736, opened 2022-01-07
Fix #34233 - Align migration file naming convention

ekohl:
Currently Foreman uses +%Y%m%d%H%M%S and Katello uses +%y%m%d%H%M%S. Historically Kafo has recommended the latter but the former uses complete timestamps, which makes them easier to recognize. It also uses underscores which is generally recommended for Ruby files.

Right now this hasn't been subject to a lot of testing. Only the included tests, no real system yet.

The reasoning for the probing of files is to make it work for scenarios that have not renamed their files yet. This may be optimized using regular expressions but I think this will be fast enough since it's all queried from the same directory so the dirent cache should suffice. Alternatively the hook could list all files in the directory (using the glob) and treat it as a regular array.

The reasoning for this is that eventually we can end up sharing migrations between foreman and katello by having the same migrations directory.


ekohl:
Marking back as draft since some migrations were added and this is out of date now. It should really add a check to ensure everything is in line.

ehelms:
I took this for a test drive and saw no significant issues. One aspect of it that I wanted to double check was intended as it generates a diff between new installations and upgrade is the "duplication" of migrations in the applied list -- the old name and the new name. For example:

```
 cat /etc/foreman-installer/scenarios.d/katello-migrations-applied | grep without
- 211108174119-disable-registration-without-templates.rb
- 20211108174119_disable_registration_without_templates.rb
```

PR #739, opened 2022-01-27
Fixes #34325 - enable pulp content cache by default

jlsherrill:


config/foreman-proxy-content-answers.yaml
```diff
@@ -14,6 +14,7 @@ certs:
 foreman_proxy_content:
   pulpcore_mirror: true
   puppet: false
+  pulpcore_cache_enabled: true
```
ekohl: This should not be needed once the PR in in puppet-foreman_proxy_content is merged. Right now it's still detecting the built in default of `false`. I'd prefer to keep this list minimal and only those where they do not match the defaults.

jlsherrill: ahh!  sounds good, i'll remove it

ekohl:
[test foreman-installer]

config/foreman-proxy-content-answers.yaml
ekohl: I think I found the reason it was needed. You should write the migration slightly more specific.

config/foreman-proxy-content.migrations/220127140752-enable-pulp-content-cache.rb
```diff
@@ -0,0 +1,3 @@
+if answers['foreman_proxy_content'].is_a?(Hash)
```
ekohl: If you write it more specific you won't need a change in the answers file. This would keep the default as built into the module. It only changes `false` to `true`.
```suggestion
if answers['foreman_proxy_content'].is_a?(Hash) && answers['foreman_proxy_content']['pulpcore_cache_enabled'] == false
```

ekohl (changes requested):

config/foreman-proxy-content.migrations/220127140752-enable-pulp-content-cache.rb
jlsherrill: so its safe to assume that it would be false and not nil?

ekohl: Good question so I had to stop and think for a moment. `nil` as a value means `undef` in Puppet. Not having the key means its default has yet to be determined. So it can also be:
```suggestion
if answers['foreman_proxy_content'].is_a?(Hash) && answers['foreman_proxy_content'].key?('pulpcore_cache_enabled')
```

Perhaps that's even cleaner.

ekohl (approved):
Thanks!

PR #744, opened 2022-02-11
Fixes #34440 - skip upgrade:run by run_rake_upgrade=false

evgeni:
this allows to override it in certain scenarios

### DIFF
--- a/hooks/post/30-upgrade.rb
+++ b/hooks/post/30-upgrade.rb
@@ -1,3 +1,8 @@
-if !app_value(:noop) && [0, 2].include?(@kafo.exit_code) && foreman_server?
+def run_rake_upgrade?
+  it = get_custom_config(:run_rake_upgrade)
+  it.nil? ? true : it
+end
+
+if !app_value(:noop) && [0, 2].include?(@kafo.exit_code) && foreman_server? && run_rake_upgrade?
   execute!('foreman-rake upgrade:run')
 end


evgeni:
Or should this be inverted? "skip_rake_upgrade", then only special scenarios need to set it?

evgeni:
So I didn't manage to set this via `/usr/share/foreman-installer/config/foreman.hiera/scenario/${scenario}/scenario-overrides.yaml`, but `/etc/foreman-installer/scenarios.d/${scenario}.yaml` works and is sufficient for my needs.

@ekohl thoughts?

ekohl (commented):
I'm not a fan of negative named values because you can get double negatives. However, `get_custom_config` doesn't have a `default` parameter so you'll get ugly code.

The alternative is to merge + release https://github.com/theforeman/kafo/pull/346 and then use `get_custom_config(:rake_upgrade, true)`. I want to do a release anyway to be able to use the `Sensitive` data type so it wouldn't be a huge deal.

I could be OK with merging this, but also wouldn't mind a more elegant solution. Thoughts?

evgeni:
I would prefer a solution that doesn't require a Kafo patch ;-)

```ruby
def should_run_rake_upgrade?
  it = get_custom_config(:rake_upgrade)
  it.nil? ? true : it
end
```

Would that do the same?

ekohl:
Yes it would. You can also write the last line as `it.nil? || it` but that's a trivial difference.

evgeni:
Updated (I went with my line, as I find it expresses the desire behind it a bit clearer)

ekohl (approved):

PR #747, opened 2022-02-16
Fixes #34379 - Update pulpcore to create a system group

ekohl:
The only bugfix here is to create the group pulp as a system group rather than a user group. This typically only means a gid between 500 and 1000 and doesn't affect existing systems.


ekohl:
@upadhyeammit for 3.1.2.

evgeni (approved):

PR #751, opened 2022-02-24
Fixes #33808 - Release 3.0.2

upadhyeammit:


Puppetfile
```diff
@@ -11,7 +11,7 @@ mod 'katello/candlepin',               '>= 12.0.0 < 12.1.0'
 mod 'theforeman/pulpcore',             '>= 5.0.0 < 5.1.0'
 mod 'katello/qpid',                    '>= 9.0.0 < 9.1.0'
 mod 'theforeman/foreman',              '>= 18.2.0 < 18.3.0'
-mod 'theforeman/foreman_proxy',        '>= 19.0.0 < 19.1.0'
+mod 'theforeman/foreman_proxy',        '>= 19.0.1 < 19.1.0'
```
ekohl: This is not needed since `>= 19.0.0` also matches `19.0.1`. Just the entry in the lockfile is sufficient.

Puppetfile.lock
```diff
@@ -120,7 +120,7 @@ DEPENDENCIES
   theforeman-dhcp (>= 7.0.0, < 7.1.0)
   theforeman-dns (>= 9.0.0, < 9.1.0)
   theforeman-foreman (>= 18.2.0, < 18.3.0)
-  theforeman-foreman_proxy (>= 19.0.0, < 19.1.0)
+  theforeman-foreman_proxy (>= 19.0.1, < 19.1.0)
```
ekohl: Like the comment in `Puppetfile`, this isn't needed.

ekohl (changes requested):
I have started with using `Fixes #...` on the bump commit so the PR processor automatically sets the Fixed in version.

ekohl (changes requested):
Looks like you also removed the crucial line in `Puppetfile.lock` that was needed, namely changing this to `19.0.1`:
https://github.com/theforeman/foreman-installer/blob/69b5f7a2e0f6f68230f7170fb0fc0bd6d4e70445/Puppetfile.lock#L75

ekohl (approved):
From the tests:
```
[Librarian] Installing theforeman-foreman_proxy/19.0.1 <https://forgeapi.puppetlabs.com/>
```

upadhyeammit:
@ekohl thanks for the review! Merging this.

PR #752, opened 2022-02-28
Drop EL8 pki-core modularity workaround

ekohl:
Now that there is a real katello module, this should no longer be needed.

Currently a draft since I'm not entirely sure about the implications.


ekohl:
Superseded by c74d373ced1295b1d7e0a99fd9c7ff0dfdf40d1f

PR #756, opened 2022-03-23
Fixes #32323 - Correct PostgreSQL service name on EL7

ekohl:
While we do install the -server-syspaths package (which allows using the non-SCL name), due to implementation details the systemd override files must be created on the original name. Otherwise they do not have any effect and only generate warnings.

See https://bugzilla.redhat.com/show_bug.cgi?id=1964394#c3 for a detailed analysis.

### DIFF
--- a/config/foreman.hiera/family/RedHat-7.yaml
+++ b/config/foreman.hiera/family/RedHat-7.yaml
@@ -5,7 +5,7 @@ postgresql::globals::version: '12'
 postgresql::globals::client_package_name: rh-postgresql12-postgresql-syspaths
 postgresql::globals::server_package_name: rh-postgresql12-postgresql-server-syspaths
 postgresql::globals::contrib_package_name: rh-postgresql12-postgresql-contrib-syspaths
-postgresql::globals::service_name: postgresql
+postgresql::globals::service_name: rh-postgresql12-postgresql
 postgresql::globals::datadir: /var/opt/rh/rh-postgresql12/lib/pgsql/data
 postgresql::globals::confdir: /var/opt/rh/rh-postgresql12/lib/pgsql/data
 postgresql::globals::bindir: /usr/bin
--- a/hooks/pre/29-el7_clean_systemd_pg.rb
+++ b/hooks/pre/29-el7_clean_systemd_pg.rb
@@ -1,0 +1,4 @@
+# On EL7 the SCL override should be used instead
+if !app_value(:noop) && el7?
+  FileUtils.rm_f('/etc/systemd/system/postgresql.service.d/postgresql.conf')
+end


ekohl:
Verified that this does make a difference. `systemctl cat postgresql` shows:
```
# /usr/lib/systemd/system/rh-postgresql12-postgresql.service
# It's not recommended to modify this file in-place, because it will be
# overwritten during package upgrades.  It is recommended to use systemd
# "dropin" feature;  i.e. create file with suffix .conf under
# /etc/systemd/system/UNITNAME.service.d directory overriding the
# unit's defaults.  Look at systemd.unit(5) manual page for more info.

[Unit]
Description=PostgreSQL database server
After=network.target

[Service]
Type=notify

User=postgres
Group=postgres

# Where to send early-startup messages from the server (before the logging
# options of postgresql.conf take effect)
# This is normally controlled by the global default set by systemd
# StandardOutput=syslog

# Disable OOM kill on the postmaster
OOMScoreAdjust=-1000
# ... but allow it still to be effective for child processes
# (note that these settings are ignored by Postgres releases before 9.5)
Environment=PG_OOM_ADJUST_FILE=/proc/self/oom_score_adj
Environment=PG_OOM_ADJUST_VALUE=0

Environment=PGDATA=/var/opt/rh/rh-postgresql12/lib/pgsql/data

ExecStartPre=/opt/rh/rh-postgresql12/root/usr/libexec/postgresql-check-db-dir %N
# Even though the $PGDATA variable is exported (postmaster would accept that)
# use the -D option here so PGDATA content is printed by /bin/ps and by
# 'systemctl status'.
ExecStart=/bin/sh -c 'source scl_source enable rh-postgresql12 ; exec postmaster -D ${PGDATA}'
ExecReload=/bin/kill -HUP $MAINPID
KillMode=mixed
KillSignal=SIGINT

# No artificial start/stop timeout (rhbz#1525477, pgrpms#2786).
TimeoutSec=0

[Install]
WantedBy=multi-user.target

# /etc/systemd/system/rh-postgresql12-postgresql.service.d/rh-postgresql12-postgresql.conf
[Service]
Environment=PGPORT=5432
Environment=PGDATA=/var/opt/rh/rh-postgresql12/lib/pgsql/data
```

ehelms (approved):
Seems reasonable to me. @evgeni also who was tracking this down

evgeni:
Thanks Ewoud!

This indeed fixes the situation on new installations. On upgrades however, I end up with two directories and two overrides, again triggering the `Warning: rh-postgresql12-postgresql.service changed on disk. Run 'systemctl daemon-reload' to reload units.`

```console
[root@sat ~]# ls -alh /etc/systemd/system/*postgresql.service.d/
/etc/systemd/system/postgresql.service.d/:
total 8.0K
drwxr-xr-x.  2 root root   29 Feb  2 08:30 .
drwxr-xr-x. 20 root root 4.0K Mar 24 07:13 ..
-rw-r--r--.  1 root root   97 Feb  2 08:30 postgresql.conf

/etc/systemd/system/rh-postgresql12-postgresql.service.d/:
total 8.0K
drwxr-xr-x.  2 root root   45 Mar 24 07:13 .
drwxr-xr-x. 20 root root 4.0K Mar 24 07:13 ..
-rw-r--r--.  1 root root   97 Mar 24 07:13 rh-postgresql12-postgresql.conf
```

Should we add a hook that purges `/etc/systemd/system/postgresql.service.d/postgresql.conf` on EL7? (The file is now unmanaged by the installer)

That said, I am happy to merge this PR as-is to improve the new-install experience, but I wouldn't call RM#32323 fixed by it.

ekohl:
> Should we add a hook that purges `/etc/systemd/system/postgresql.service.d/postgresql.conf` on EL7? (The file is now unmanaged by the installer)

Done. Mind checking?

evgeni:
> > Should we add a hook that purges `/etc/systemd/system/postgresql.service.d/postgresql.conf` on EL7? (The file is now unmanaged by the installer)
> 
> Done. Mind checking?

Reads correctly to me, will run once on my reproducer to be sure.

hooks/pre/29-el7_upgrade_postgresql.rb
```diff
@@ -0,0 +1,4 @@
+# On EL7 the SCL override should be used instead
+if !app_value(:noop) && el7
```
evgeni:
```suggestion
if !app_value(:noop) && el7?
```

hooks/pre/29-el7_upgrade_postgresql.rb
```diff
@@ -0,0 +1,4 @@
...
+  FileUtils.rm_f('/etc/systemd/system/postgresql.service.d/postgresql.conf')
```
evgeni: Should we also `FileUtils.rmdir('/etc/systemd/system/postgresql.service.d/')`? Probably with a `rescue`, as it might be non-empty, but then it's not our fault. (The rm of the conf file is sufficient for fixing the bug, this one is just cosmetics)

evgeni (changes requested):

hooks/pre/29-el7_upgrade_postgresql.rb
ekohl: Bleh, I should really get in the habit of writing tests for hooks.

hooks/pre/29-el7_upgrade_postgresql.rb
ekohl: I thought about that, but given EL7 is on its way out I was a bit lazy.

hooks/pre/29-el7_upgrade_postgresql.rb
ekohl: For now I just fixed it. It would require a fair bit of stubbing to get things to work.

evgeni (approved):

PR #758, opened 2022-04-06
Fixes #35677 - Add ForemanGoogle plugin

stejskalleos:


ekohl:
This will fail CI until a matching patch in https://github.com/theforeman/puppet-foreman is merged. Once the PR is available, please link it.

stejskalleos:
Puppet PR: https://github.com/theforeman/puppet-foreman/pull/1040

ekohl:
[test foreman-installer]

evgeni:
[test foreman-installer]

config/foreman-answers.yaml
```diff
@@ -23,7 +23,6 @@ foreman::cli::tasks: false
 foreman::cli::templates: false
 foreman::cli::webhooks: false
 foreman::compute::ec2: false
-foreman::compute::gce: false
```
ekohl, line 26: If you're dropping gce, you'll also need to reflect that in the migration. You could even use the existing answer to determine if the google plugin should be enabled.

ekohl (changes requested):

stejskalleos:
google rpm has been merged, rebased & updated

config/foreman.migrations/20220406150414_foreman_google.rb
```diff
@@ -0,0 +1,5 @@
+answers['foreman::plugin::google'] ||= false
+
+if answers.key?('foreman::compute::gce')
+  answers.delete('foreman::compute::gce')
```
ekohl: Wouldn't it make sense to assume that if previously someone used gce, they now want the google plugin?

stejskalleos: Yop, make sense, updated.

However `rspec ./spec/migration_spec.rb` tests are still failing on mismatch with answers, not sure why

config/katello.migrations/20220406141823_foreman_google.rb
```diff
@@ -0,0 +1 @@
+answers['foreman::plugin::google'] ||= false
```
ehelms: Need the same logic here as the foreman migration

config/foreman.migrations/20220406150414_foreman_google.rb
```diff
@@ -0,0 +1,2 @@
+answers['foreman::plugin::google'] = answers['foreman::compute::gce']
+answers.delete('foreman::compute::gce')
```
ekohl: `Hash#delete` returns the value so you can also write this as:
```suggestion
answers['foreman::plugin::google'] = answers.delete('foreman::compute::gce')
```

config/katello.migrations/20220406141823_foreman_google.rb
```diff
@@ -0,0 +1,2 @@
+answers['foreman::plugin::google'] = answers['foreman::compute::gce']
+answers.delete('foreman::compute::gce')
```
ekohl:
```suggestion
answers['foreman::plugin::google'] = answers.delete('foreman::compute::gce')
```

ekohl (changes requested):

stejskalleos:
@ekohl PR updated, however `rspec ./spec/migration_spec.rb` tests are still failing on mismatch with answers, not sure why

config/foreman.migrations/20220406150414_foreman_google.rb
```diff
@@ -0,0 +1 @@
+answers['foreman::plugin::google'] = answers.delete('foreman::compute::gce')
```
ekohl: Looking closer at it, the problem is that this migration runs on `config/foreman-answers.yaml`. There is no answer for `foreman::compute::gce` so it's nil. That means it is essentially:
```ruby
answers['foreman::plugin::google'] = nil
```

I think a safe version would be:
```suggestion
old_answer = answers.delete('foreman::compute::gce')
answers['foreman::plugin::google'] ||= old_answer || false
```

The latter part (`|| false`) because `false || nil` is `nil`. The former part `||=` to respect if the answer is already `true`. The delete is on its own line because it must happen and if the answer is already true, it won't be evaluated.

stejskalleos:
@ekohl migration files updated but it still fails :/

spec/fixtures/pulpcore-migration/katello-answers-after.yaml
```diff
@@ -21,6 +20,7 @@ foreman::plugin::dlm: false
 foreman::plugin::dhcp_browser: false
 foreman::plugin::expire_hosts: false
 foreman::plugin::git_templates: false
+foreman::plugin::google: false
```
ekohl, line 24: This is a modification that's missing in the other fixtures, causing the tests to fail.

stejskalleos:
@ekohl and it's :green_apple: !

ekohl (changes requested):
Can you include https://github.com/theforeman/foreman-installer/pull/815 here? Then also do some additional logic in the migration that if `foreman::cli` and `foreman::plugin::google` it also enabled `foreman::cli::google` so users have a smooth migration?

stejskalleos:
@ekohl https://github.com/theforeman/foreman-installer/pull/815 added

stejskalleos:
[test foreman-installer]

stejskalleos:
Tests are failing on `[Librarian] Conflict between theforeman-pulpcore (>= 7.0.1, < 8.0.0) <https://forgeapi.puppet.com/> (from katello-foreman_proxy_content) and theforeman-pulpcore/7.0.0 <https://github.com/theforeman/puppet-pulpcore#master>
Could not resolve the dependencies.`, not sure if it is related to my changes

config/foreman.migrations/20220406150414_foreman_google.rb
```diff
@@ -0,0 +1,4 @@
+old_answer = answers.delete('foreman::compute::gce')
+
+answers['foreman::plugin::google'] ||= old_answer || false
+answers['foreman::cli::google'] ||= old_answer || false
```
ekohl: Users can have disabled the CLI and if you then enable a CLI plugin it'll fail. I think this should take care of it.
```suggestion
answers['foreman::cli::google'] ||= answers['foreman::cli'] && old_answer || false
```

But what I'd prefer is to be more explicit in the logic, which is to enable it if the CLI and Google plugin are installed:
```suggestion
answers['foreman::cli::google'] ||= answers['foreman::cli'] && answers['foreman::plugin::google']
```

And copy the same to the Katello migration.

spec/fixtures/pulpcore-migration-dont-use-content-plugins-on-upgrades/katello-answers-before.yaml
```diff
@@ -4,7 +4,6 @@ foreman::cli::azure: false
 foreman::cli::kubevirt: false
 foreman::cli::virt_who_configure: false
 foreman::compute::ec2: false
-foreman::compute::gce: false
```
ekohl: This can remain. It was a snapshot of how the migration was back then.

spec/fixtures/pulpcore-migration-rpm-only/katello-answers-before.yaml
```diff
@@ -4,7 +4,6 @@ foreman::cli::azure: false
 foreman::cli::kubevirt: false
 foreman::cli::virt_who_configure: false
 foreman::compute::ec2: false
-foreman::compute::gce: false
```
ekohl: And this can remain too.

ekohl:
No, that pulpcore failure was because I should have done https://github.com/theforeman/puppet-pulpcore/commit/e124f89c117d221ecd350daf2d0627df4a573531 before merging the other PR.

stejskalleos:
@ekohl cool, now it's green again

ekohl (approved):
I think this is correct now. I'd say it's now blocked on packaging.

PR #759, opened 2022-04-11
Refs #34758: Match rename of remote_execution::ssh to remote_executio…

ehelms:
…n::script

See https://github.com/theforeman/puppet-foreman_proxy/pull/734

### DIFF
--- a/config/foreman-proxy-content.migrations/220411132130-rename-rex-ssh-rex-script.rb
+++ b/config/foreman-proxy-content.migrations/220411132130-rename-rex-ssh-rex-script.rb
@@ -1,0 +1,3 @@
+if answers.key?('foreman_proxy::plugin::remote_execution::ssh')
+  answers['foreman_proxy::plugin::remote_execution::script'] = answers.delete('foreman_proxy::plugin::remote_execution::ssh')
+end
--- a/config/katello.migrations/220411132123-rename-rex-ssh-rex-script.rb
+++ b/config/katello.migrations/220411132123-rename-rex-ssh-rex-script.rb
@@ -1,0 +1,3 @@
+if answers.key?('foreman_proxy::plugin::remote_execution::ssh')
+  answers['foreman_proxy::plugin::remote_execution::script'] = answers.delete('foreman_proxy::plugin::remote_execution::ssh')
+end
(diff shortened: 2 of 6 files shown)


config/foreman-proxy-content.migrations/220411132130-rename-rex-ssh-rex-script.rb
```diff
@@ -0,0 +1,4 @@
+if answers['foreman_proxy::plugin::remote_execution::ssh']
+  answers['foreman_proxy::plugin::remote_execution::script'] = answers['foreman_proxy::plugin::remote_execution::ssh']
+  answers.delete('foreman_proxy::plugin::remote_execution::ssh')
```
ekohl:
```suggestion
  answers['foreman_proxy::plugin::remote_execution::script'] = answers.delete('foreman_proxy::plugin::remote_execution::ssh')
```

config/katello.migrations/220411132123-rename-rex-ssh-rex-script.rb
```diff
@@ -0,0 +1,4 @@
+unless answers['foreman_proxy::plugin::remote_execution::ssh'].nil?
```
ekohl: Any reason you have a different migration here than the others?

ekohl (changes requested):
I'd prefer if you all have the same timestamps so it's easy to correlate. The easiest way to achieve this is to call `./bin/create-migration` with all scenarios it applies to.

adamruzicka:
Also inconsistent dashes vs underscores in migration file names?

ekohl:
Sadly that was already the case before and now we at least (try) to keep them consistent per scenario. In https://github.com/theforeman/foreman-installer/pull/736 I tried to align them but didn't get around to finishing it.

config/katello.migrations/220411132123-rename-rex-ssh-rex-script.rb
ehelms: It's like ordering ice cream but getting two flavors, its still ice cream. j/k total copy/paste failure.

ekohl: It's why I got to the workflow of:
```
cat <<EOF | ./bin/create-migration name $scenario
here is the content
EOF
```

ehelms:
[test foreman-installer]

ehelms:
@evgeni This is also ready now

evgeni (approved):

PR #760, opened 2022-04-22
Refs #34239: Require voxpupuli/puppet-mosquitto

ehelms:
This is needed to support REX pull transport via MQTT as puppet-mosquitto
is a soft dependency of puppet-foreman_proxy.


ehelms:
See https://github.com/theforeman/puppet-foreman_proxy/pull/737 for details

evgeni (approved):

evgeni:
On the other hand… any reason to use git here?

ehelms:
> On the other hand… any reason to use git here?

Just following the trend mostly? On our release branches we tend to change over from git to release. Let's follow that trend for now I guess.

evgeni:
I think that's only really true for *our* modules (e.g. PostgreSQL is a released version), as it is beneficial to test nightlies with the version that will be in the release and we do not release foreign modules during branching ;)

evgeni:
(I'd also be okay making it a hard dependency of the proxy module - while Ewoud is right, it is an edge case and not needed for most users, the module is very small and shouldn't harm anyone)

PR #764, opened 2022-05-06
Fixes #34875: Prevent certs regeneration on every installer run

ehelms:
Given answers are stored, if a user supplies --certs-regenerate then
every installer run thereafter will regenerate certificates. Ensure this
value is reset after installation run.

### DIFF
--- a/hooks/pre_exit/20-certs_regenerate.rb
+++ b/hooks/pre_exit/20-certs_regenerate.rb
@@ -1,0 +1,6 @@
+if module_enabled?('certs') && param('certs', 'regenerate').value == true
+  answers = kafo.config.answers
+  answers['certs']['regenerate'] = false
+
+  kafo.config.store(answers)
+end


evgeni:
forgeapi-cdn.puppet.com seems broken (I had similar results locally yesterday), shall we switch to forgeapi.puppet.com for the time being?

ehelms:
This is rather tricky, as it has to be done after the installation run but it requires altering and then re-saving the answers again. Here is a version that works, we should just double think about the potential issues, if any.

ehelms:
> forgeapi-cdn.puppet.com seems broken (I had similar results locally yesterday), shall we switch to forgeapi.puppet.com for the time being?

What's the difference between the two?

evgeni:
> > forgeapi-cdn.puppet.com seems broken (I had similar results locally yesterday), shall we switch to forgeapi.puppet.com for the time being?
> 
> What's the difference between the two?

cdn supports v6

https://github.com/theforeman/foreman-installer/pull/735

evgeni:
> cdn supports v6

turns out, forgeapi also now points at cloudfront and does v6: https://twitter.com/zhenech/status/1523984052033933312 & https://twitter.com/binford2k/status/1524027541270540289

evgeni:
How does that affect the certs-proxy scenario, where we always want to regen stuff?

https://github.com/theforeman/foreman-installer/blob/59e3e54005f92a2212d92a566bcb3bb7e13dc7e4/katello_certs/config/foreman-proxy-certs-answers.yaml#L3

(see https://github.com/theforeman/foreman-installer/pull/608 for details)

ehelms:
certs-proxy generate uses its own configuration with it's own hooks directory. So it will now encounter this post hook.

evgeni:
> certs-proxy generate uses its own configuration with it's own hooks directory. So it will now encounter this post hook.

Good, I wasn't too sure about all those configs.

ehelms:
https://github.com/theforeman/foreman-installer/blob/develop/katello_certs/config/foreman-proxy-certs.yaml

Since no hooks are defined here, it has only the default hooks directory which is rooted based on the `:installer_dir`. So it resolves as `:installer_dir/hooks` which would be https://github.com/theforeman/foreman-installer/tree/develop/katello_certs/hooks

evgeni (approved):

PR #766, opened 2022-05-09
update theforeman/foreman for the puma restart fix

evgeni:


evgeni:
this is to pull in the fix for https://projects.theforeman.org/issues/34824

upadhyeammit (approved):

PR #770, opened 2022-05-10
Fixes #34888: Display katello-certs-check output only if error

ehelms:


chris1984 (approved):
ACK, i keep seeing this and was wondering why, now it makes sense.

evgeni (approved):

PR #776, opened 2022-05-26
Drop devlog check that was only valid for Pulp 2

ehelms:
See https://projects.theforeman.org/issues/16778 for reason it was
added originally.


evgeni:
The old Katello installer PR is 404 🤣

evgeni (approved):

PR #781, opened 2022-06-02
Add Foreman Git templates plugin

dgoetz:
Depends on https://github.com/theforeman/puppet-foreman/pull/1059

### DIFF
--- a/config/foreman-answers.yaml
+++ b/config/foreman-answers.yaml
@@ -38,6 +38,7 @@ foreman::plugin::default_hostgroup: false
 foreman::plugin::dhcp_browser: false
 foreman::plugin::discovery: false
 foreman::plugin::expire_hosts: false
+foreman::plugin::git_templates: false
 foreman::plugin::hooks: false
 foreman::plugin::host_extra_validator: false
 foreman::plugin::host_reports: false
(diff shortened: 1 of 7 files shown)


upadhyeammit:
@dgoetz rebase please

hmm i think it will also require release of new puppet-foreman

ekohl:
[test foreman-installer]

The installer should use git builds for our modules. However, when this PR was submitted the dependency wasn't merged yet. I expect it to pass now.

config/foreman.migrations/20220602164000_foreman_git_templates.yaml
```diff
@@ -0,0 +1 @@
+answers['foreman::plugin::git_templates'] ||= false
```
ekohl: The migration should be a `.rb` file, not `.yaml`. Same with the other file.

ekohl (changes requested):

config/foreman.migrations/20220602164000_foreman_git_templates.yaml
dgoetz: Renamed it here and in all other pull requests (except the first one where I did it correct).

ekohl (approved):
Thanks!

ekohl:
Looks like you'll need to adjust the tests too.

dgoetz:
I adjusted the tests (as the migration now really works, tests had to be adjusted). This one is already green, others PRs tests are running and should become green. If some rebasing is needed just drop me a comment. Thanks for reviewing!

ekohl (approved):

PR #782, opened 2022-06-02
Add Foreman SCC manager plugin

dgoetz:
Depends on https://github.com/theforeman/puppet-foreman/pull/1057 and it is Katello only as the plugin only works on it, not pure Foreman.


upadhyeammit:
As per the test failure we need https://github.com/theforeman/puppet-foreman/pull/1057 ; @dgoetz can you please rebase?
Edit: rebase wont help we should need puppet-foreman

ekohl:
[test foreman-installer]
It just needs to be merged. CI ran before that.

ekohl (approved):

PR #785, opened 2022-06-21
Drop EL7 support

ehelms:


hooks/pre/32-install_selinux_packages.rb
```diff
@@ -2,17 +2,12 @@
...
-    packages << 'rh-redis5-redis' if local_redis?
-  end
-
   packages << 'foreman-selinux' if foreman_server?
   packages << 'katello-selinux' if katello_enabled?
   packages << 'candlepin-selinux' if katello_enabled?
   packages << 'pulpcore-selinux' if pulpcore_enabled?
 
-  if katello_enabled? && el8?
+  if katello_enabled?
```
ekohl: I think this entire hack is no longer needed now that we have proper DNF modules. That would allow us to get rid of the `el8?` definition too.

hooks/pre/32-install_selinux_packages.rb
```diff
@@ -2,20 +2,10 @@
...
-    packages << 'rh-postgresql12-postgresql-server' if local_postgresql?
-    packages << 'rh-redis5-redis' if local_redis?
-  end
-
   packages << 'foreman-selinux' if foreman_server?
   packages << 'katello-selinux' if katello_enabled?
   packages << 'candlepin-selinux' if katello_enabled?
   packages << 'pulpcore-selinux' if pulpcore_enabled?
 
-  if katello_enabled? && el8?
```
ekohl: Perhaps this could be a separate PR and already merged. No need to wait for it IMHO.

ekohl: https://github.com/theforeman/foreman-installer/pull/786

ehelms:
The nightly testing pipelines have dropped EL7 testing paths, therefore this should be mergable.

ekohl (approved):
I think we should also drop the EL7 migrations https://github.com/theforeman/foreman-installer/blob/6be1e967f934aee9d8a9350ab52246c93da60858/config/foreman.migrations/20200113133837_reset_bind-host-foreman-proxy.rb and https://github.com/theforeman/foreman-installer/blob/develop/config/katello.migrations/200113133837-reset-bind-host-foreman-proxy.rb.

ehelms:
> I think we should also drop the EL7 migrations https://github.com/theforeman/foreman-installer/blob/6be1e967f934aee9d8a9350ab52246c93da60858/config/foreman.migrations/20200113133837_reset_bind-host-foreman-proxy.rb and https://github.com/theforeman/foreman-installer/blob/develop/config/katello.migrations/200113133837-reset-bind-host-foreman-proxy.rb.

Dropped like they were hot

ekohl (approved):

PR #788, opened 2022-06-29
Fixes #35134: Introduce development tuning profile, update default

ehelms:
Introduces a development tuning profile that exists to have the lowest
possible values for a development or custom CI environment. Additionally
this moves the default tuning to match the performance tuning guide values for
default.

See [Performance Tuning Guide](https://docs.theforeman.org/nightly/Performance_Tuning_Guide/index-foreman-el.html#Puma_Workers_and_Threads_Recommendation_performance-tuning) for table of values for CPUs and memory.

Additional follow up work I intend to persue:

 * Use development profile in development environments and CI in places such as Forklift
 * Removal of `--disable-system-checks` where we expect that the development profile covers a reasonable bare minimum

### DIFF
--- a/hooks/boot/13-tuning.rb
+++ b/hooks/boot/13-tuning.rb
@@ -1,7 +1,8 @@
 if has_custom_fact?('tuning')
-  # TODO: automatically get the tuning sizes - standard doesn't exist and only
-  # loads base. The rest maps to config/foreman.hiera/tuning/sizes/$size.yaml
-  TUNING_SIZES = ['default'] + ['medium', 'large', 'extra-large', 'extra-extra-large']
+  # Tuning sizes are based off of config/foreman.hiera/tuning/common.yaml
+  # Additional configuration and overrides of base configuration are mapped to
+  # config/foreman.hiera/tuning/sizes/$size.yaml
+  TUNING_SIZES = ['default', 'medium', 'large', 'extra-large', 'extra-extra-large', 'development'].freeze
   TUNING_FACT = 'tuning'.freeze
 
   app_option(
--- a/hooks/pre_commit/13-tuning.rb
+++ b/hooks/pre_commit/13-tuning.rb
@@ -1,7 +1,8 @@
 if app_option?(:tuning)
   # A mapping of tuning profile to the required CPU cores and memory in GB
   TUNING_SIZES = {
-    'default' => { cpu_cores: 1, memory: 8 },
+    'development' => { cpu_cores: 1, memory: 6 },
+    'default' => { cpu_cores: 4, memory: 20 },
     'medium' => { cpu_cores: 8, memory: 32 },
     'large' => { cpu_cores: 16, memory: 64 },
     'extra-large' => { cpu_cores: 32, memory: 128 },


ekohl (commented):
The development scenario already has a tuning fact (so the option is exposed):
https://github.com/theforeman/forklift/blob/0c965ed9a48796781b1292df222f2f0edacbd6c5/roles/foreman_installer_devel_scenario/files/katello-devel.yaml#L19-L20

That should be changed to `development` for it to become effective, but users can set the option in their forklift config.

Another question I have is whether we want to expose this in all scenarios or just in development.

Thinking out loud: you could name it `minimal` or `minimal-lab` and show a warning (in a post hook?) that it's not intended for production loads. That way users who want to set up just a bare minimal to play in a lab know what to look for.

ehelms:
The use of `minimal` was debated as an option in some prior conversations. The worry was that the word alone did not convey that it is not for production use to users and we ended up with development. 

> Another question I have is whether we want to expose this in all scenarios or just in development. 

I think this was discussed too, and if I recall led down a path of defining the profiles and properties of the profiles per scenario and that it would be a bigger change than just level setting with a development profile.

@evgeni @upadhyeammit @ShimShtein can maybe help add any of the salient thoughts I am missing from prior discussions

evgeni:
One thing that we discussed was to use that profile in CI which uses production scenarios

ehelms:
Expanding a bit on `Removal of --disable-system-checks where we expect that the development profile covers a reasonable bare minimum`. There are two places where this flag is currently used: tuning profiles and vm overcommit check.

vm_overcommit is being raised here to drop https://github.com/theforeman/foreman-installer/pull/790

If we accept a minimal tuning profile and that it is reasonable that no one should need to bypass it then we enable the removal of the `--disable-system-checks` flag from the code base for good. That in affect helps solve the `--skip-checks-i-know-better` (which only affect `checks/`) vs. `--disable-system-checks` debate. And leaves only `--skip-checks-i-know-better` for skipping over `checks/` (of which there are currently 3 -- https://github.com/theforeman/foreman-installer/tree/develop/checks).

hooks/pre_commit/13-tuning.rb
```diff
@@ -1,7 +1,8 @@
 if app_option?(:tuning)
   # A mapping of tuning profile to the required CPU cores and memory in GB
   TUNING_SIZES = {
-    'default' => { cpu_cores: 1, memory: 8 },
+    'development' => { cpu_cores: 1, memory: 8 },
```
evgeni: https://github.com/theforeman/forklift/blob/0c965ed9a48796781b1292df222f2f0edacbd6c5/vagrant/lib/forklift/box_distributor.rb#L27 says 6 -- do you intend to adjust that?

ehelms: I did wonder if anybody would ask if this was the right set of values for development setups :D We should opt for some minimum that makes sense for Foreman and Katello general development environments. I think these days that means environments without qpid/puppetserver infrastructure so 6 GB could be do-able?

evgeni: 6 is what our nightly pipeliness pass with, so yeah, def do-able.
(even tho I would think increasing these could have benefits to the pipes…)

ehelms: > 6 is what our nightly pipeliness pass with, so yeah, def do-able. (even tho I would think increasing these could have benefits to the pipes…)

You have a suggestamendation?

evgeni: I do not, no.

hooks/boot/13-tuning.rb
```diff
@@ -1,7 +1,7 @@
 if has_custom_fact?('tuning')
   # TODO: automatically get the tuning sizes - standard doesn't exist and only
   # loads base. The rest maps to config/foreman.hiera/tuning/sizes/$size.yaml
-  TUNING_SIZES = ['default'] + ['medium', 'large', 'extra-large', 'extra-extra-large']
+  TUNING_SIZES = ['development', 'default', 'medium', 'large', 'extra-large', 'extra-extra-large'].freeze
```
ekohl: My reasoning for the previous code style was that you always list `default` as the first entry, then all the non-default sizes. I think this is something we should keep since it's user facing (when they run with `--help`).

ekohl (commented):
Other than a small comment :+1: for this. Looking forward to getting rid of the system checks.

hooks/boot/13-tuning.rb
ehelms: Ah ha, should I put development at the end then to put it furthest away?

ekohl: Sounds good to me.

hooks/boot/13-tuning.rb
```diff
@@ -1,7 +1,7 @@
 if has_custom_fact?('tuning')
   # TODO: automatically get the tuning sizes - standard doesn't exist and only
   # loads base. The rest maps to config/foreman.hiera/tuning/sizes/$size.yaml
-  TUNING_SIZES = ['default'] + ['medium', 'large', 'extra-large', 'extra-extra-large']
+  TUNING_SIZES = ['default'] + ['medium', 'large', 'extra-large', 'extra-extra-large', 'development'].freeze
```
ekohl: I wonder if it should freeze the whole:
```suggestion
  TUNING_SIZES = (['default'] + ['medium', 'large', 'extra-large', 'extra-extra-large', 'development']).freeze
```

Or it could be a single array.
```suggestion
  TUNING_SIZES = ['default', 'medium', 'large', 'extra-large', 'extra-extra-large', 'development'].freeze
```

hooks/boot/13-tuning.rb
```diff
@@ -1,7 +1,7 @@
 if has_custom_fact?('tuning')
   # TODO: automatically get the tuning sizes - standard doesn't exist and only
```
ekohl: Looks like this is out of date
```suggestion
  # TODO: automatically get the tuning sizes - default doesn't exist and only
```

hooks/boot/13-tuning.rb
```diff
@@ -1,7 +1,7 @@
...
   # loads base. The rest maps to config/foreman.hiera/tuning/sizes/$size.yaml
```
ekohl: There is no specific down tuning file? Perhaps that should be included in the comment.

ekohl (commented):
Apologies for not catching this earlier. It's why I normally don't add code comments: they go out of date anyway ;)

hooks/boot/13-tuning.rb
ehelms: `down tuning file` ? what does this mean?

ekohl: Perhaps I'm inventing things, but in my mind you can tune up (for larger systems) or tune down (for smaller systems). Not saying this is what we should use, but as an example:
https://github.com/theforeman/puppet-puppet/blob/9421bd4ec9fb819fb40c15789886f7bf55b12fab/spec/acceptance/hieradata/common.yaml#L5-L8

ehelms: You can change tuning profiles upward or downward if that is what you are asking? Not sure I see how it fits into this comment.

ekohl: I'll be more concrete: it states:
> The rest maps to config/foreman.hiera/tuning/sizes/$size.yaml

Which is this directory: https://github.com/theforeman/foreman-installer/tree/develop/config/foreman.hiera/tuning/sizes

There I don't see a development.yaml. The comment basically describes how `TUNING_SIZES` is populated. If we're now saying `development` exists, but doesn't have a matching file then the comment doesn't reflect reality. I'd like it to match though.

Does that help?

ehelms: Can I just get rid of this TODO?

ekohl: We could drop the TODO, but it could be helpful to understand where the values come from. You could also say: `TUNING_SIZES` should be equal to `TUNING_SIZES.keys` from `hooks/pre_commit/13-tuning.rb`. Then in the pre_commit hook you can point to `config/foreman.hiera/tuning/sizes/$size.yaml`.

ehelms: Let me know if that new wording helps

ekohl: I still think a hint that it should map to `hooks/pre_commit/13-tuning.rb` could be useful. And we will have a TODO if we want to add tuning sizes to a scenario different from katello: it'll have different requirements. Foreman by default doesn't require 20 GB RAM.

ekohl (approved):
I'd be OK with merging this. @evgeni mind having a last look?

evgeni (approved):

PR #790, opened 2022-06-30
Fixes #35133: Drop VM overcommit check

ehelms:
This check was originally introduced prior to having controls
set to establish memory usage among the java applications in the stack,
namely Candlepin at the time. As we now by default set Xmx this
check should no longer be needed.


hooks/pre_commit/13-tuning.rb
```diff
@@ -19,7 +19,9 @@
     new_tuning = current_tuning
   end
 
-  unless app_value(:disable_system_checks)
+  if app_value(:disable_system_checks)
+    logger.warn 'Skipping system checks.'
+  else
```
ekohl, line 24: This change is here so we don't lose the warning a user is disabling system checks?

ekohl (approved):

hooks/pre_commit/13-tuning.rb
ehelms: Correct, I was tempted to put it in it's own hook but given it's only used in this place I localized it (and well am trying to get rid of it all together).

PR #791, opened 2022-07-01
Fixes #35149: Drop --disable-system-checks installer option

ehelms:
Draft to start as it relies on:

 * https://github.com/theforeman/foreman-installer/pull/790
 * https://github.com/theforeman/foreman-installer/pull/788

I did not opt to keep the option around for a release with a warning because I felt like the option *implies* something is being disabled when used when in fact nothing is. There are a number of scripts that will need to be updated which is the downside. I was more worried about the false impression it would be sending and that forcing scripts to drop using this "workaround" is a good thing.

### DIFF
--- a/hooks/pre_commit/13-tuning.rb
+++ b/hooks/pre_commit/13-tuning.rb
@@ -20,34 +20,30 @@ if app_option?(:tuning)
     new_tuning = current_tuning
   end
 
-  if app_value(:disable_system_checks)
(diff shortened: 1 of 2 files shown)


ehelms:
@ekohl We need to decide if we want to include this change in Foreman 3.4

ekohl (approved):
Let's get this in. We should write an upgrade warning though.

PR #792, opened 2022-07-01
Drop validation hook that was for Pulp 2 nodes

ehelms:


ekohl (approved):

ekohl:
Just to check: there is no Redmine for this, right? That's intended?

ehelms:
> Just to check: there is no Redmine for this, right? That's intended?

Correct, I think this is obscure enough not to need one.

PR #793, opened 2022-07-01
Move check for existence of certs tarball to hooks directory

ehelms:


hooks/pre_validations/12-check_certs_tar.rb
```diff
@@ -0,0 +1,10 @@
+if module_enabled?('foreman_proxy_content') and !module_enabled?('katello')
```
ekohl: Ruby typically uses `&&` and Rubocop is unhappy about it
```suggestion
if module_enabled?('foreman_proxy_content') && !module_enabled?('katello')
```

ekohl: Thinking more about it, I think we should even change it up. If you pass the parameter, it will be used and the location should be checked. So you can directly change it to:
```suggestion
if module_enabled?('certs')
```

ekohl (approved):

PR #794, opened 2022-07-05
Fixes #35157 - Enable Pulp's content caching on Katello

ekohl:
Like 4155c66f59ea42fe10d709434d6ab5dcdb66b572 but for the Katello scenario.

### DIFF
--- a/config/katello.migrations/220705101935-enable-pulp-content-cache.rb
+++ b/config/katello.migrations/220705101935-enable-pulp-content-cache.rb
@@ -1,0 +1,3 @@
+if answers['foreman_proxy_content'].is_a?(Hash) && answers['foreman_proxy_content'].key?('pulpcore_cache_enabled')
+  answers['foreman_proxy_content']['pulpcore_cache_enabled'] = true
+end


evgeni (approved):

ehelms:
This commentary does not block this change, as it is the current state of our code. Following up from a discussion on IRC, this setting seems to now be required by Katello and therefore is a not optional. There was then some follow on discussion around whether this option should be removed and hard set to avoid user's inadvertently breaking their setups. @evgeni do you recall if that discussion went further or if we arrived at an outcome?

evgeni:
I think we (@ianballou and I) reached a "uuuh, not sure" point in that IRC discussion.

The only real *dependency* of Katello needing Pulp with caching enabled is the code in [ping.rb](https://github.com/Katello/katello/blob/edefdb877ca9566b53c1e1d8f51ff144567329d6/app/models/katello/ping.rb#L199-L201).

I think what broke us is the change in https://github.com/pulp/pulpcore/pull/1752 which made the Pulp status endpoint report `"redis_connection": {"connected": False}` when the cache is disabled instead if `"redis_connection": None` as it was before. The older result is correctly caught by `if json['redis_connection']` in `ping.rb`.

I'd argue this is a bug in Pulp, reporting a Redis connection which is not configured as "disconnected" vs "not configured".

(The change in this PR is still good, we do want cache enabled by default)

Edit: I pinged in `#pulp-dev` to look at this comment.

fao89:
> The only real _dependency_ of Katello needing Pulp with caching enabled is the code in [ping.rb](https://github.com/Katello/katello/blob/edefdb877ca9566b53c1e1d8f51ff144567329d6/app/models/katello/ping.rb#L199-L201).
> 
> I think what broke us is the change in [pulp/pulpcore#1752](https://github.com/pulp/pulpcore/pull/1752) which made the Pulp status endpoint report `"redis_connection": {"connected": False}` when the cache is disabled instead if `"redis_connection": None` as it was before. The older result is correctly caught by `if json['redis_connection']` in `ping.rb`.
> 


I agree with your assessment, but I wonder if changing it now would break other users

ianballou:
Yeah, I'd agree that, ideally, the `connected: False` would only pop up if it was specifically configured as such.  We might be able to drop this check entirely in Katello, but we'd need to test without Redis.  I'll add a redmine to check up on that.

With the current state of Katello's code though, I'd say let's get this merged to avoid more user issues.

PR #797, opened 2022-08-04
Refs #35149 - Remove katello hook dirs

ekohl:
In 10a30ba937f8e4c4cbbad889a05e3f5a8965eaf5 the top level directory katello was removed. This means the hook_dirs entry is no longer needed.  It also means the directory no longer needs to be copied during installation.


ehelms (approved):

ekohl:
I ran `rake install`, didn't check `rake spec`.

PR #798, opened 2022-08-10
Pin modules for Foreman 3.4

upadhyeammit:


Puppetfile
```diff
@@ -1,29 +1,19 @@
 forge 'https://forgeapi.puppet.com/'
 
-# Dependencies
-# For Puppet 7 support
 mod 'puppetlabs/xinetd',             :git => 'https://github.com/puppetlabs/puppetlabs-xinetd', :commit => 'e742608dccdf42236144acf9f05e483b47c576f1'
```
ekohl, line 3: Note to self: we should really drop xinetd now that we've dropped EL7. It's unused.

ekohl (changes requested):
Could you drop the empty migration?

Puppetfile
upadhyeammit: I was about to ask about this on when we are going to change it or maybe update the instruction in procedure as i always forget to pin it specifically :\

ekohl: https://github.com/theforeman/puppet-tftp/pull/136 would do that.

ekohl (approved):

PR #808, opened 2022-09-26
Fixes #35553 - Disable ansible-runner repo on upgrade

ekohl:
ansible-runner is now included in the Foreman repositories, so the repository should no longer be managed.


ehelms:
The management of the ansible-runner repository has been dropped entirely -- https://github.com/theforeman/puppet-foreman_proxy/commit/7f6601fcf0aa2b776429e7cabe0a0ad7cf038ddb

Thus I don't think this accurately captures the current scenario, perhaps this particular instance of the change needs to be only against a release branch?

ekohl:
Yes, I was intending this for the 3.4 branch where it does exist. It's fine to have a migration only in a stable branch I guess.

ehelms (approved):

PR #811, opened 2022-09-28
Fixes #35564 - Use Redis 6 on EL8

ehelms:
Requires: https://github.com/voxpupuli/puppet-redis/pull/450

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -6,6 +6,9 @@ mod 'puppetlabs/apache',             '>= 8.3.0'
 # Ensure Debian 11 support
 mod 'puppetlabs/postgresql',         '>= 7.4.0'
 
+# Dnfmodule support for Redis 6+ support
+mod 'puppet/redis',                  '>= 8.5.0'
+
 # Dependencies
 mod 'theforeman/dhcp',               :git => 'https://github.com/theforeman/puppet-dhcp'
 mod 'theforeman/dns',                :git => 'https://github.com/theforeman/puppet-dns'
--- a/config/foreman.hiera/family/RedHat-8.yaml
+++ b/config/foreman.hiera/family/RedHat-8.yaml
@@ -1,3 +1,6 @@
 ---
 postgresql::globals::manage_dnf_module: true
 postgresql::globals::version: "12"
+
+redis::dnf_module_stream: "6"
+redis::package_ensure: ">=6.0"


evgeni:
What will handle the migration of existing installations?

Those will have the `redis:5` module stream enabled, and you can't enable `redis:6` without a `dnf module reset` first (well, there is `dnf module switch-to`, but not in EL8). See https://docs.fedoraproject.org/en-US/modularity/using-modules/#_switching_module_streams for more explicit "how to switch" docs.

ekohl:
I think because of this line Puppet won't switch modules:
https://github.com/voxpupuli/puppet-redis/blob/cfa21daf10a11bbd40aafd409cffc39b2babe2f7/manifests/dnfmodule.pp#L15

evgeni:
> I think because of this line Puppet won't switch modules: https://github.com/voxpupuli/puppet-redis/blob/cfa21daf10a11bbd40aafd409cffc39b2babe2f7/manifests/dnfmodule.pp#L15

That's a nice pitfall! ;-)
`enable_only => true` means Puppet will execute `dnf module enable <module>` instead of `dnf module install <module>`.

```console
[root@6db2ac08f14e ~]# echo "package { 'redis': ensure => present }" | puppet apply -v
Notice: Compiled catalog for 6db2ac08f14e.redhat.com in environment production in 0.18 seconds
Info: Using environment 'production'
Info: Applying configuration version '1665125076'
Notice: /Stage[main]/Main/Package[redis]/ensure: created
Info: Creating state file /opt/puppetlabs/puppet/cache/state/state.yaml
Notice: Applied catalog in 1.83 seconds

[root@6db2ac08f14e ~]# dnf module list --enabled
Last metadata expiration check: 0:01:23 ago on Fri Oct  7 06:43:32 2022.
CentOS Stream 8 - AppStream
Name                                             Stream                                              Profiles                                             Summary                                                                      
redis                                            5 [d][e]                                            common [d]                                           Redis persistent key-value database                                          

Hint: [d]efault, [e]nabled, [x]disabled, [i]nstalled

[root@6db2ac08f14e ~]# echo "package { 'redis module': ensure => '6', name => 'redis', provider => 'dnfmodule', enable_only => true }" | puppet apply -v
Notice: Compiled catalog for 6db2ac08f14e.redhat.com in environment production in 0.18 seconds
Info: Using environment 'production'
Info: Applying configuration version '1665125223'
Notice: /Stage[main]/Main/Package[redis module]/ensure: ensure changed '5' to '6'
Notice: Applied catalog in 3.55 seconds

[root@6db2ac08f14e ~]# dnf module list --enabled
Last metadata expiration check: 0:03:39 ago on Fri Oct  7 06:43:32 2022.
CentOS Stream 8 - AppStream
Name                                             Stream                                            Profiles                                              Summary                                                                       
redis                                            6 [e]                                             common [d]                                            Redis persistent key-value database                                           

Hint: [d]efault, [e]nabled, [x]disabled, [i]nstalled

[root@6db2ac08f14e ~]# rpm -q redis
redis-5.0.3-5.module_el8.4.0+955+7126e393.x86_64
```

Interestingly, it seems to switch the module just fine, but obviously does not update the package if already installed.

evgeni:
@ehelms is this something we want to have for 3.5? (I would think so?)

ehelms:
> @ehelms is this something we want to have for 3.5? (I would think so?)

Yes! I need to finish upgrade testing.

config/foreman.hiera/family/RedHat-8.yaml
```diff
@@ -8,3 +8,6 @@ apache::mod::ssl::ssl_protocol:
 
 postgresql::globals::manage_dnf_module: true
 postgresql::globals::version: "12"
+
+redis::dnf_module_stream: "6"
+redis::package_ensure: "latest"
```
ehelms: This solves the update problem, but not sure if it's what we want? The dnf module does get reset properly, the package just needs to then get updated as its by default set to 'installed'. It feels right that the puppet layer handles all of this nicely for it. I suppose the other option is to do a hook that calls basically the same puppet code?

hooks/pre/40-update-redis.rb
```
apply_puppet_code("package { 'redis module': ensure => '6', name => 'redis', provider => 'dnfmodule', enable_only => true }")
apply_puppet_code("package { 'redis': ensure => 'latest' })
```

evgeni: This would mean that it *always* ensures latest, not only during the upgrade, which doesn't feel like what we want, indeed.

Puppetfile
```diff
@@ -22,6 +22,7 @@ mod 'theforeman/puppet',             :git => 'https://github.com/theforeman/pupp
 mod 'katello/foreman_proxy_content', :git => 'https://github.com/theforeman/puppet-foreman_proxy_content'
 mod 'katello/certs',                 :git => 'https://github.com/theforeman/puppet-certs'
 mod 'katello/katello',               :git => 'https://github.com/theforeman/puppet-katello'
+mod 'puppet/redis',                  :git => 'https://github.com/voxpupuli/puppet-redis', :ref => 'd5008dd41e92a66323c854b220c197425fdc7232'
```
ehelms: For now until we get a release of the module

config/foreman.hiera/family/RedHat-8.yaml
ehelms: The same would happen in the hook as well, unless we do a check of the Redis version and check for it. Or we have to add this enable/reset code to foreman-maintain for the module to happen during the upgrade process.

evgeni: In the past (el7 postgresql upgrade), we checked if the migration is needed and only executed it once, but that was also a more complicated migration.

Maintain feels like the wrong place - some people omit it and would end in a weird place (and not fully supported upstream anyways).

I'm torn, the current code is so nice and short… and in *most* cases the latest ensure would be a noop anyways.

ekohl: I've debated whether we would want to ensure it for every package (foreman, plugins, etc) in our installer. Still also a bit torn. I mean, ideally we'd have something like `redis::package_ensure: ">= 6.0"` but I don't think that's supported by Puppet.

ehelms: That worked on RPM:

```
redis::package_ensure: ">6"
```

According to Puppet -- https://puppet.com/docs/puppet/7/types/package.html
```
On supported providers, version ranges can also be ensured. For example, inequalities: <2.0.0, or intersections: >1.0.0 <2.0.0.
```

Question is if this works for Debian as well

ekohl: It's in the RH-8 file so this only applies to RPM so I'd say it's not a concern.

Puppetfile
ekohl: https://github.com/voxpupuli/puppet-redis/pull/454

ekohl:
```suggestion
mod 'puppet/redis',                  '>= 8.5.0'
```

config/foreman.hiera/family/RedHat-8.yaml
evgeni: Interestingly, https://puppet.com/docs/puppet/7/types/package.html#package-provider-features doesn't list the dnf provider as supporting ranges. But if it works…

ehelms: Heh, it does say that, but it works:

```
2022-10-28 18:27:44 [DEBUG ] [configure] Executing: '/bin/dnf list redis --showduplicates | sed -e '1,/Available Packages/ d' | awk '{print $2}''
2022-10-28 18:27:46 [DEBUG ] [configure] Package[redis](provider=dnf): Detected Arch argument in package! - Moving arch to end of version string
2022-10-28 18:27:46 [DEBUG ] [configure] Executing: '/bin/rpm -q redis --nosignature --nodigest --qf %{NAME} %|EPOCH?{%{EPOCH}}:{0}| %{VERSION} %{RELEASE} %{ARCH}\n'
2022-10-28 18:27:46 [DEBUG ] [configure] Package[redis](provider=dnf): Upgrading package redis from version 5.0.3-5.module_el8.4.0+955+7126e393 to 6.2.7-1.module_el8.7.0+1144+ab2666e8
2022-10-28 18:27:46 [DEBUG ] [configure] Facter: resolving fact with user_query: operatingsystemmajrelease
2022-10-28 18:27:46 [DEBUG ] [configure] Facter: Searching fact: operatingsystemmajrelease in file: operatingsystemmajrelease.rb
2022-10-28 18:27:46 [DEBUG ] [configure] Facter: Searching fact: operatingsystemmajrelease in core facts and external facts
2022-10-28 18:27:46 [DEBUG ] [configure] Facter: Loading all internal facts
2022-10-28 18:27:46 [DEBUG ] [configure] Facter: List of resolvable facts: [#<Facter::SearchedFact:0x00000000054042a0 @name="operatingsystemmajrelease", @fact_class=Facts::Rhel::Os::Release, @user_query="operatingsystemmajrelease", @type=:legacy, @file=nil>]
2022-10-28 18:27:46 [DEBUG ] [configure] Facter: Loading external facts
2022-10-28 18:27:46 [DEBUG ] [configure] Facter: fact "operatingsystemmajrelease" has resolved to: 8
2022-10-28 18:27:46 [DEBUG ] [configure] Executing: '/bin/dnf -d 0 -e 1 -y upgrade redis-6.2.7-1.module_el8.7.0+1144+ab2666e8'
2022-10-28 18:27:50 [DEBUG ] [configure] Executing: '/bin/rpm -q redis --nosignature --nodigest --qf %{NAME} %|EPOCH?{%{EPOCH}}:{0}| %{VERSION} %{RELEASE} %{ARCH}\n'
2022-10-28 18:27:50 [INFO  ] [configure] /Stage[main]/Redis::Install/Package[redis]/ensure: ensure changed '5.0.3-5.module_el8.4.0+955+7126e393' to '>=6.0'
2022-10-28 18:27:50 [DEBUG ] [configure] /Package[redis]: The container Class[Redis::Install] will propagate my refresh event
```

evgeni: Ah, interesting, it was intentionally implemented in https://tickets.puppetlabs.com/browse/PUP-10298, but seems never has been documented.

This is present in 6.15.0+, which we depend on anyways. I like!

evgeni (approved):

config/foreman.hiera/family/RedHat-8.yaml
evgeni: https://github.com/puppetlabs/puppet/pull/8943

Puppetfile
```diff
@@ -7,6 +7,7 @@ mod 'theforeman/dns',                :git => 'https://github.com/theforeman/pupp
 mod 'theforeman/git',                :git => 'https://github.com/theforeman/puppet-git'
 mod 'theforeman/puppetserver_foreman', :git => 'https://github.com/theforeman/puppet-puppetserver_foreman'
 mod 'theforeman/tftp',               :git => 'https://github.com/theforeman/puppet-tftp'
+mod 'puppet/redis',                  '>= 8.5.0'
```
evgeni: FTR, you might want to add this in the section with explanation headers (like why we pin postgresql and apache):

https://github.com/theforeman/foreman-installer/blob/ed50daee8015f93acd26c3afba8e5dcd5b7f4c00/Puppetfile#L3-L7

PR #814, opened 2022-10-13
Fixes #35629 - Default Apache to PROFILE=system ciphers

ekohl:
At least on EL8 it's possible to use PROFILE=system for SSLCipherSuite and SSLProxyCipherSuite. This allows admins to configure the cipher suite on a system level and it also means we don't have to keep our cipher suite up to date.

Today SSLProxyCipherSuite is not yet an option (https://github.com/puppetlabs/puppetlabs-apache/pull/2335 should add it), but Hiera will ignore unknown keys. When the option becomes available, it will be set.

I don't know if Debian/Ubuntu can do the same so for now I've set it only for RH-8.

### DIFF
--- a/config/foreman.hiera/family/RedHat-8.yaml
+++ b/config/foreman.hiera/family/RedHat-8.yaml
@@ -1,4 +1,7 @@
 ---
+apache::mod::ssl::ssl_ciper: 'PROFILE=system'
+# TODO: depends on https://github.com/puppetlabs/puppetlabs-apache/pull/2335
+apache::mod::ssl::ssl_proxy_ciper_suite: 'PROFILE=system'
 # EL8 doesn't have support for SSLv3 anymore and errors out on it. This
 # overrides security.yaml
 apache::mod::ssl::ssl_protocol:


ekohl:
> I don't know if Debian/Ubuntu can do the same so for now I've set it only for RH-8.

https://wiki.debian.org/CryptoPolicy suggests that it was mostly RH patches but update-crypto-policies is in Debian sid (https://packages.debian.org/unstable/crypto-policies) with RH Gitlab as upstream so I suspect over time it'll flow into Debian stable. For now we'll need something else.

I opened https://github.com/puppetlabs/puppetlabs-apache/pull/2336 to match OS defaults. Perhaps with that we could stop setting the ciphers altogether.

evgeni:
> https://wiki.debian.org/CryptoPolicy suggests that it was mostly RH patches but update-crypto-policies is in Debian sid (https://packages.debian.org/unstable/crypto-policies) with RH Gitlab as upstream so I suspect over time it'll flow into Debian stable. For now we'll need something else.

Looking at https://tracker.debian.org/pkg/crypto-policies, I wouldn't hold my breath on that one.

evgeni (approved):
I'm cool with setting this here explicitly, until `puppet-apache` keeps up with the defaults.

ekohl:
Turns out it's `PROFILE=SYSTEM`, not PROFILE=system`. See https://github.com/puppetlabs/puppetlabs-apache/pull/2336#issuecomment-1292232421 as well.

PR #817, opened 2022-10-26
Fixes #35631 - Enable HTTP/2 by default

ekohl:
Apache doesn't load HTTP/2 by default since it's incompatible with the prefork MPM. We use the event MPM where it should work. Enabling HTTP/2 allows clients to retrieve resources in parallel which means pages load faster.

Currently this is blocked on loading mod_http2, which will be done automatically once https://github.com/puppetlabs/puppetlabs-apache/pull/2337 is released. Apache ignores unknown protocols, so it can probably be merged, but then we don't get the benefit. I'd prefer to ensure we have a sufficiently recents puppetlabs/apache module.

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -1,7 +1,9 @@
 forge 'https://forgeapi.puppet.com/'
 
+# Ensure Debian 11 support
+mod 'puppetlabs/postgresql',         '>= 7.4.0'
+
 # Dependencies
-mod 'puppetlabs/postgresql',         '>= 7.0.0'
 mod 'theforeman/dhcp',               :git => 'https://github.com/theforeman/puppet-dhcp'
 mod 'theforeman/dns',                :git => 'https://github.com/theforeman/puppet-dns'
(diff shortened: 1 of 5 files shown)


ekohl:
I threw in some related changes: it now uses the unreleased puppetlabs/apache module.

Puppetfile
```diff
@@ -1,7 +1,14 @@
 forge 'https://forgeapi.puppet.com/'
 
+# Unreleased module for:
+# http2 loading: 6fa784a66061551e844203f17ad9367f7ad41d4b
+# ssl support: b3b9a82f50442c9d197ee8c30cb3df286d7eb099 & 4bd6143eba4bf58eabdcd0044bc0e3eadb1400b9
+mod 'puppetlabs/apache',             :git => 'https://github.com/puppetlabs/puppetlabs-apache', :ref => '47f0d7259cb7a368a685068a45d8422cc15b456b'
```
ekohl, line 6: It'd need https://github.com/puppetlabs/puppetlabs-apache/pull/2341 for a release.

ehelms (approved):

PR #818, opened 2022-10-28
Refs #35629 - Switch back to puppetlabs/apache releases

ekohl:
8.3.0 was released with the features we need.


ehelms (approved):
Well, that was fast

PR #820, opened 2022-10-31
drop chef plugin

evgeni:


config/katello.migrations/221031100739-drop-chef-proxy-plugin.rb
```diff
@@ -0,0 +1 @@
+answers.delete('foreman_proxy::plugin::chef')
```
evgeni: there never was a migration to add this to foreman, or to add the non-proxy one, so I only added the "unmigration" for that?

ekohl: Should this removal also be applied to the other 2 scenarios? And I think this should also remove `foreman::plugin::chef`.

ehelms (approved):

ekohl (approved):

PR #828, opened 2023-01-24
Fixes #35985 - Use Katello integration in theforeman/puppet

ekohl:
Related PRs:
* https://github.com/theforeman/puppet-puppet/pull/862 
* https://github.com/theforeman/puppet-foreman_proxy_content/pull/440
* https://github.com/theforeman/forklift/pull/1649

### DIFF
--- a/config/foreman.hiera/scenario/katello.yaml
+++ b/config/foreman.hiera/scenario/katello.yaml
@@ -1,0 +1,2 @@
+---
+puppet::server::foreman::katello: true
--- a/config/foreman-hiera.yaml
+++ b/config/foreman-hiera.yaml
@@ -15,6 +15,7 @@ hierarchy:
   - name: "Built in"
     paths:
       - "scenario/%{facts.kafo.scenario.id}/family/%{facts.os.family}-%{facts.os.release.major}.yaml"
+      - "scenario/%{facts.kafo.scenario.id}.yaml"
       - "family/%{facts.os.family}-%{facts.os.release.major}.yaml"
       - "family/%{facts.os.family}.yaml"
       - "security.yaml"
--- a/config/foreman.hiera/scenario/foreman-proxy-content.yaml
+++ b/config/foreman.hiera/scenario/foreman-proxy-content.yaml
@@ -1,0 +1,2 @@
+---
+puppet::server::foreman::katello: true


ekohl:
Looks like I missed at least the ssl key:
```console
# dnf install https://yum.puppet.com/puppet7-release-el-8.noarch.rpm https://yum.theforeman.org/releases/nightly/el8/x86_64/foreman-release.rpm https://yum.theforeman.org/katello/nightly/katello/el8/x86_64/katello-repos-latest.rpm
# dnf module enable katello:el8 pulpcore:el8
# dnf install https://ekohl.nl/foreman-installer-3.6.0-0.1.develop.el8.noarch.rpm https://ekohl.nl/foreman-installer-katello-3.6.0-0.1.develop.el8.noarch.rpm
# foreman-installer --scenario katello --enable-puppet --puppet-server true --foreman-proxy-puppet true --foreman-proxy-puppetca true -s --tuning development --enable-foreman-plugin-puppet
# grep ssl /etc/puppetlabs/puppet/foreman.yaml 
:ssl_ca: "/etc/pki/katello/puppet/puppet_client_ca.crt"
:ssl_cert: "/etc/pki/katello/puppet/puppet_client.crt"
:ssl_key: "/etc/puppetlabs/puppet/ssl/private_keys/centos8-stream.wisse.example.com.pem"
```

But it at least demonstrates the effect. Fixing the ssl key and adding some tests for it now.

ekohl:
> Fixing the ssl key

Did that

> adding some tests for it now.

Right now it's not trivial since I need to look at how I'd simulate Hiera in rspec-puppet.

ehelms:
This will need a migration alongside of it to I imagine to capture the previous choice of the user had for `puppet-foreman_proxy_content::puppet`.

ekohl:
> This will need a migration alongside of it to I imagine to capture the previous choice of the user had for `puppet-foreman_proxy_content::puppet`.

Not really. We drop the parameter (while all the others remain). Kafo automatically drops non-existent parameters from the answers file.

There is the implication that the user can no longer pass in the certificate parameters. Or rather, they're being ignored. That may be misleading and I can update the help for it.

config/foreman.hiera/scenario/katello.yaml
```diff
@@ -0,0 +1,2 @@
+---
+puppet::server::foreman::katello: true
```
ehelms, line 2: So I fully understand, this will only trigger this class if the main puppet class is enabled?

ehelms:
We will need a change to the bats tests to handle the new and old way of enabling puppet: https://github.com/theforeman/forklift/blob/master/bats/fb-test-puppet.bats#L22

config/foreman.hiera/scenario/katello.yaml
ekohl: Pretty much, but I'm going to be pedantic.

Technically only if the `puppet::server::foreman` class ends up being evaluated without a more specific answer. Hiera is more like a datastore. If Puppet doesn't ask about the value, it doesn't affect anything.

For example, this will trigger a lookup:
```puppet
include puppet::server::foreman
```

But this won't:
```puppet
class { 'puppet::server::foreman':
  katello => true, # or false - doesn't matter
}
```

For us that typically means you need to: `include puppet`, make sure both `puppet::server` and `puppet::server_foreman` are true.

ekohl:
https://github.com/theforeman/forklift/pull/1649 should be that.

ekohl:
@ehelms could you take a look again?
[test foreman-installer]

ehelms (approved):

PR #830, opened 2023-02-01
Fixes #36028 - ensure compressed assets are returned if available

evgeni:


Puppetfile
```diff
@@ -17,7 +17,7 @@ mod 'theforeman/tftp',                 '>= 8.0.0 < 8.1.0'
 mod 'katello/candlepin',               '>= 13.1.0 < 13.2.0'
 mod 'theforeman/pulpcore',             '>= 7.1.0 < 7.2.0'
 mod 'katello/qpid',                    '>= 9.1.0 < 9.2.0'
-mod 'theforeman/foreman',              '>= 22.1.0 < 22.2.0'
+mod 'theforeman/foreman',              '>= 22.1.2 < 22.2.0'
```
ekohl, line 20: In the past we haven't bumped the `Puppetfile` and relied on the lockfile, though I guess it doesn't hurt to be explicit.

ekohl (approved):

Puppetfile
evgeni: Depends on the definition of "we", I certainly did in the past (https://github.com/theforeman/foreman-installer/commit/ab725b40a455b4517993243c539550cef9ca0316) and didn't get caught :see_no_evil:

PR #835, opened 2023-03-01
drop memcache plugin support

evgeni:
it's broken since Foreman 3.4 and nobody is fixing it


ehelms (approved):

ekohl (approved):
I wouldn't mind a Redmine issue tracking this. Also cherry pick this to 3.6.

PR #837, opened 2023-03-07
Fixes #36170 - drop memcache plugin support

evgeni:
it's broken since Foreman 3.4 and nobody is fixing it

(cherry picked from commit bb767c98d979017bdbc5536638c85e25abafe0ad)


evgeni:
technically we don't need to update `puppet-foreman` to include https://github.com/theforeman/puppet-foreman/commit/787033f3187c42638096c28b14a6b7fa61b91a53, so I didn't for now.

ekohl (approved):
Indeed. No harm in it being there.

PR #848, opened 2023-05-04
correct upstream tag template in packit config

evgeni:


.packit.yaml
```diff
@@ -13,7 +13,7 @@ upstream_package_name: foreman-installer
 # downstream (Fedora) RPM package name
 downstream_package_name: foreman-installer
 
-upstream_tag_template: v{version}
+upstream_tag_template: "{version}"
```
evgeni, line 16: technically, this is the default, and the line could be completely omitted, I am just never sure whether I want to rely on that or not.

ekohl (approved):

PR #862, opened 2023-07-04
Fixes #36561 - Run katello-certs-check on proxy cert generation

ekohl:
This is similar to how it runs on the regular installer. The advantage of running this in foreman-proxy-certs-generate is that the user gets the feedback earlier. So rather than running:

* foreman-proxy-certs-generate
* copy tar file to proxy
* run installer

And then finding out there was something wrong, the user gets the feedback at the first step.

There could be a minor mismatch in version for n-1 version support, but it's safe to assume the server will only be stricter but still compatible. The check can be skipped if needed.

### DIFF
--- a/bin/foreman-proxy-certs-generate
+++ b/bin/foreman-proxy-certs-generate
@@ -1,5 +1,7 @@
 #!/usr/bin/env ruby
 require 'rubygems'
+require 'shellwords'
+require 'open3'
 require 'kafo'
 
 CONFIG_DIR = './katello_certs/config/'.freeze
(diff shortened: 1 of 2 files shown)


bin/foreman-proxy-certs-generate
```diff
@@ -28,5 +30,32 @@ Kafo::KafoConfigure.hooking.register_pre(:init) do
...
+  )
+end
+
+Kafo::KafoConfigure.hooking.register_pre(:pre_commit) do
+  ca_file   = param('certs', 'server_ca_cert').value
+  cert_file = param('certs', 'server_cert').value
+  key_file  = param('certs', 'server_key').value
+
+  unless app_value(:certs_skip_check) || [cert_file, ca_file, key_file].all? { |v| v.to_s.empty? }
+    # TODO: branding for foreman-proxy
```
ekohl: @evgeni should we modify katello-certs-check to always accept `foreman-proxy` for `-t`, even if it's branded?

ekohl:
I did a bit of testing to run the command inside rspec but it fails for me because I'm trying to run things as non-root and Puppet is not happy with me on Fedora:

```
Debug: Loading facts from /home/ekohl/dev/foreman-: 0 00:00:00\nDebug: /Stage[main]/Certs::Install/before: before : 0 0.0/s 00:00:01\nDebug: Prefetching dnf resources for package      : 0 0.0/s 00:00:01\nCould not prefetch package provider 'dnf': The yum provider can only be used as root\nDone                                              : 1 0.0/s 00:00:01\nDone                                              : 1 0.0/s 00:00:01\n
```

bin/foreman-proxy-certs-generate
evgeni: Doesn't it do that? (I have no memories about how I implemented branding there)

ekohl:
I just realized I ended up doing this in https://github.com/theforeman/foreman-installer/pull/895 too.

bin/foreman-proxy-certs-generate
ekohl: As I read this part I think it only accepts the branded value:
https://github.com/theforeman/foreman-installer/blob/8b7b58dfd4b5d50a391b6c21d9b8db777896fc7f/bin/katello-certs-check#L21

PR #868, opened 2023-07-06
Make specs RuboCop clean

ekohl:
This is a preparation to update RuboCop to verison 1.54.1.

### DIFF
--- a/spec/hook_context_extension_spec.rb
+++ b/spec/hook_context_extension_spec.rb
@@ -2,31 +2,32 @@ require 'spec_helper'
 require 'kafo/hook_context'
 require_relative '../hooks/boot/01-kafo-hook-extensions'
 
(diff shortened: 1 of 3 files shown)


spec/hook_context_extension_spec.rb
```diff
@@ -120,23 +121,23 @@
       end
 
       after do
-        expect(Kafo::PuppetCommand).to have_received(:search_puppet_path).twice
+        expect(Kafo::PuppetCommand).to have_received(:search_puppet_path).twice # rubocop:disable RSpec/ExpectInHook
```
ekohl: This won't be needed until we update RuboCop
```suggestion
        expect(Kafo::PuppetCommand).to have_received(:search_puppet_path).twice
```

PR #870, opened 2023-07-11
Restructure installation to avoid duplicate files

ekohl:
Split off from https://github.com/theforeman/foreman-installer/pull/822

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -50,10 +50,11 @@ directory BUILDDIR
 directory PKGDIR
 directory "#{BUILDDIR}/parser_cache"
 file "#{BUILDDIR}/parser_cache" => BUILDDIR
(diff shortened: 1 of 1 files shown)


evgeni:
My Rakefile knowledge is a bit limited, but this reads like the built/packaged result now differs a bit (RPM build failed with `File not found: /builddir/build/BUILDROOT/foreman-installer-3.8.0-0.20230711084104766030.pr870.720.gcff75b3.el8.x86_64/etc/foreman-installer/scenarios.d/foreman.migrations`) and I wonder if this will make existing installations forget their "applied migrations" list?

ekohl:
I should have mentioned this needs packaging changes. I've split off https://github.com/theforeman/foreman-packaging/pull/9564 from https://github.com/theforeman/foreman-packaging/pull/8814. Going to do a scratch build of that now.

PR #872, opened 2023-07-21
Ensure maven:3.8 is enabled on EL8

ehelms:


ekohl (changes requested):
I feel this is good to have a Redmine issue for, just so it's clear this changes in a patch release. In hindsight, perhaps something we should have already done for the puppet-candlepin patch.

ehelms:
Moving this to draft because we have identified that the packages between maven do not update properly -- see https://github.com/theforeman/puppet-candlepin/pull/236

PR #876, opened 2023-08-11
Clean up unused keys related to katello-agent

wbclark:
This will be accompanied by a check in foreman-maintain that halts the upgrade if katello-agent is enabled and warns the user to have an alternative deployed before upgrading.

It will be followed by complete removal of katello-agent and qpid code from the installer modules, after a release to ensure that upgrading users have everything cleaned up.

### DIFF
--- a/config/foreman.hiera/common.yaml
+++ b/config/foreman.hiera/common.yaml
@@ -18,5 +18,3 @@ dhcp::config_comment: |
 foreman::config::apache::proxy_params:
   retry: '0'
   timeout: '900'
-
-katello::globals::enable_katello_agent: "%{alias('foreman_proxy_content::enable_katello_agent')}"
--- a/config/foreman.hiera/tuning/common.yaml
+++ b/config/foreman.hiera/tuning/common.yaml
@@ -8,9 +8,6 @@ lookup_options:
   postgresql::server::config_entries:
     merge: hash
 
-qpid::open_file_limit: 65536
-qpid::router::open_file_limit: 150100
-
 postgresql::server::config_entries:
   checkpoint_completion_target: 0.9
   max_connections: 500


evgeni:
> This will be accompanied by a check in foreman-maintain that halts the upgrade if katello-agent is enabled and warns the user to have an alternative deployed before upgrading.

What if the user doesn't use f-m for upgrades, like all upstream users? ;)

config/foreman-proxy-content.migrations/230811192103-disable-katello-agent.rb
```diff
@@ -0,0 +1,3 @@
+if answers['foreman_proxy_content'].is_a?(Hash) && answers['foreman_proxy_content'].key?('enable_katello_agent')
+  answers['foreman_proxy_content'].delete('enable_katello_agent')
+end
```
ehelms: While this would remove it when the migration is ran the first time, if the parameter still exists coming from the puppet module, I am fairly confident the parameter would just come back. And users will see it as an option. The hiera config can work to ensure that no matter what the user sets it is ignored. 

I have been thinking we should drop the parameter (from puppet-katello and puppet-foreman_proxy_content) and hard-code the value inside the puppet module to false. In that way, we can ensure that the puppet module does all the right clean up but that the user does not see the installer option and cannot try to enable it.

ehelms:
> > This will be accompanied by a check in foreman-maintain that halts the upgrade if katello-agent is enabled and warns the user to have an alternative deployed before upgrading.
> 
> What if the user doesn't use f-m for upgrades, like all upstream users? ;)

Best we can do is document it in that case? And hope users read the docs / release notes before upgrading since there is no central way to control this for those users.


...one of these days we might finish that unifying of upgrade paths.

config/foreman-proxy-content.migrations/230811192103-disable-katello-agent.rb
wbclark: It will come back, but with the default value of `false`. I could remove the parameter from the modules however and hard-code the value to be false, if it's necessary to prevent users from attempting to re-enable it. (But Katello still won't let them use it if they tried)

wbclark:
> Best we can do is document it in that case? And hope users read the docs / release notes before upgrading since there is no central way to control this for those users.

The interface has had banners and notifications announcing the deprecation, too.

evgeni:
Yeah, I just think we should be really bold in the release notes/upgrade warnings about the fact it's now really gone.

ianballou:
> Yeah, I just think we should be really bold in the release notes/upgrade warnings about the fact it's now really gone.

We can certainly do that for Katello 4.10! It'll be a headline feature. Plus, it can be mentioned in the community post numerous times.

CC release captains @chris1984 and @damoore044

jeremylenz:
> What if the user doesn't use f-m for upgrades, like all upstream users? ;)

Seconding what was said above about documentation, banners, etc. But also wanted to throw in that I get the sense katello-agent use upstream is quite uncommon nowadays and most users have already moved to REX, so this is less of a concern for me upstream.

config/foreman-proxy-content.migrations/230811192103-disable-katello-agent.rb
jeremylenz: Flipping this: I don't see a reason to allow users to re-enable it, since Katello's code won't let you use it anyway.

ehelms: > Flipping this: I don't see a reason to allow users to re-enable it, since Katello's code won't let you use it anyway.

That was my thinking, if we can avoid even presenting the potential to users the better off we are. And I think we can do that in this scenario pretty easily.

wbclark: To prevent users from trying to re-enable it, this is roughly what the change looks like in puppet-katello: https://github.com/theforeman/puppet-katello/pull/471

(puppet-FPC in-transit)

config/foreman.hiera/common.yaml
```diff
@@ -19,4 +19,4 @@ foreman::config::apache::proxy_params:
   retry: '0'
   timeout: '900'
 
-katello::globals::enable_katello_agent: "%{alias('foreman_proxy_content::enable_katello_agent')}"
+katello::globals::enable_katello_agent: false
```
ekohl: In https://github.com/theforeman/puppet-katello/pull/471 this parameter is removed, so it can be dropped.

ekohl (changes requested):
If we merge https://github.com/theforeman/puppet-katello/pull/471 & https://github.com/theforeman/puppet-foreman_proxy_content/pull/455 the paraemters are gone and Kafo will remove them from the answer files automatically. That means you can drop the migrations.

wbclark:
Well, this ended up getting a whole lot simpler without the migrations.

Merge with https://github.com/theforeman/puppet-katello/pull/471 and https://github.com/theforeman/puppet-foreman_proxy_content/pull/455

config/foreman.hiera/common.yaml
wbclark: Thanks!

ekohl (approved):

PR #877, opened 2023-08-21
Fixes #36311 - expose apache::mod::status in the installer

evgeni:


config/foreman-proxy-content.yaml
```diff
@@ -20,6 +20,7 @@
   - foreman_proxy::plugin::pulp
   - foreman_proxy_content
   - puppet
+  - apache::mod::status
```
evgeni: is this entry even needed? :)

ekohl: The compile order in Puppet can matter. For this particular one, probably not since it just includes apache if needed:
https://github.com/puppetlabs/puppetlabs-apache/blob/c97699bf72efe4d86b6aec7ba2b665ee41b36803/manifests/mod/status.pp#L37

evgeni: compilation seems to be happy without, yay

ekohl (approved):

PR #886, opened 2023-10-04
Refs #36760 - Reset candlepin key- and truststore

ekohl:
This takes a 2 step approach where the cached password is removed during migration (which ends up running during the RPM installation). Since we don't want to break the system, the actual stores are only removed when the installer runs. This will recreate them and update the matching password files. Allowing a reset, even if theforeman/certs can't deal with it.

This is one option that doesn't rely on https://github.com/theforeman/puppet-certs/commit/6fea0bbb4143ca439cff01bf9f0e54cf88140d10, but instead implements a hook.


ekohl:
Closing in favor of https://github.com/theforeman/foreman-installer/pull/887

PR #887, opened 2023-10-04
Refs #36760 - Reset candlepin key- and truststore

ekohl:
This takes a 2 step approach where the cached password is removed during migration (which ends up running during the RPM installation). The installer then handles replacing the stores when it really runs.

This version is an alternative to https://github.com/theforeman/foreman-installer/pull/886, but relies on https://github.com/theforeman/puppet-certs/commit/6fea0bbb4143ca439cff01bf9f0e54cf88140d10 to do the heavy lifting. When cherry picking, it must include a Puppetfile update that includes this support.


ehelms:
This would be my preferred approach to avoid the migration

PR #888, opened 2023-10-05
Refs #36801: Set Redis as the default cache on upgrade

ehelms:


ehelms:
Puppet change -- https://github.com/theforeman/puppet-foreman/pull/1134

config/foreman.migrations/20231005004305_redis_default_cache.rb
```diff
@@ -0,0 +1,3 @@
+if answers['foreman'].is_a?(Hash) && answers['foreman']['rails_cache_store'] == { 'type' => 'file' }
```
ekohl: Would it be better to make this more robust and only check if type equals file?
```suggestion
if answers['foreman'].is_a?(Hash) && answers['foreman']['rails_cache_store'].is_a?(Hash) && answers['foreman']['rails_cache_store']['type'] == 'file'
```

config/foreman.migrations/20231005004305_redis_default_cache.rb
```diff
@@ -0,0 +1,3 @@
+if answers['foreman'].is_a?(Hash) && answers.dig('foreman', 'rails_cache_store', 'type') == 'file'
```
ekohl, line 1: I thought about dig, but didn't suggest it because I wasn't 100% sure if `rails_cache_store` was always a hash or `nil`. Now that I think more about it, I think that's all good.

ekohl (approved):
I think this is good, but tests are red for an unrelated reason. I want to do the big puppetlabs/stdlib 9.x upgrade in our modules, but that will be bumpy.

ehelms:
> I think this is good, but tests are red for an unrelated reason. I want to do the big puppetlabs/stdlib 9.x upgrade in our modules, but that will be bumpy.

Is this https://github.com/theforeman/foreman-installer-modulesync/pull/152 ?

ekohl:
It's related. The stupid thing is that puppetlabs/stdlib 8.x deprecated some functions without replacement and puppetlabs/stdlib 9.0 removed those functions. So the upgrade to 9.x requires to bump the lower bound too. It's really painful. If they only made a compatibility release in 8.x it would have been much smoother.

ekohl:
If you rebase, it should be green because https://github.com/theforeman/foreman-installer/commit/b324bda1f2836327658f2255729e0e1ac798d34f pinned things.

evgeni (approved):

ekohl (approved):

PR #895, opened 2023-10-19
Fixes #36843 - run katello-certs-check in fp-certs-generate

ekohl:
Untested now, but submitting for Packit.

### DIFF
--- a/bin/foreman-proxy-certs-generate
+++ b/bin/foreman-proxy-certs-generate
@@ -28,5 +28,32 @@ Kafo::KafoConfigure.hooking.register_pre(:init) do
   org_param.value = organization
 end
 
+Kafo::KafoConfigure.hooking.register_pre_commit(:certs_check) do
+  ca_file   = param('certs', 'server_ca_cert').value
+  cert_file = param('certs', 'server_cert').value
+  key_file  = param('certs', 'server_key').value
+
+  unless app_value(:certs_skip_check) || [ca_file, cert_file, key_file].all?(&:nil?)
+    command = ['katello-certs-check', '-c', cert_file, '-k', key_file, '-b', ca_file]
+
+    logger.debug do
+      require 'shellwords'
+      "Executing #{Shellwords.escape(command)}"
+    end
+
+    begin
+      stdout_stderr, status = Open3.capture2e(command)
+    rescue Errno::ENOENT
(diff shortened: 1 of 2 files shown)


ekohl:
This came up while reviewing https://github.com/theforeman/foreman-documentation/pull/2522 and I think the whole workflow isn't great for users. The aim is run fewer commands (as a user).

ekohl:
Updated. Now locally verified it works in both cases where there are and there aren't custom certs. Also added some other improvements to katello-certs-check.

bin/foreman-proxy-certs-generate
```diff
@@ -28,5 +28,32 @@ Kafo::KafoConfigure.hooking.register_pre(:init) do
   org_param.value = organization
 end
 
+Kafo::KafoConfigure.hooking.register_pre_commit(:certs_check) do
+  ca_file   = param('certs', 'server_ca_cert').value
+  cert_file = param('certs', 'server_cert').value
+  key_file  = param('certs', 'server_key').value
+
+  unless app_value(:certs_skip_check) || [ca_file, cert_file, key_file].all?(&:nil?)
+    command = ['katello-certs-check', '-t', 'foreman-proxy', '-c', cert_file, '-k', key_file, '-b', ca_file]
```
ekohl: This uses the unbranded version. Now I'm debating if we should even pass it at all. Right now it's only used to output commands, which are discarded if the command passes. Should we drop `-t foreman-proxy`?

ehelms: Pass what?

ekohl: The `-t foreman-proxy` part. For Satellite it should be `-t capsule`, but looking at the code this is the only place it's used:
https://github.com/theforeman/foreman-installer/blob/8b7b58dfd4b5d50a391b6c21d9b8db777896fc7f/bin/katello-certs-check#L251C7-L251C7

It's only printed if the exit code is 0, but then we discard the output. So should it be
```suggestion
    command = ['katello-certs-check', '-c', cert_file, '-k', key_file, '-b', ca_file]
```

Or do we imagine the target to become something we care about in our checks?

ehelms: Yes - let's drop it.

evgeni (approved):

PR #900, opened 2023-11-23
Add hammer foreman_rh_cloud package

ShimShtein:


ekohl (changes requested):
You also need to create a migration using `bin/create-migration` for upgrading users with `answers['foreman::cli::rh_cloud'] ||= false`. You can pass the content to stdin.

ekohl:
Like this indeed, but the tests are unhappy now. I'm starting to wonder if those still provide good value, but for now please just make them pass.

ShimShtein:
Done. Tests are happy now.

ekohl (approved):

PR #904, opened 2023-12-15
fix packit to use prod nightly

evgeni:


ehelms (approved):

.packit.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
       centos-stream-8:
```
ekohl, line 34: We should also change this to rhel-8

PR #906, opened 2023-12-18
Fix packit configuration for 3.9 building

ekohl:
This uses the correct spec file and staging repo. Also builds with rhel-8 instead of centos-stream-8 since that's what we end up doing when we really build.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -17,7 +17,7 @@ upstream_tag_template: "{version}"
 
 actions:
   post-upstream-clone:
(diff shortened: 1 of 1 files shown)


.packit.yaml
```diff
@@ -31,10 +31,10 @@ jobs:
   - job: copr_build
     trigger: pull_request
     targets:
-      centos-stream-8:
+      rhel-8:
         additional_modules: "foreman:el8"
         additional_repos:
-          - http://koji.katello.org/releases/yum/foreman-nightly/el8/x86_64/
+          - https://stagingyum.theforeman.org/foreman/3.9/el8/x86_64/
```
evgeni: in other places (also here in `develop`), we went ahead and started to use `yum.tfm.o` directly

evgeni (approved):

PR #909, opened 2024-01-18
Fixes #37063 - Add feature in katello-certs-check to verify if CA bundle has any certificates with trust rules

jpasqualetto:


bin/katello-certs-check
```diff
@@ -182,6 +182,18 @@ function check-ca-bundle-size () {
     fi
 }
 
+function check-ca-bundle-trust-rules () {
+    printf "Checking if CA bundle has trust rules: "
+    CHECK=$(awk -v cmd='openssl x509 -noout -text' ' /BEGIN/{close(cmd)};{print | cmd}'  $CA_BUNDLE_FILE |grep 'Trusted Uses'|wc -l)
+    printf $CHECK
+    if [[ $CHECK -lt 1 ]]; then
+                success
```
ehelms: Can you fix the indentation here to align with the rest of the file?

bin/katello-certs-check
```diff
@@ -182,6 +182,18 @@ function check-ca-bundle-size () {
...
+    else
+            error 10 "The CA bundle contains $CHECK certificate(s) with trust rules. This may create problems for older systems to trust the bundle. Please, recreate the bundle using certificates without trust rules"
+                echo
```
ehelms: Can you fix the indentation here to align with the rest of the file?

ehelms:
Do you think you could add a test case for this? We have certificate fixtures where we generate different of these failing scenarios (https://github.com/theforeman/foreman-installer/blob/develop/spec/fixtures/katello-certs-check/create_cert.sh) and then we test that the script does what we expect when encountering said type (https://github.com/theforeman/foreman-installer/blob/develop/spec/katello_certs_check_spec.rb).

bin/katello-certs-check
jpasqualetto: sure, will do

bin/katello-certs-check
jpasqualetto: I don't see what's wrong with indentation here. What's the problematic line?

jpasqualetto:
I'll have a look at the test case. I may need some help with it.

bin/katello-certs-check
ehelms: The file has 4 space indentation and this all lines should be consistent with it.

jpasqualetto: Ack. Thanks for clarifying.

jpasqualetto:
> I'll have a look at the test case. I may need some help with it.

I pushed an update with a test case and fixed the indentation issues.  Let me know if anything else is needed.

spec/fixtures/katello-certs-check/create_cert.sh
```diff
@@ -28,6 +28,16 @@ else
   echo "CA certificate bundle exists. Skipping."
 fi
 
+CA_BUNDLE=ca-bundle-trust-rules
+CA_CERT_WITH_TRUST_RULES=ca_with_trust_rules
```
ehelms: Could this follow the same name convention of using dashes?

bin/katello-certs-check
```diff
@@ -182,6 +182,18 @@ function check-ca-bundle-size () {
     fi
 }
 
+function check-ca-bundle-trust-rules () {
+    printf "Checking if CA bundle has trust rules: "
+    CHECK=$(awk -v cmd='openssl x509 -noout -text' ' /BEGIN/{close(cmd)};{print | cmd}'  $CA_BUNDLE_FILE |grep 'Trusted Uses'|wc -l)
+    printf $CHECK
+    if [[ $CHECK -lt 1 ]]; then
+		success
```
ehelms:
```suggestion
	success
```

bin/katello-certs-check
```diff
@@ -182,6 +182,18 @@ function check-ca-bundle-size () {
...
+    else
+		error 10 "The CA bundle contains $CHECK certificate(s) with trust rules. This may create problems for older systems to trust the bundle. Please, recreate the bundle using certificates without trust rules"
```
ehelms:
```suggestion
	error 10 "The CA bundle contains $CHECK certificate(s) with trust rules. This may create problems for older systems to trust the bundle. Please, recreate the bundle using certificates without trust rules"
```

bin/katello-certs-check
```diff
@@ -182,6 +182,18 @@ function check-ca-bundle-size () {
...
+		echo
```
ehelms:
```suggestion
	echo
```

bin/katello-certs-check
jpasqualetto: I finally understood what you mean about the indentation. I was copying the indentation of the function check-ca-bundle-size which was wrong.  I was confused because my indentation was matching that one which I assumed was a good template to follow. Guess I was wrong.  I'll adjust the indentation of that one too in my next update.

bin/katello-certs-check
jpasqualetto: ack

spec/fixtures/katello-certs-check/create_cert.sh
jpasqualetto: Sure. Updated

ehelms (approved):

PR #912, opened 2024-02-16
build packit for el9

evgeni:


.packit.yaml
```diff
@@ -36,6 +36,11 @@ jobs:
         additional_repos:
           - https://yum.theforeman.org/releases/nightly/el8/x86_64/
           - https://yum.puppet.com/puppet7/el/8/x86_64/
+      rhel-9:
+        additional_modules: "foreman:el9"
```
ekohl: Build fails with:
> missing groups or modules: foreman:el9

```suggestion
```

ekohl (changes requested):

.packit.yaml
evgeni: indeed, this doesn't need (or use) the -devel module, so it fails

ekohl (approved):

PR #915, opened 2024-02-20
Allow pin_modules to handle a maximum version

ehelms:
We currently have in our Puppetfile `mod 'puppetlabs/apache',               '>= 8.3', '< 12'` and the pin modules step cannot handle having the third argument.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -29,10 +29,12 @@ begin
       PuppetForge.host = url
     end
 
-    def mod(name, options = nil)
+    def mod(name, options = nil, max_version = nil)
       if options.is_a?(Hash) && !options.include?(:ref)
         release = PuppetForge::Module.find(name.tr('/', '-')).current_release
         @new_content << ['mod', name, "~> #{release.version}"]
+      elsif max_version
+        @new_content << ['mod', name, options, max_version]
       else
         @new_content << ['mod', name, options]
       end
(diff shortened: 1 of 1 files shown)


ekohl (dismissed):

Rakefile
```diff
@@ -29,10 +29,12 @@ begin
       PuppetForge.host = url
     end
 
-    def mod(name, options = nil)
+    def mod(name, options = nil, max_version = nil)
       if options.is_a?(Hash) && !options.include?(:ref)
```
ekohl, line 33: Though I now wonder if this needs to be
```suggestion
    def mod(name, max_version = nil, **options)
      if options.any? && !options.include?(:ref)
```

ehelms: I don't think that would work, cause this is dependent on the order as defined in the file, e.g.

`mod 'puppetlabs/apache',               '>= 8.3', '< 12'`

ehelms:
@ekohl was there any additional comment or shall I merge this?

Rakefile
ekohl: Still, you can't put options in the middle. If so, I think it should be
```suggestion
    def mod(name, min_version = nil, max_version = nil, **options)
      if options.any? && !options.include?(:ref)
```

Technically the API is even a bit weaker: https://github.com/puppetlabs/r10k/blob/04b594d57e67591843a56d785666bcf4df81fda3/lib/r10k/module_loader/puppetfile/dsl.rb#L14-L26

So it it's either a hash or there are arguments.

I'll take a stab at it.

ekohl: https://github.com/theforeman/foreman-installer/pull/930 is a bit larger, but that's because I decided to write tests for it.

PR #917, opened 2024-02-21
Drop inclusion of Qpid module

ehelms:
Related PRs:

- https://github.com/theforeman/puppet-katello/pull/491
- https://github.com/theforeman/puppet-foreman_proxy_content/pull/476
- https://github.com/theforeman/puppet-certs/pull/414

Qpid has been dropped for 3 versions of Katello now, so we can safely cleanup all the code paths.

### DIFF
--- a/hooks/boot/04-services.rb
+++ b/hooks/boot/04-services.rb
@@ -9,8 +9,6 @@ module ServicesHookContextExtension
     'httpd.service', # Apache on Red Hat
     'postgresql*', # Used by Foreman/Dynflow and Pulpcore
     'pulpcore*',
-    'qdrouterd.service', # Used by Katello for katello-agent
-    'qpidd.service', # Used by Katello for katello-agent
     'smart_proxy_dynflow_core.service', # Used by Foreman Proxy
     'tomcat.service', # Candlepin
   ].freeze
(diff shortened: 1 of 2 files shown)


hooks/boot/04-services.rb
```diff
@@ -10,7 +10,6 @@ module ServicesHookContextExtension
     'postgresql*', # Used by Foreman/Dynflow and Pulpcore
     'pulpcore*',
     'qdrouterd.service', # Used by Katello for katello-agent
```
evgeni: I'd argue this one can go to?

hooks/boot/04-services.rb
```diff
@@ -10,7 +10,6 @@ module ServicesHookContextExtension
...
-    'qpidd.service', # Used by Katello for katello-agent
     'smart_proxy_dynflow_core.service', # Used by Foreman Proxy
```
evgeni: Unrelated, but this hasn't been a thing since… A while.

evgeni (changes requested):

ehelms:
Cleaned up and added a commit to drop the smart_proxy_dynflow_core service

evgeni (approved):

PR #920, opened 2024-03-18
Check if /opt/puppetlabs is mounted noexec

ekohl:
It was [reported](https://community.theforeman.org/t/problem-foreman-installer-ruby-spawn-errno-enoent/34332/5) that facter can't be found if puppet is installed on a filesystem that's mounted noexec.

### DIFF
--- a/checks/noexec
+++ b/checks/noexec
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+TARGET="/opt/puppetlabs"
+
+if findmnt --target "$TARGET" --noheadings --output options | grep -q noexec ; then
+  echo "$TARGET is on mount with the noexec option." >&2
+  echo "Installation requires access to binaries which are stored under $TARGET and must therefore be mounted with exec." >&2
+  exit 2
+fi


checks/noexec
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+TARGET="/opt/puppetlabs"
+
+if findmnt --target "$TARGET" --noheadings --output options | grep -q noexec ; then
+  echo "$TARGET is on mount with the noexec option which causes problems" >&2
```
ehelms: Perhaps we can be more specific here?

```suggestion
  echo "$TARGET is on mount with the noexec option. Installation requires access to binaries which are stored under $TARGET and must therefore be mounted with exec" >&2
```

evgeni: Sounds good! Also maybe add a newline after the dot. Nobody wants to read long lines ;)

ekohl: I like it.

ekohl:
I also considered the case where non-AIO Puppet is used. In that case you won't have `/opt/puppetlabs` and the check should (silently) pass.

I've debated looking at `/tmp` and `/var/tmp` but I'm not 100% sure we need to mount those with exec now.

checks/noexec
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
```
evgeni, line 1: I think >&2 is a bashism, but sh is dash on Debian ;)

ekohl: I think this works on Debian 12 at least:
```console
$ drr debian:12 /bin/sh
# echo test >&2
test
# echo test >&2 > /dev/null
# 
```

ehelms (approved):
Content is good - merge whenever the sh/bash/dash is resolved

evgeni (approved):
checkbashisms is happy, so I shut up

checks/noexec
evgeni: btw, I'm stupid, the bashism I meant is `&>` (to redirect both streams into one file), `>&2` is of course correct to redirect smth to stderr.

PR #918, opened 2024-03-14
Fixes #37261 Add gitlab ci configuration

dosas:

PR #921, opened 2024-03-19
Add basic tests in packit

ekohl:
This uses tmt to run basic tests on packit builds.

Inspired by https://github.com/theforeman/foreman-installer/pull/918 without duplicating the whole test setup. I've chosen tmt for its integration in packit. It would be interesting to see if this can also be used standalone with a development install.


ehelms:
Do you have any docs on what tmt is, why we want to use it and where we draw the line between GHA and whatever tmt is?

ekohl:
It utilizes [tmt](https://tmt.readthedocs.io/en/stable/) and Packit in turn uses [Testing Farm](https://docs.testing-farm.io/) to run those. The idea is that you run some tests after you've built your package on a real system. If this works, we can give a much better guarantee our packages actually work. Unlike the `%check` phase in RPM, this runs on actually machines (containers or VMs).

nforro:
@ekohl You need to set [`use_internal_tf: true`](https://packit.dev/docs/configuration/upstream/tests#optional-parameters) on the `tests` job if you want to use RHEL composes.

PR #925, opened 2024-03-26
Fixes #37296 - Drop foreman-hooks

adamruzicka:


evgeni:
I think this needs migrations like https://github.com/theforeman/foreman-installer/pull/927

alexjfisher:
Is the plugin broken these days?  I've not used it recently, but use to quite a bit in the past.

adamruzicka:
Not exactly broken, but unmaintained. It was superseded by foreman_webhooks, which should now be the way to go

alexjfisher:
I hadn't seen this.  Just watched https://www.youtube.com/watch?v=sTld0H_PevA though :)

ehelms (approved):

PR #926, opened 2024-03-26
Fixes #37298 - Drop foreman_setup plugin

ekohl:
This plugin is unmaintained and largely broken.


ehelms (approved):

ekohl:
I just noticed I forgot to add the migrations. A case of forgotten `git add`.

PR #931, opened 2024-04-26
Refs #37384 - Update katello/certs to 17.0.1

evgeni:


ekohl (approved):
I wonder if some thing broke again.
```
cannot load such file -- racc/parser.rb
```

It looks like it's running on Ruby 3.3 so perhaps our current code in this repo isn't compatible with that. Since the other CI passes, I think this is OK to ignore for now.

PR #935, opened 2024-04-30
Add certs command

ehelms:
The idea here is to introduce a command that can stand-alone generate certificates as is done within the installer using the same logic of puppet-certs module. This intentionally does not cover deployment, it simply allows generation of the certificates. This relies upon https://github.com/theforeman/puppet-certs/pull/449 as it introduces a class that makes this process works and avoids the complication that the individual certificate classes create via their params.

The full design idea is to split generation from deployment and allow generation to happen prior to full `puppet apply` within the installer and allow the installer part to only handle deployment. This will put us a step closer to allowing users to provider their own certificates by standardizing on the set of certificates as inputs where those inputs can come directly from the user or from the self-generated set.

### DIFF
--- a/bin/foreman-certs
+++ b/bin/foreman-certs
@@ -1,0 +1,8 @@
+#!/usr/bin/env ruby
+require 'rubygems'
+require 'kafo'
+
+CONFIG_DIR = './foreman_certs/config/'.freeze
+
+@result = Kafo::KafoConfigure.run
+exit((@result.nil? || @result.exit_code == 2) ? 0 : @result.exit_code)
--- a/foreman_certs/config/foreman-certs-answers.yaml
+++ b/foreman_certs/config/foreman-certs-answers.yaml
@@ -1,0 +1,6 @@
+certs:
+  generate: true
+  regenerate: true
+  deploy: false
+  group: root
+certs::generate: true
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -17,7 +17,7 @@ upstream_tag_template: "{version}"
 
 actions:
   post-upstream-clone:
-    - "wget https://raw.githubusercontent.com/theforeman/foreman-packaging/rpm/develop/packages/foreman/foreman-installer/foreman-installer.spec -O foreman-installer.spec"
+    - "wget https://raw.githubusercontent.com/evgeni/foreman-packaging/ansible-installer/packages/foreman/foreman-installer/foreman-installer.spec -O foreman-installer.spec"
   get-current-version:
     - "sed 's/-develop//' VERSION"
   create-archive:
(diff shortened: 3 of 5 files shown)


foreman_certs/config/foreman-certs-answers.yaml
```diff
@@ -0,0 +1,6 @@
+certs:
+  generate: true
+  regenerate: true
+  deploy: false
+  group: foreman
```
evgeni:
```suggestion
  group: root
```
Wouldn't that make more sense, making it work in environments where there is no `foreman` group?
(OTOH it probably doesn't matter much as long `deploy` is `false`?)

bin/foreman-certs
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env ruby
+require 'rubygems'
+require 'kafo'
+
+CONFIG_DIR = './foreman_certs/config/'.freeze
+LAST_SCENARIO_PATH = './config/last_scenario.yaml'.freeze
```
evgeni: these are unused, right?

evgeni: argh. kafo uses `defined?(CONFIG_DIR) && CONFIG_DIR)`. whyyyyyyyyyyyyyyyyyyyyy.

last scenario is unused tho

evgeni: https://github.com/theforeman/foreman-installer/pull/1014 is related :)

ehelms:
@evgeni I updated this with updates from https://github.com/theforeman/foreman-installer/pull/1016

ehelms:
This requires https://github.com/theforeman/puppet-certs/pull/449 and will remain in draft until it's merged.

pablomh:
It seems that https://github.com/theforeman/puppet-certs/pull/449 has been already merged for a while.

I've discovered this because while trying to test `foremanctl`, I've hit:
```
TASK [foreman_installer_certs : Enable foreman-installer PR 935 Copr repo] *****
Error: : Task failed: Module failed: Chroot rhel-9-x86_64 does not exist in packit/theforeman-foreman-installer-935
```

evgeni:
this needed a rebase to get the updated version after branching, which I just did

ehelms:
@evgeni at this point, do you think we should just merge this?

evgeni:
Do/will we need this, outside of CI?

ehelms:
Originally I had some ideas for this use case, but this is largely for CI purposes and testing of migration from foreman-installer to `foremanctl`. With the change in design in https://github.com/theforeman/foremanctl/pull/421 we could move to a full foreman-installer run instead of fixtures and partial generation of the certificates.

evgeni:
I'm not opposed in merging this, but we should see that we don't ship it as part of the regular installer package to not confuse (end) users.

evgeni:
@ehelms the alternative would be to put the three files here as fixtures into the `foremanctl` testsuite, given we do not plan to use them anywhere else?

ehelms:
@evgeni How's about https://github.com/theforeman/foremanctl/pull/673 ?

PR #937, opened 2024-05-05
Refs #33974 - Update the db_pool value 

ShimShtein:
The new value should be at least the number of puma threads + 4. 
More details: https://community.theforeman.org/t/rails-connection-pool-size-optimizations/36675

### DIFF
--- a/config/foreman.migrations/20240505103832_change_db_pool_size.rb
+++ b/config/foreman.migrations/20240505103832_change_db_pool_size.rb
@@ -1,0 +1,3 @@
+if answers['foreman'].is_a?(Hash) && answers.dig('foreman', 'db_pool').to_i == 5
+  answers['foreman'].delete('db_pool')
+end
--- a/config/katello.migrations/240506081834-change-db-pool-size.rb
+++ b/config/katello.migrations/240506081834-change-db-pool-size.rb
@@ -1,0 +1,3 @@
+if answers['foreman'].is_a?(Hash) && answers.dig('foreman', 'db_pool').to_i == 5
+  answers['foreman'].delete('db_pool')
+end


ehelms:
This also needs a katello migration -- use https://github.com/theforeman/foreman-installer/blob/develop/bin/create-migration and then copy the logic into the file it creates

config/foreman.migrations/20240505103832_change_db_pool_size.rb
```diff
@@ -0,0 +1,4 @@
+if answers['foreman'].is_a?(Hash) && answers.dig('foreman', 'db_pool')
+  puma_threads_max = answers.dig('foreman', 'foreman_service_puma_threads_max').to_i
+  answers['foreman']['db_pool'] = [answers.dig('foreman', 'db_pool').to_i, puma_threads_max + 4].max
+end
```
evgeni: This would reset *any* `db_pool` setting, while we only want to reset those that have it at the default (`5`), right?
Also, I think it's sufficient to delete the value and let Kafo/Puppet recalculate it, that way you don't need to keep the calculation logic in multiple places

evgeni (changes requested):

ShimShtein:
@ehelms added the same migration also to Katello. 

@evgeni: currently the code makes sure the db_pool size is at least the number we recommend. If the user has the number bigger - we won't change it. This duplicates the logic, but I think it's not that bad, especially since the migration is a one time thing.

config/foreman.migrations/20240505103832_change_db_pool_size.rb
ekohl: If `foreman_service_puma_threads_max` is somehow `nil`, `puma_threads_max` is `0`. Then the value comes down to 0 + 4. Probably an edge case, but I'm a big fan of defensive code.

ShimShtein:
Updated the minimum to 5

evgeni:
[test foreman-installer]

config/katello.migrations/240506081834-change-db-pool-size.rb
```diff
@@ -0,0 +1,3 @@
+if answers['foreman'].is_a?(Hash) && answers.dig('foreman', 'db_pool').to_i == 5
+  answers['foreman']['db_pool'] = nil
```
ekohl:
```suggestion
  answers['foreman'].delete('db_pool')
```

config/foreman.migrations/20240505103832_change_db_pool_size.rb
ekohl, line 1: Please keep the implementations in sync

ekohl (changes requested):
Please use `Refs #33974` in the commit message.

ehelms:
Tests are currently failing as we have dependency mismatch while installer modules are being released.

ekohl:
That should be resolved. Please rebase

ShimShtein:
@ekohl Rebased, the tests are green!

ekohl (approved):

ekohl:
I think @evgeni's concerns are addressed, but leaving it to him to press the green button.

evgeni (approved):

PR #938, opened 2024-05-06
Ensure the system encoding is UTF-8

evgeni:
Checking for C is not sufficient, it could also be ISO-8859 or something
else.

By checking Enconding.locale_charmap we can check one place and have one
way to spell "UTF-8". ENV['LANG'] and ENV['LC_*'] support a variety of
notations which we'd have to be able to parse otherwise.


checks/lang.rb
evgeni, line 1: I wonder if we should rename this to `encoding.rb`?

ekohl: Makes sense to me.

ekohl (approved):

PR #939, opened 2024-05-08
add GHA based tests

evgeni:


.github/workflows/test.yml
```diff
@@ -0,0 +1,60 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - ruby: '2.7'
+            puppet: '7.0'
+          - ruby: '3.0'
+            puppet: '7.0'
+          - ruby: '3.2'
+            puppet: '8.0'
```
evgeni, line 33: I copied this from kafo and do not remember why we had added 3/7 here, but I am sure there was a reason :D

ekohl (approved):

.github/workflows/test.yml
ekohl: On EL9 we run Ruby 3, but often with Puppet 7. It would be even better if we installed actual AIO Puppet packages and avoided installing Puppet via bundler altogether. Our current `Gemfile` doesn't support that, but we could special case it with `PUPPET_VERSION=system`. But let's only do that in develop and cherry pick this PR just to our stable branches.

ekohl: https://github.com/theforeman/foreman-installer/pull/942 would be an attempt at that.

evgeni:
picks in https://github.com/theforeman/foreman-installer/pull/944 and https://github.com/theforeman/foreman-installer/pull/943

PR #942, opened 2024-05-08
Use system Puppet in CI

ekohl:
This uses the official Puppet AIO packages, which is also what we use in production. This makes the tests more realistic.

Currently just testing out if this even works.


.github/workflows/test.yml
```diff
@@ -26,17 +26,23 @@ jobs:
...
     env:
-      PUPPET_VERSION: ${{ matrix.puppet }}
+      BUNDLE_WITHOUT: puppet:rubocop
     steps:
       - uses: actions/checkout@v4
+      - name: Set up Puppet repositories
+        run: |
+          . /etc/os-release
+          wget "https://apt.puppet.com/puppet${{ matrix.puppet }}-release-${VERSION_CODENAME}.deb"
+          sudo apt install "${PWD}/puppet${{ matrix.puppet }}-release-${VERSION_CODENAME}.deb"
```
evgeni, line 42: You need to do ./puppet...., apt refuses local installs from things that don't look like paths

ekohl: That's why I added `${PWD}`. I guess `./` would have also worked.

ekohl:
I think librarian isn't friendly with the AIO Puppet: `cannot load such file -- concurrent`. I suspect it doesn't properly purge the bundler environment variables (like we do in Kafo).

.github/workflows/test.yml
evgeni: ah yes, the race between typing a comment on mobile and forgetting to hit send and you pushing more changes :)

ekohl:
As I wrote somewhere else: I'm leaning to adding a step that builds the modules as separate from the others. Then store it in a cache and reuse it in the other axes.

PR #946, opened 2024-05-14
pkg:generate_source already prints the path to the tarball

evgeni:
create-archive needs to "return" (print) the path, but there is no need to do it twice

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -25,7 +25,6 @@ actions:
     - bundle config set --local without development:test
     - bundle install
     - bundle exec rake pkg:generate_source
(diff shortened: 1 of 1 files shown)


ehelms (approved):

ekohl (approved):
I knew 6dcf5bf413376e33cb8ca80a529975521368e8b2 would one day come in handy.

evgeni:
> I knew [6dcf5bf](https://github.com/theforeman/foreman-installer/commit/6dcf5bf413376e33cb8ca80a529975521368e8b2) would one day come in handy.

1887 days later… ;)

PR #953, opened 2024-07-25
Fixes #37686 - Stop puppetserver if java is too old

ekohl:
When upgrading from puppetserver 7 to puppetserver 8 it'll typically be configured with /usr/bin/java and that usually points to Java 8. Puppetserver 8 has started to require Java 11, so upgrading fails.

By properly stopping the service before running Puppet we know for sure that it can be started safely again. It also helps free up CPU from the constant restart loop, which reduces the slowdown.

### DIFF
--- a/hooks/pre/31-puppet_puppet_server_invalid_java.rb
+++ b/hooks/pre/31-puppet_puppet_server_invalid_java.rb
@@ -1,0 +1,31 @@
+# With Puppetserver 8 the minimum java version has been increased to Java 11
+# If the user is upgrading from Puppet 7 to Puppet 8 then /usr/bin/java may
+# point to Java 8, which is what Puppetserver uses by default. The installer
+# will explicitly configure Java 17, but puppetserver.service can end up in a
+# restart loop where puppet can't properly restart the service.
+#
+# This hook detects Java 8 and Puppetserver 8 and explicitly stops the service.
+# The installer should then reconfigure it and start it again.
+#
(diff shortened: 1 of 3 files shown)


evgeni (approved):
ack once tests pass

hooks/pre/31-puppet_puppet_server_invalid_java.rb
```diff
@@ -0,0 +1,31 @@
...
+# See https://github.com/theforeman/puppet-puppet/pull/910 which is currently
+# only implemented for Red Hat based systems.
+sysconfig_file = '/etc/sysconfig/puppetserver'
+if File.exist?(sysconfig_file)
+  puppetserver_stdout_stderr, _status = execute_command('puppetserver --version', false, true)
+  puppetserver_stdout_stderr&.match(/puppetserver version: (?<version>\d+)\.\d+\.\d+/) do |puppetserver_match|
+    logger.debug("Found Puppetserver #{puppetserver_match[:version]}")
+    if puppetserver_match[:version] == '8'
+      java_stdout_stderr, _status = execute_command("source #{sysconfig_file} ; $JAVA_BIN -version", false, true)
+      java_stdout_stderr&.match(/version "\d+\.(?<version>\d+)\.\d+/) do |java_match|
```
evgeni, line 19: This works well for:
```
openjdk version "1.8.0_412"
```

But not for:
```
openjdk version "17.0.13" 2024-10-15 LTS
```
or
```
openjdk version "11.0.25" 2024-10-15 LTS
```

PR #955, opened 2024-08-06
Fixes #37717 - Update evr extension ownership to foreman

sjha4:
Update ownership of evr extension to owner foreman to allow migrations to operate on it, specifically disable it.

### DIFF
--- a/hooks/boot/01-kafo-hook-extensions.rb
+++ b/hooks/boot/01-kafo-hook-extensions.rb
@@ -1,5 +1,6 @@
 require 'English'
 require 'open3'
+require 'shellwords'
 
 module HookContextExtension
   # FIXME: remove when #23332 is released
@@ -85,6 +86,12 @@ module HookContextExtension
     module_enabled?('foreman_proxy_content')
   end
 
+  def local_db_exists?(database)
+    db_existence_command = pg_sql_statement("SELECT 1 FROM pg_database WHERE datname = '#{database}';")
+    db_existence_output, = execute_preformatted_as('postgres', db_existence_command, false, true)
+    db_existence_output&.strip == '1'
+  end
+
   def log_and_say(level, message, do_say = true, do_log = true)
     style = case level
             when :error
@@ -164,13 +171,19 @@ module HookContextExtension
     execute!(runuser_command, do_say, do_log, extra_env)
   end
 
+  def execute_preformatted_as(user, command, do_say = true, do_log = true, extra_env = {})
+    runuser_command = ['runuser', '-l', user, '-c', command]
+    execute_command(runuser_command, do_say, do_log, extra_env)
+  end
+
   def execute(command, do_say, do_log, extra_env = {})
     _stdout_stderr, status = execute_command(command, do_say, do_log, extra_env)
     status
   end
 
   def execute_command(command, do_say, do_log, extra_env = {})
-    log_and_say(:debug, "Executing: #{command}", do_say, do_log)
+    display_command = command.is_a?(Array) ? Shellwords.join(command) : command
+    log_and_say(:debug, "Executing: #{display_command}", do_say, do_log)
 
     begin
       stdout_stderr, status = Open3.capture2e(*Kafo::PuppetCommand.format_command(command, extra_env))
--- a/hooks/pre/35-change-evr-extension-ownership.rb
+++ b/hooks/pre/35-change-evr-extension-ownership.rb
@@ -1,0 +1,28 @@
+# In Katello 4.16, the 'evr' extension is removed from PostgreSQL and integrated into the Katello database via a migration.
+# This hook ensures the 'evr' extension's ownership is transferred to the 'foreman' user so migrations can act on it.
+
+if (local_foreman_db? || devel_scenario?) && execute("rpm -q postgresql-evr", false, false)
+  if app_value(:noop)
+    logger.debug("Would start postgresql service")
+  else
+    start_services(['postgresql'])
+  end
+
+  database = param_value('foreman', 'db_database') || 'foreman'
+  username = param_value('foreman', 'db_username') || 'foreman'
+
+  if local_db_exists?(database)
+    sql = "UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='#{username}') WHERE extname='evr'"
+    if app_value(:noop)
+      logger.debug("Would execute the following SQL statement to update ownership of the evr extension: #{sql}")
+    else
+      logger.debug("Updating ownership of the evr extension")
+      _, success = execute_preformatted_as('postgres', "psql -d #{database} -c \"#{sql}\"", false, true)
+      unless success
+        fail_and_exit("Failed to update ownership of the evr extension. Please make sure the following sql succeeds before proceeding: #{sql}")
+      end
+    end
+  else
+    logger.notice("The Foreman database #{database} does not exist.")
+  end
+end


ianballou:
@ekohl mentioned before that the `pg_shdepend` table might also need updating, but at least functionally it doesn't seem necessary since the migration passes with @sjha4 's changes alone. The bits that would get updated in `pg_shdepend` may get dropped with the evr extension.

ianballou:
I confirmed that dropping the extension also drops the related entry from `pg_shdepend`.

sjha4:
I tested the installer with the chages here although I had to put the new code in a separate pre-hook file to be run..Suggestions on where to put this code if not this file?

hooks/boot/06-postgresql-upgrade-extensions.rb
```diff
@@ -43,6 +43,11 @@ def postgresql_upgrade(new_version)
 
     execute_as!('postgres', 'vacuumdb --all --analyze-in-stages', false, true)
 
+    if new_version.to_i == 13
```
ekohl: We already had users upgrade their installs on Foreman 3.11 so I don't think this is the right place. Even if we cherry pick this to 3.11, we'll still need a solution for existing Katello users.

I think you need a separate hook. You should check if the `postgresql-evr` package is installed before you execute it.

ekohl (changes requested):

hooks/pre/35-change-evr-extension-ownership.rb
```diff
@@ -0,0 +1,5 @@
+logger.notice("Updating ownership of the evr extension if it is enabled")
+result = execute!("runuser -l postgres -c \"psql -d foreman -c \\\"SELECT 1 from pg_extension WHERE extname='evr';\\\"\"", false, true)
+unless (result.include?('(0 rows)'))
+  execute!("runuser -l postgres -c \"psql -d foreman -c \\\"UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='foreman') WHERE extname='evr';\\\"\"", false, true)
+end
```
sjha4, line 28: @ekohl : Updated to check if extension evr exists in db. This check seems more relevant to updating ownership. The update ownership query itself is idempotent and doesn't fail if evr extension isn't enabled so we don't need any checks before we run it but we can keep the checks if preferred.

Let me know if this works.

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -113,6 +113,7 @@ def execute!(command, do_say = true, do_log = true)
       log_and_say(:error, "#{command} failed! Check the output for error!", do_say, do_log)
       exit 1
     end
+    stdout_stderr
   end
```
sjha4: Making execute! return the output here cause otherwise it's nil and calling function doesn't have access to the output.

evgeni:
What about remote db users? Will those need to perform manual steps (we usually don't have enough permissions to change ownership of objects there) before the upgrade?

sjha4:
> What about remote db users? Will those need to perform manual steps (we usually don't have enough permissions to change ownership of objects there) before the upgrade?

Ya..They'll need to run the update ownership query manually as part of their upgrades. Will need to be documented.

ianballou:
@evgeni docs update will go in https://github.com/theforeman/foreman-documentation/pull/3167 unless someone gets to it first.

ianballou:
Prepping a system for a test.

ianballou:
Found an issue -- the postgres command cannot run because the postgres service is down during the installer run. We'll need to turn postgres on first.

ianballou:
Aside from the postgres issue it's working.

sjha4:
> Aside from the postgres issue it's working.

Ok..So couple of changes..
1. Added a check to run only on local postgres.
2. Added code to check status of postgresql service, start it if stopped and stop it after the hook.

Should work better now.

ianballou:
Tested again by upgrading Katello 4.12 to nightly via 4.13. Worked well.

hooks/pre/35-change-evr-extension-ownership.rb
ekohl, line 1: Please respect noop mode, which you can check with `app_value(:noop)`

hooks/pre/35-change-evr-extension-ownership.rb
```diff
@@ -0,0 +1,17 @@
+if local_postgresql?
+  response, _status = execute_command("sudo systemctl is-active postgresql", false, true)
```
ekohl: The installer already runs as root, so no need for sudo
```suggestion
  response, _status = execute_command("systemctl is-active postgresql", false, true)
```

hooks/pre/35-change-evr-extension-ownership.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  unless response.equal? "active"
+    start_services(['postgresql'])
+  end
+
+  logger.notice("Updating ownership of the evr extension if it is enabled")
+  result = execute!("runuser -l postgres -c \"psql -d foreman -c \\\"SELECT 1 from pg_extension WHERE extname='evr';\\\"\"", false, true)
```
ekohl: Is this needed? I'd expect that the `UPDATE` line is simply a noop if it's not installed.

hooks/pre/35-change-evr-extension-ownership.rb
```diff
@@ -0,0 +1,17 @@
+if local_postgresql?
```
ekohl: I'd also limit it to when `postgresql-evr` is installed
```suggestion
if local_postgresql? && execute("rpm -q postgresql-evr", false, false)
```

We already have that code in 2 other places so perhaps it's also time for a common wrapper method like `package_installed?` `oauth_aio_gem_installed?` already has most of that logic anyway

ekohl (changes requested):

hooks/pre/35-change-evr-extension-ownership.rb
sjha4: Updated..

hooks/pre/35-change-evr-extension-ownership.rb
sjha4: Updated with suggested code..Didn't mess with wrapper right now..

hooks/pre/35-change-evr-extension-ownership.rb
sjha4: Removed the checks..The query is a no-op if evr extension isn't found..

hooks/pre/35-change-evr-extension-ownership.rb
sjha4: Updated code to respect no-op and address other review comments..

sjha4:
Not sure if test failure is related.. 😕

ehelms:
> Not sure if test failure is related.. 😕

Looks like minitar 1.0.0 was released today and we made need to pin it for the time being -- https://rubygems.org/gems/minitar/versions/1.0.0

sjha4:
> > Not sure if test failure is related.. 😕
> 
> Looks like minitar 1.0.0 was released today and we made need to pin it for the time being -- https://rubygems.org/gems/minitar/versions/1.0.0

Ya..That was it..

https://github.com/theforeman/foreman-installer/pull/956

hooks/pre/35-change-evr-extension-ownership.rb
sjha4: Question: What are these numbers before the filenames? I picked 35 cause 34- existed. 🙊

sjha4: Looks like foreman release versions..I can update the filename to 311- to follow convention if that is what it is..

ekohl: They're for order. All hooks get sorted and that determines when they're executed. It allows you to ensure your hook is executed after the PostgreSQL migration itself

ekohl: I'd prefer to implement the noop mode slightly different. In our other hooks we log debug statements what would be done. Otherwise you can simply reduce it to `if !app_value(:noop) &&local_postgresql? && execute("rpm -q postgresql-evr", false, false)`.

hooks/pre/35-change-evr-extension-ownership.rb
```diff
@@ -0,0 +1,24 @@
+# In Katello 4.14, the 'evr' extension is removed from PostgreSQL and integrated into the Katello database via a migration.
+# This hook ensures the 'evr' extension's ownership is transferred to the 'foreman' user so migrations can act on it.
+
+if local_postgresql? && execute("rpm -q postgresql-evr", false, false)
```
ekohl: I think we should also have an `else` for `local_postgresql?` where we check if the extension exists with an incorrect owner. We have all the parameters to instruct `psql` to connect to the right location. You can't fix it, but you can halt the installer to prevent a half-upgraded system. Though that should probably be done in a `pre_commit` hook.

https://github.com/theforeman/puppet-foreman/blob/5f38479c870ef03a4a1565007be68b61e8950d53/manifests/init.pp#L73-L85 are the various answers. foreman_maintain has code that already does something similar, but based on `database.yml`:
https://github.com/theforeman/foreman_maintain/blob/6cc814b8c669bd02a15452fd488a0a6221ed1fd8/lib/foreman_maintain/concerns/base_database.rb#L148-L156

hooks/pre/35-change-evr-extension-ownership.rb
```diff
@@ -0,0 +1,24 @@
...
+  # Ensure the PostgreSQL service is running
+  unless is_postgresql_active || app_value(:noop)
+    logger.debug("Starting postgresql service")
+    start_services(['postgresql'])
+  end
+
+  # Update the ownership of the evr extension
+  unless app_value(:noop)
+    logger.debug("Updating ownership of the evr extension if it is enabled")
+    execute!("runuser -l postgres -c \"psql -d foreman -c \\\"UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='foreman') WHERE extname='evr';\\\"\"", false, true)
```
ekohl: First of all, it can be simplified to use `execute_as!`. And I think you should look at the answers as well. Then the final version is:
```suggestion
    database = param_value('foreman', 'db_database')
    username = param_value('foreman', 'db_username')
    sql = "UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='#{username}') WHERE extname='evr'"
    execute_as!('postgres', "psql -d #{database} -c \"#{sql}\"", false, true)
```

But escaping will be tricky, so I'd enhance `execute_as!` to pass it as an array:
```ruby
def execute_as!(user, command, do_say = true, do_log = true)
  runuser_command = ['runuser', '-l', user, '-c', command] 
  execute!(runuser_command, do_say, do_log)
end
```

Then I'd expect it to work if users chose a different DB and/or username as well.

ekohl (changes requested):

hooks/pre/35-change-evr-extension-ownership.rb
ekohl: I just realized that `Kafo::PuppetCommand.format_command` probably doesn't deal with commands as arrays.

sjha4: Updated to do this:
```
  database = param_value('foreman', 'db_database') || 'foreman'
  username = param_value('foreman', 'db_username') || 'foreman'
  sql = "runuser -l postgres -c \"psql -d '#{database}' -c \\\"UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='#{username}') WHERE extname='evr';\\\"\""
  logger.debug("Executing: #{sql}")
  execute!(sql, false, true) unless app_value(:noop)
```

Still using execute! as escaping was being really painful.

ekohl: I opened https://github.com/theforeman/kafo/pull/374 for that support.

hooks/pre/35-change-evr-extension-ownership.rb
sjha4: Circling back to this, I will look at a pre-commit hook for non-local DBs to get this moving..

hooks/pre/35-change-evr-extension-ownership.rb
```diff
@@ -0,0 +1,26 @@
+# In Katello 4.14, the 'evr' extension is removed from PostgreSQL and integrated into the Katello database via a migration.
```
ianballou:
```suggestion
# In Katello 4.15, the 'evr' extension is removed from PostgreSQL and integrated into the Katello database via a migration.
```

hooks/pre/35-change-evr-extension-ownership.rb
ianballou: I can add the pre_commit hook in a separate PR to keep this one more focused.

sjha4: @ekohl : Ian's got a PR out : https://github.com/theforeman/foreman-installer/pull/984/ to add the pre-commit hook to check ownership on remote dbs. The local DBs would be handled with this PR..

sjha4:
@ekohl : I am not sure I remember if we were supposed to change anything here following the kafo updates..

hooks/pre/35-change-evr-extension-ownership.rb
```diff
@@ -0,0 +1,26 @@
+# In Katello 4.15, the 'evr' extension is removed from PostgreSQL and integrated into the Katello database via a migration.
+# This hook ensures the 'evr' extension's ownership is transferred to the 'foreman' user so migrations can act on it.
+
+if local_postgresql? && execute("rpm -q postgresql-evr", false, false)
+  is_postgresql_active = execute_command("systemctl is-active postgresql", false, true)&.first&.strip == "active"
```
ekohl: Isn't it better to check the exit code is 0 instead of parsing the output? I have no idea if systemctl has localization, but that's one example where it could break.

ekohl (commented):
I'm not sure about the noop mode implementation here. I'd be tempted to change the log messages to `Would do X` instead of Doing x`.

hooks/pre/35-change-evr-extension-ownership.rb
```diff
@@ -0,0 +1,26 @@
...
+  unless is_postgresql_active
+    logger.debug("Starting postgresql service")
+    start_services(['postgresql']) unless app_value(:noop)
+  end
+
+  # Update the ownership of the evr extension
+  logger.debug("Updating ownership of the evr extension if it is enabled")
+  database = param_value('foreman', 'db_database') || 'foreman'
+  username = param_value('foreman', 'db_username') || 'foreman'
+  sql = "psql -d '#{database}' -c \\\"UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='#{username}') WHERE extname='evr';\\\""
```
ehelms: Can you re-factor the way the SQL is handled to re-use the functions added via https://github.com/theforeman/foreman-installer/pull/984/files to ensure no passwords are leaked?

ekohl: This one is a bit different in that we run it as `postgres` (because we assume that's a superuser), relying on ident auth so no actual passwords are used. Since we don't consider the DB name and username to be that secret, it probably is ok.

hooks/pre/35-change-evr-extension-ownership.rb
sjha4: Ack..Updated to check exit status code

ehelms (approved):
LGTM -- I'll let @ekohl have the final say.

hooks/pre/35-change-evr-extension-ownership.rb
```diff
@@ -0,0 +1,26 @@
+# In Katello 4.16, the 'evr' extension is removed from PostgreSQL and integrated into the Katello database via a migration.
+# This hook ensures the 'evr' extension's ownership is transferred to the 'foreman' user so migrations can act on it.
+
+if local_postgresql? && execute("rpm -q postgresql-evr", false, false)
```
ekohl: Probably better to specifically check that Foreman is local instead of there being any local DB:
```suggestion
if local_foreman_db? && execute("rpm -q postgresql-evr", false, false)
```

Though perhaps you can also include `devel_scenario?` in there if you want to accommodate for the katello-devel boxes.

hooks/pre/35-change-evr-extension-ownership.rb
```diff
@@ -0,0 +1,26 @@
...
+  is_postgresql_active = execute_command("systemctl is-active postgresql", false, true)&.last&.exitstatus == 0
```
ekohl: I'm debating simplification and wonder what you think. And I'm only talking about the `op` "mode", not about `noop`.

You can always call `start_services()` even if it was already running, so the check really is only useful for stopping. But do we need to stop the service? The installer will later start it anyway. Perhaps just leave it running?

ekohl (commented):
Style wise I'm not that happy with the many conditionals and prefer the more verbose:
```ruby
if app_value(:noop)
  logger.debug('Would do X')
else
  logger.debug('Doing X')
  x()
end
```

My reasoning is that it's very easy to visually see the control flow.

Additionally, we already have debug statements in various methods, like `start_services` and `stop_services`. So you can simplify it to:

```ruby
if app_value(:noop)
  logger.debug('Would stop service postgresql')
else
  stop_services(['postgresql'])
end
```

I'm wondering what you think about this.

hooks/pre/35-change-evr-extension-ownership.rb
sjha4: Hmm..The idea was to leave the service in the state this hook finds it in..I don't feel strongly about leaving it running after this hook cause it would get stopped by anything after this that needs it stopped. I'll push to simplify the thing..

hooks/pre/35-change-evr-extension-ownership.rb
sjha4: Updated to 
```
if (local_foreman_db? || devel_scenario?) && execute("rpm -q postgresql-evr", false, false)
```

hooks/pre/35-change-evr-extension-ownership.rb
```diff
@@ -0,0 +1,21 @@
+# In Katello 4.16, the 'evr' extension is removed from PostgreSQL and integrated into the Katello database via a migration.
+# This hook ensures the 'evr' extension's ownership is transferred to the 'foreman' user so migrations can act on it.
+
+if (local_foreman_db? || devel_scenario?) && execute("rpm -q postgresql-evr", false, false)
+  if app_value(:noop)
+    logger.debug("Would start postgresql service") if app_value(:noop)
```
ekohl:
```suggestion
    logger.debug("Would start postgresql service")
```

hooks/pre/35-change-evr-extension-ownership.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  # Update the ownership of the evr extension
+  database = param_value('foreman', 'db_database') || 'foreman'
+  username = param_value('foreman', 'db_username') || 'foreman'
+  sql = "psql -d '#{database}' -c \\\"UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='#{username}') WHERE extname='evr';\\\""
+  if app_value(:noop)
+    logger.debug("Would execute: #{sql} to update ownership of the evr extension")
+  else
+    logger.debug("Updating ownership of the evr extension")
+    execute_as!('postgres', sql, false, true)
```
ekohl: This has me thinking: how could this fail and how would a user resolve it?

One I can think of is that the user has `postgresql-evr` installed but the DB doesn't actually exist. I'm wondering if we need to catch that. Would that affect backup restores?

Any others that you can think of?

ekohl (commented):
Thanks for updating. I think this is much easier to reason about and see where it could fail.

hooks/pre/35-change-evr-extension-ownership.rb
ekohl: Thinking out loud further: there's `--reset-data` which could interact in interesting ways

hooks/pre/35-change-evr-extension-ownership.rb
sjha4: Unresolving this and I am going to push an update to use this cause I was running into some escaping issues on the sql command execution.

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -164,6 +164,11 @@ def execute_as!(user, command, do_say = true, do_log = true, extra_env = {})
     execute!(runuser_command, do_say, do_log, extra_env)
   end
 
+  def execute_as_formatted!(user, command, do_say = true, do_log = true, extra_env = {})
```
ekohl: I like using https://github.com/theforeman/kafo/commit/c463e3f38f5477a94460f1b7dbd8d8eedd87801d to greatly reduce the amount of escaping needed, but have some questions about the implementation.

It's called `execute_as_formatted!` and `!` implies it raises an exception if it fails (copied from Rails, see https://stackoverflow.com/questions/612189/why-are-exclamation-marks-used-in-ruby-methods for more discussion on the convention). `execute_command` doesn't, so that's a mismatch.

On the name itself, I think formatted could be confusing because I don't see any formatting. Naming is hard, but any thoughts on `execute_as_user_verbatim`?

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -164,6 +164,11 @@ def execute_as!(user, command, do_say = true, do_log = true, extra_env = {})
...
+    runuser_command = ['runuser', '-l', user, '-c', command]
+    execute_command(runuser_command, do_say, do_log, extra_env)
```
ekohl, line 176: I wonder if `do_say` and `do_log` can deal with an array properly.

hooks/pre/35-change-evr-extension-ownership.rb
sjha4: Tested --reset-data and it was failing..I added a check to confirm if local foreman database exists before proceeding.

For backup-restore, I'd assume the supported workflow would be databases at the same version of foreman where the migrations in [katello PR](https://github.com/Katello/katello/pull/11159) have run.

hooks/boot/01-kafo-hook-extensions.rb
sjha4: How does `execute_preformatted_as` sound?

hooks/boot/01-kafo-hook-extensions.rb
sjha4:
```
2024-12-03 12:38:14 [DEBUG ] [pre] Updating ownership of the evr extension
2024-12-03 12:38:14 [DEBUG ] [root] Executing: ["runuser", "-l", "postgres", "-c", "psql -d foreman -c \"UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='foreman') WHERE extname='evr'\""]
2024-12-03 12:38:14 [DEBUG ] [root] UPDATE 1
```

This is how it's logged..

sjha4: I'll update logging to join the array and log so it's displayed properly.

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -164,13 +170,19 @@ def execute_as!(user, command, do_say = true, do_log = true, extra_env = {})
...
+  end
+
   def execute(command, do_say, do_log, extra_env = {})
     _stdout_stderr, status = execute_command(command, do_say, do_log, extra_env)
     status
   end
 
   def execute_command(command, do_say, do_log, extra_env = {})
-    log_and_say(:debug, "Executing: #{command}", do_say, do_log)
+    display_command = command.is_a?(Array) ? command.join(' ') : command
```
ekohl: Should we use `shelljoin` to properly escape it?

I prefer the more explicit call:
```suggestion
    display_command = command.is_a?(Array) ? Shellwords.join(command) : command
```

But this technically also works:
```suggestion
    display_command = command.is_a?(Array) ? command.shelljoin : command
```

Be sure that `shellwords` is required somewhere

hooks/boot/01-kafo-hook-extensions.rb
```diff
@@ -85,6 +85,12 @@ def pulpcore_enabled?
     module_enabled?('foreman_proxy_content')
   end
 
+  def local_db_exists?(database = 'foreman')
```
ekohl: I don't think we need a default
```suggestion
  def local_db_exists?(database)
```

hooks/pre/35-change-evr-extension-ownership.rb
```diff
@@ -0,0 +1,25 @@
...
+  if local_db_exists?(database)
+    sql = "UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='#{username}') WHERE extname='evr'"
+    if app_value(:noop)
+      logger.debug("Would execute: #{sql} to update ownership of the evr extension")
+    else
+      logger.debug("Updating ownership of the evr extension")
+      execute_preformatted_as('postgres', "psql -d #{database} -c \"#{sql}\"", false, true)
+    end
+  else
+    logger.debug("The foreman database does not exist.")
```
ekohl: Perhaps log the explicit DB name too?
```suggestion
    logger.debug("The Foreman database '#{database}' does not exist.")
```

hooks/pre/35-change-evr-extension-ownership.rb
```diff
@@ -0,0 +1,25 @@
...
+  database = param_value('foreman', 'db_database') || 'foreman'
+  username = param_value('foreman', 'db_username') || 'foreman'
+
+  if local_db_exists?(database)
+    sql = "UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='#{username}') WHERE extname='evr'"
+    if app_value(:noop)
+      logger.debug("Would execute: #{sql} to update ownership of the evr extension")
+    else
+      logger.debug("Updating ownership of the evr extension")
+      execute_preformatted_as('postgres', "psql -d #{database} -c \"#{sql}\"", false, true)
```
ekohl: This will now silently fail. Should it check the status and at least log a warning?

ekohl (commented):
I like how this is shaping up now. Some minor style comments.

hooks/boot/01-kafo-hook-extensions.rb
sjha4:
```
2024-12-03 13:47:28 [DEBUG ] [pre] Starting services postgresql
2024-12-03 13:47:28 [DEBUG ] [root] Executing: runuser -l postgres -c psql\ -t\ -c\ \"SELECT\ 1\ FROM\ pg_database\ WHERE\ datname\ \=\ \'foreman\'\;\"
2024-12-03 13:47:28 [DEBUG ] [root]         1
2024-12-03 13:47:28 [DEBUG ] [root] 
2024-12-03 13:47:28 [DEBUG ] [pre] Updating ownership of the evr extension
2024-12-03 13:47:28 [DEBUG ] [root] Executing: runuser -l postgres -c psql\ -d\ foreman\ -c\ \"UPDATE\ pg_extension\ SET\ extowner\ \=\ \(SELECT\ oid\ FROM\ pg_authid\ WHERE\ rolname\=\'foreman\'\)\ WHERE\ extname\=\'evr\'\"
2024-12-03 13:47:28 [DEBUG ] [root] UPDATE 1
2024-12-03 13:47:28 [DEBUG ] [pre] Hook /usr/share/foreman-installer/hooks/pre/35-change-evr-extension-ownership.rb returned nil
```
Confirmed that both the shell formatted strings can be run directly on the terminal although the number of `\` is making it a little hard to read. 🤷🏼

ekohl (approved):
@evgeni I made may suggestions so I'd appreciate if you gave it a final ACK.

hooks/pre/35-change-evr-extension-ownership.rb
```diff
@@ -0,0 +1,28 @@
...
+    start_services(['postgresql'])
+  end
+
+  database = param_value('foreman', 'db_database') || 'foreman'
+  username = param_value('foreman', 'db_username') || 'foreman'
+
+  if local_db_exists?(database)
+    sql = "UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='#{username}') WHERE extname='evr'"
+    if app_value(:noop)
+      logger.debug("Would execute: #{sql} to update ownership of the evr extension")
```
evgeni:
```suggestion
      logger.debug("Would execute the following SQL statement to update ownership of the evr extension: #{sql}")
```

That way it's more clear where text ends and SQL begins

evgeni (approved):

hooks/pre/35-change-evr-extension-ownership.rb
sjha4: Updated to this message. One more change I hadn't noticed I had locally was updating a couple of log warnings to logger.notice because that seemed to look better in the output.

hooks/pre/35-change-evr-extension-ownership.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  if local_db_exists?(database)
+    sql = "UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='#{username}') WHERE extname='evr'"
+    if app_value(:noop)
+      logger.debug("Would execute the following SQL statement to update ownership of the evr extension: #{sql}")
+    else
+      logger.debug("Updating ownership of the evr extension")
+      _, success = execute_preformatted_as('postgres', "psql -d #{database} -c \"#{sql}\"", false, true)
+      unless success
+        logger.notice("Failed to update ownership of the evr extension.")
```
evgeni: I think failing to update should be a warning?

hooks/pre/35-change-evr-extension-ownership.rb
```diff
@@ -0,0 +1,28 @@
...
+      end
+    end
+  else
+    logger.notice("The Foreman database #{database} does not exist.")
```
evgeni, line 26: When does this happen? When the system is not yet fully installed?

hooks/pre/35-change-evr-extension-ownership.rb
ekohl: Warning or even error that halts execution?

hooks/pre/35-change-evr-extension-ownership.rb
ekohl: In https://github.com/theforeman/foreman-installer/pull/955#discussion_r1866425877 I suggested it. I'm not sure if we'll hit it, but I can imagine there are weird edge cases with `--reset-data`, DB restores or equivalent odd cases that I'd like to handle gracefully.

hooks/pre/35-change-evr-extension-ownership.rb
evgeni: Maybe. It was a warning before, so that's what caught my eye.

What happens if it is not updated? Later the removal of the extension fails?

hooks/pre/35-change-evr-extension-ownership.rb
evgeni: Makes sense. I don't think I care for debug vs notice here.

hooks/pre/35-change-evr-extension-ownership.rb
sjha4: Ya..The evr extension is dropped in katello migration here https://github.com/Katello/katello/pull/11159/ which assumes the foreman user is the owner of the extension. We could fail here instead of later on during db:migrate.

sjha4: Updated to fail and exit:
```
fail_and_exit("Failed to update ownership of the evr extension. Please make sure the following sql succeeds before proceeding: #{sql}")
```

sjha4: @evgeni @ekohl Is this what y'all had in mind?

evgeni (approved):

sjha4:
I will need someone with superpowers to merge this one if this is good to go.. 
Thanks @ekohl , @ehelms , @evgeni and @ianballou  ❤️

ekohl (approved):
Thanks!

PR #958, opened 2024-08-14
Disable reverse proxy by default

ehelms:


spec/migration_spec.rb
```diff
@@ -141,3 +141,21 @@
...
+    context 'reset reverse_proxy to default value to disable by default' do
+      let(:answers) do
+        {
+          'foreman_proxy_content' => {
+            'reverse_proxy' => true,
+          },
+        }
+      end
+
+      it 'changes the default to Redis' do
```
ekohl: Copy-paste error?

spec/migration_spec.rb
```diff
@@ -141,3 +141,21 @@
     end
   end
 end
+
+migration '240814115647-disable-reverse-proxy-by-default' do
```
ekohl: I swear I created a PR for it, but I only had a local branch. Now I opened https://github.com/theforeman/foreman-installer/pull/959 to split it out over multiple files. I'd prefer that over one huge file.

ekohl (changes requested):
:+1: for the migration itself, just some comments on the tests.

spec/migration_spec.rb
ehelms: :facepalm:

spec/migration_spec.rb
ehelms: Rebased on your merged split out

spec/migrations/240814115647-disable_reverse_proxy_by_default_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    context 'reset reverse_proxy to default value to disable by default' do
+      let(:answers) do
+        {
+          'foreman_proxy_content' => {
+            'reverse_proxy' => true,
+          },
+        }
+      end
+
+      it 'changes the default to the puppet module' do
```
ekohl: `it` describes the expected behavior. How about:
```suggestion
      it 'removes the stored answer' do
```

spec/migrations/240814115647-disable_reverse_proxy_by_default_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'spec_helper'
+
+migration '240814115647-disable-reverse-proxy-by-default' do
+  scenarios %w[foreman-proxy-content] do
+    context 'reset reverse_proxy to default value to disable by default' do
```
ekohl: `context` is used to describe the context, typically the `let` blocks. How about
```suggestion
    context 'with a stored answer' do
```

Not asking for it here, but from that you can also easily see the context `without a stored answer`. That could have an `it "doesn't crash" do` or something like that.

At least, that's how I interpret them and structure my RSpec tests. See https://rspec.help/getting-to-know-describe-context/ or https://rspec.help/rspec/basics/ if you're interested in more.

ekohl (approved):
Sorry for not suggesting them earlier. It looks ok, but we can improve the descriptions a bit.

PR #964, opened 2024-08-29
Fixes #37828: Ignore system CA trust when verifying certificates

ehelms:


ehelms:
See https://community.theforeman.org/t/the-use-of-openssl-in-katello-certs-check-with-system-default-ca-path/39233 for details

evgeni (approved):

evgeni:
Probably good to have a redmine for this

ekohl (changes requested):
Like @evgeni I think this should have a Redmine issue.

bin/katello-certs-check
```diff
@@ -157,7 +157,7 @@ function check-priv-key () {
 function check-ca-bundle () {
     printf "Checking CA bundle against the certificate file: "
     ERROR_PATTERN="error [0-9]+ at"
-    CHECK=$(openssl verify -CAfile $CA_BUNDLE_FILE -purpose sslserver -verbose $CERT_FILE 2>&1)
+    CHECK=$(openssl verify -no-CApath -CAfile $CA_BUNDLE_FILE -purpose sslserver -verbose $CERT_FILE 2>&1)
```
evgeni: today I learned (you do not want to know how), that it's sometimes required to also add `-no-CAstore` here

otherwise, at least on Debian with `ca-certificates` installed `/usr/lib/ssl/certs` is still consulted.

ehelms:
Redmine added

evgeni (approved):

PR #965, opened 2024-08-29
Add section on certificates layout

ehelms:
I couldn't find a spot where we laid this out in detail, so I wrote it up and am starting here to capture the information. Perhaps this best lives here, perhaps best in foreman-documentation.

There is some information, that is out of date in some spots, also collected in these locations:

- https://github.com/theforeman/foreman-installer/blob/develop/KATELLO.md
- https://github.com/theforeman/puppet-certs

Consolidating and making this better will be a useful activity that I will continue to look into. I do not want that cleanup to necessarily detract from this information which I hope can be helpful in understanding and being able to make changes in the future.

A few questions:

- Is there anything missing?
- Is there additional information that would be helpful in understanding?


README.md
```diff
@@ -111,6 +111,107 @@ This repository also contains the tooling required to install Foreman Katello.
 
 See [KATELLO.md](/KATELLO.md) for details and specifics of that setup.
 
+# Certificates
+
+## Terminology
+
+* **default CA** - a CA generated by the installer, and used to generate server and client certificates. This CA is used by Candlepin to generate client certificates.
+* **server CA** - a CA that can be provided by the user, or is a copy of the default CA, used by all public facing interfaces
```
ekohl, line 119: This is really Katello terminology that we don't use for Foreman. I'd be explicit in that.

ekohl (commented):
Yesterday I was thinking about writing a blog series "Katello from scratch" which doesn't use the installer but just regular packages and manual work. Could be interesting to discuss certificates at some point.

> * Is there anything missing?
> 
> * Is there additional information that would be helpful in understanding?

I think there are 2 ways of looking at it: the various settings and where files end up being deployed.

I'd like a table with all the installer parameters that affect these and what their default values are.

PR #966, opened 2024-08-30
Rely on official documentation for how to perform operations with the…

ehelms:
… installer

I feel like our official documentation is a much better place for anything installation related and that we can instead focus the README on pointing folks there and including development, contribution and maybe release related operations.


README.md
```diff
@@ -1,102 +1,18 @@
...
-
-Releases of foreman-installer are made via RPM and deb packages and are
-published via the normal Foreman repositories.
+Releases of foreman-installer are made via RPM and deb packages and are published via the normal Foreman repositories.
 
 See the manual and quickstart guide available at [theforeman.org](http://theforeman.org/).
 
-## Using GIT
-
-git clone --recursive git://github.com/theforeman/foreman-installer.git -b develop
```
ekohl, line 21: Fun fact: GitHub disabled their `git://` access a long time ago.

README.md
```diff
@@ -1,102 +1,18 @@
-# Puppet modules for managing Foreman
+# Overview
 
-Installs Foreman as a standalone application or using apache passenger.
-
-Installs Foreman Proxy
-
-May install an example puppet master setup using passenger as well, including the tweaks required for foreman.
+A puppet-based installer that relies on scenarios to provide configuration and default scenario values.
```
ekohl: IMHO Kafo is more important:
```suggestion
A [Kafo](https://github.com/theforeman/kafo)-based installer that relies on scenarios to provide configuration and default scenario values.
```

ekohl (commented):
I largely agree that the manual is better for users. What I think we should document is how this is laid out and how to get started.

For example, creating a migration (use `bin/create-migration`, write a test, etc). Now that I write that, I'm wondering if `create-migration` should also create a skeleton spec file to test.

PR #969, opened 2024-09-04
Cherry pick for Foreman 3.10.1

ekohl:


evgeni (commented):
https://github.com/theforeman/foreman-installer/commit/d7c9decfe5fe8d0a1f551088573f2b44b4672015 would fix the ruby3 issue

evgeni (approved):

ekohl:
I think it's weird it's not picking up puppet_forge 5.0.4 but I think I need to get back to https://github.com/voxpupuli/librarian-puppet/issues/122#issuecomment-2286461454 for that to work.

PR #971, opened 2024-09-09
Fixes #37797 - ensure PostgreSQL upgrade uses the right locales

evgeni:


bastelfreak (approved):
looks good to me (I haven't tested this code, but I used PGSETUP_INITDB_OPTIONS for this in other places)

hooks/boot/06-postgresql-upgrade-extensions.rb
```diff
@@ -12,6 +12,9 @@ def os_needs_postgresql_upgrade?
   def postgresql_upgrade(new_version)
     logger.notice("Performing upgrade of PostgreSQL to #{new_version}")
 
+    start_services(['postgresql'])
+    (_name, _owner, _enconding, collate, ctype, _privileges) = `runuser -l postgres -c 'psql -lt | grep -E "^\s+postgres"'`.chomp.split('|').map(&:strip)
```
ekohl: Can we use `execute_as!` here?

evgeni: not without refactoring, as it currently does not return stdout

ekohl: Oh yes, that also came up in https://github.com/theforeman/foreman-installer/pull/955#discussion_r1709833385. https://github.com/theforeman/kafo/pull/374 may make it easier, but you actually need a shell so that'll break.

evgeni: I could get away without the `|grep` and filter in Ruby, but I didn't feel like I want to open up the refactoring pot right now and the code above is copied from the olde el7-pg12 upgrade :)

evgeni: At least refactored it to use CSV output and drop the `grep`

hooks/boot/06-postgresql-upgrade-extensions.rb
```diff
@@ -12,6 +14,11 @@ def os_needs_postgresql_upgrade?
   def postgresql_upgrade(new_version)
     logger.notice("Performing upgrade of PostgreSQL to #{new_version}")
 
+    start_services(['postgresql'])
+    postgres_list = `runuser -l postgres -c 'psql --list --tuples-only --csv'`
```
bastelfreak, line 18: oh nice!

evgeni: `--csv` is "new" in 12: https://www.postgresql.org/docs/release/12.0/

ekohl (approved):

hooks/boot/06-postgresql-upgrade-extensions.rb
```diff
@@ -12,6 +14,11 @@ def os_needs_postgresql_upgrade?
   def postgresql_upgrade(new_version)
     logger.notice("Performing upgrade of PostgreSQL to #{new_version}")
 
+    start_services(['postgresql'])
```
evgeni, line 17: Argh. This won't work in all scenarios. Especially not in those where f-m already did `services stop` and `swtich-to`. Fuck.

evgeni: https://github.com/theforeman/foreman-installer/pull/974

hooks/boot/06-postgresql-upgrade-extensions.rb
```diff
@@ -35,7 +42,7 @@ def postgresql_upgrade(new_version)
     # see https://bugzilla.redhat.com/show_bug.cgi?id=1935301
     execute!("sed -i '/^data_directory/d' /var/lib/pgsql/data/postgresql.conf", false, true)
 
-    execute_as!('postgres', 'postgresql-setup --upgrade', false, true)
+    execute_as!('postgres', "PGSETUP_INITDB_OPTIONS=\"--lc-collate=#{collate} --lc-ctype=#{ctype} --locale=#{collate}\" postgresql-setup --upgrade", false, true)
```
ekohl, line 45: While trying to understand https://github.com/theforeman/foreman-installer/pull/974 I also noticed there's `PGSETUP_PGUPGRADE_OPTIONS` which is the options to pass to `pg_upgrade`. Would that have been more correct?

evgeni: No, because we need to *init* the cluster with the right locale

PR #974, opened 2024-09-10
Refs #37797 - ensure we can start PostgreSQL by forcing the old version

evgeni:
Something (f-m) might have already called switch-to postgresql:13, which
- upgraded packages
- tried to restart services

The later obviously fails as the on-disk data is in the wrong format.

### DIFF
--- a/hooks/boot/06-postgresql-upgrade-extensions.rb
+++ b/hooks/boot/06-postgresql-upgrade-extensions.rb
@@ -2,9 +2,13 @@ require 'csv'
 
 module PostgresqlUpgradeHookContextExtension
   def needs_postgresql_upgrade?(new_version)
-    File.read('/var/lib/pgsql/data/PG_VERSION').chomp.to_i < new_version.to_i
+    current_version.to_i < new_version.to_i
   rescue Errno::ENOENT
     false
+  end
+
+  def current_version
+    File.read('/var/lib/pgsql/data/PG_VERSION').chomp
   end
 
   def os_needs_postgresql_upgrade?
@@ -14,6 +18,7 @@ module PostgresqlUpgradeHookContextExtension
   def postgresql_upgrade(new_version)
     logger.notice("Performing upgrade of PostgreSQL to #{new_version}")
 
+    execute!("dnf module switch-to postgresql:#{current_version} -y", false, true)
     start_services(['postgresql'])
     postgres_list = `runuser -l postgres -c 'psql --list --tuples-only --csv'`
     postgres_databases = CSV.parse(postgres_list)
--- a/spec/postgresql_upgrade_hook_context_extension_spec.rb
+++ b/spec/postgresql_upgrade_hook_context_extension_spec.rb
@@ -31,6 +31,7 @@ describe PostgresqlUpgradeHookContextExtension do
     subject { context.postgresql_upgrade(13) }
 
     before do
+      allow(File).to receive(:read).with('/var/lib/pgsql/data/PG_VERSION').and_return('12')
       allow(context).to receive(:logger).and_return(logger)
       allow(context).to receive(:'execute!')
       allow(context).to receive(:ensure_packages)


hooks/boot/06-postgresql-upgrade-extensions.rb
```diff
@@ -2,18 +2,23 @@
...
+  end
+
   def os_needs_postgresql_upgrade?
     el8? && needs_postgresql_upgrade?(13)
   end
 
   def postgresql_upgrade(new_version)
     logger.notice("Performing upgrade of PostgreSQL to #{new_version}")
 
+    execute!("dnf module switch-to postgresql:#{current_version} -y", false, true)
```
evgeni, line 21: Yes this means we execute:
- (maintain/user) switch-to 13
- (installer) switch-to 12
- (installer) switch-to 13

But I didn't find another way to get the locale of the existing (12) cluster w/o it being up, and it can only be up if it's PostgreSQL 12 that's installed on the system.

ehelms: Do we need to move the PG13 upgrade into f-maintain and have it in two spots based on entry point?

evgeni: No, because the user (w/o maintain) also needs to perform `switch-to 13` to get the new installer.

We need to get rid of this monstrosity called modularity

ehelms: I guess you have figured it all out by now so might as well take this in. My second thought when reading this is -- should we add a check to catch this before the upgrade is started and have the user take a corrective action first?

ehelms: Also, is this user / cluster locale mismatch a known issue within PG land?

ekohl: `postgresql-upgrade` installs `/usr/lib64/pgsql/postgresql-12` which has (as I can see) a full PG 12 installation. I don't know how realistic it is to use this, but that could be a solution. It's also what the upgrade command itself uses.

evgeni: > I guess you have figured it all out by now so might as well take this in. My second thought when reading this is -- should we add a check to catch this before the upgrade is started and have the user take a corrective action first?

What kind of corrective action? And what check? The system is "broken" the moment switch-to is executed, regardless what executes kt

evgeni: > `postgresql-upgrade` installs `/usr/lib64/pgsql/postgresql-12` which has (as I can see) a full PG 12 installation. I don't know how realistic it is to use this, but that could be a solution. It's also what the upgrade command itself uses.

That's technically correct, but requires us to correctly start the cluster ourselves instead of relying on the existing systemd unit.

ekohl: Can we use `systemd-run postgresql --property` to run it temporarily with a different `ExecStart`?

ekohl: Looking better at it, `systemd-run` doesn't load an existing unit file. You can only specify the fields

evgeni: > Also, is this user / cluster locale mismatch a known issue within PG land?

Yes. The Debian tooling does this (extracting the locale of the old cluster and using it for the new) automatically, RHEL's does not.

evgeni: And I've opened https://issues.redhat.com/browse/RHEL-58410 to track this problem as a whole

ehelms (approved):
You've convinced me there is no good other option right now.

hooks/boot/06-postgresql-upgrade-extensions.rb
evgeni: Well, turned out starting a cluster in single-user mode is not *that* hard, and solves stupid problems with dnf…

https://github.com/theforeman/foreman-installer/pull/982

PR #976, opened 2024-09-16
accept ECC keys in katello-certs-check

evgeni:


bin/katello-certs-check
```diff
@@ -145,9 +145,9 @@ function check-passphrase () {
 
 function check-priv-key () {
     printf "Checking to see if the private key matches the certificate: "
-    CERT_MOD=$(openssl x509 -noout -modulus -in $CERT_FILE)
-    KEY_MOD=$(openssl rsa -noout -modulus -in $KEY_FILE)
-    if [[ "$CERT_MOD" != "$KEY_MOD" ]]; then
+    CERT_PUBKEY_SUM=$(openssl x509 -noout -pubkey -in $CERT_FILE | openssl sha512)
+    KEY_PUBKEY_SUM=$(openssl pkey -pubout -in $KEY_FILE | openssl sha512)
```
evgeni, line 149: ECC keys do not have a modulus.
But we can instead just make OpenSSL print the pubkey from the cert and from the private key, and compare those two.

ehelms (approved):

PR #978, opened 2024-09-17
Rewrite katello-certs-check to Ruby

ekohl:
This makes it easier to maintain because the Ruby API is easier to handle than "parsing" the OpenSSL command line output.

As you can see, still very much a work in progress but I'm opening this up for early feedback on the concept.

### DIFF
--- a/bin/katello-certs-check.rb
+++ b/bin/katello-certs-check.rb
@@ -71,9 +71,11 @@ rescue Exception => e
 end
 
 key = check("Reading private key '#{key_path}'") do
-  OpenSSL::PKey.read(File.read(key_path))
+  OpenSSL::PKey.read(File.read(key_path), '')
 rescue Exception => e
-  # TODO: check if there's a password?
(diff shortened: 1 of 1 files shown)


bin/katello-certs-check.rb
```diff
@@ -0,0 +1,175 @@
...
+chain = check('Verify certificate is singed by bundle') do
+  unless store.verify(cert)
+    raise FailedCheck, "Certificate is not signed by bundle: #{store.error_string}"
+  end
+  store.chain
+end
+
+# TODO
+check('Verifying root is included in bundle has CA:TRUE') do
+  # TODO: oid for basicConstraints?
```
ekohl, line 130: I get the impression you can rely on the name, but I wonder if that's always true.

bin/katello-certs-check.rb
ekohl, line 1: Ideally this would overwrite the previous implementation and a drop in replacement, but now it's useful to have them side by side to compare.

PR #982, opened 2024-10-02
Fixes #37874 - don't try to swap modules, this can fail

evgeni:
instead use single-user mode cluster based on the postgresql-upgrade
binaries

### DIFF
--- a/hooks/boot/06-postgresql-upgrade-extensions.rb
+++ b/hooks/boot/06-postgresql-upgrade-extensions.rb
@@ -1,5 +1,3 @@
-require 'csv'
-
 module PostgresqlUpgradeHookContextExtension
   def needs_postgresql_upgrade?(new_version)
     current_version.to_i < new_version.to_i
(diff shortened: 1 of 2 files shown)


hooks/boot/06-postgresql-upgrade-extensions.rb
```diff
@@ -41,6 +33,20 @@ def postgresql_upgrade(new_version)
...
+    # doesn't detect those on its own: https://issues.redhat.com/browse/RHEL-58410
+    # dnf switch-to has already replaced the packages and the cluster is down
+    # We have to start the cluster in single-user mode with the old postgres binary
+    postgres_result = `echo "select datcollate,datctype from pg_database where datname='postgres';" | runuser -l postgres -c '/usr/lib64/pgsql/postgresql-#{current_version}/bin/postgres --single -D /var/lib/pgsql/data postgres'`
+    data = {}
+    postgres_result.each_line do |line|
+      line.match(/dat(collate|ctype)\s+=\s+(\S+)/) do |m|
+        data[m[1]] = m[2].delete('"')
+      end
+    end
```
ehelms, line 46: What an ewwww block.

ehelms (approved):

hooks/boot/06-postgresql-upgrade-extensions.rb
evgeni: Yeah, the output in single user mode is a bit eww (well, it's used for debugging, not running real stuff), and we gotta parse that. :/

ekohl: Using `/dat(collate|ctype)\s+=\s+"(\S+)"/` as the regex removes the need for `.delete('"')`, making it more ew than ewwww.

I do wonder about error handling: what if `data` doesn't contain any values because the process failed to start up?

evgeni: Then the upgrade will fail as we pass invalid data in `PGSETUP_INITDB_OPTIONS` -- nothing new compared to how it was before this whole hack :see_no_evil:

PR #984, opened 2024-10-04
Fixes #37883 - halt if remote DB does not own EVR

ianballou:
Checks if the `postgres` user owns the `evr` extension. If it does, halt the foreman-installer and report an error with the command to run to fix the permissions.

Also check if `postgres` is the foreman DB owner. In that case, don't do anything because the migration will succeed.

There's no provision if the admin user is called something other than postgres. Should there be?

### DIFF
--- a/hooks/pre_commit/42-evr_extension_permissions.rb
+++ b/hooks/pre_commit/42-evr_extension_permissions.rb
@@ -2,29 +2,30 @@
 return if local_postgresql?
 
 database = param_value('foreman', 'db_database') || 'foreman'
-username = param_value('foreman', 'db_username') || 'foreman'
-password = param_value('foreman', 'db_password')
-host = param_value('foreman', 'db_host')
-port = param_value('foreman', 'db_port') || 5432
+config = load_db_config(database)
 
 # If postgres is the owner of the DB, then the permissions will not matter.
-return if username == 'postgres'
+return if config[:username] == 'postgres'
 
 check_evr_owner_sql = "SELECT CASE" \
-                      " WHEN r.rolname = 'postgres' THEN 1" \
-                      " ELSE 0" \
+                      " WHEN r.rolname = '#{config[:username]}' THEN 0" \
+                      " ELSE 1" \
                       " END AS evr_owned_by_postgres" \
                       " FROM pg_extension e" \
                       " JOIN pg_roles r ON e.extowner = r.oid" \
                       " WHERE e.extname = 'evr';"
 
-command = "PGPASSWORD='#{password}' psql -U #{username} -h #{host} -p #{port} -d #{database} -t -c \"#{check_evr_owner_sql}\""
+command = pg_sql_statement(check_evr_owner_sql)
 logger.debug "Checking if the evr extension is owned by the postgres user via #{command}"
-output, = execute_command(command, false, true)
-unless output.nil?
-  if output.strip == '1'
-    fail_and_exit("The evr extension is owned by postgres and not the foreman DB owner. Please run the following command to fix it: " \
-                  "UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='#{username}');")
-  end
-end
+output, = execute_command(command, false, true, pg_env(config))
 
+case output&.strip
+when '0'
+  # The evr extension is owned by the foreman DB owner, so we can skip this check.
+  return
+when '1'
+  fail_and_exit("The evr extension is not owned by the #{database} DB owner. Please run the following command to fix it: " \
+                "UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='#{config[:username]}') WHERE extname='evr';")
+else
+  fail_and_exit("Failed to check the ownership of the evr extension.")
+end
--- a/hooks/pre_commit/42-evr_extension_permissions.rb
+++ b/hooks/pre_commit/42-evr_extension_permissions.rb
@@ -1,0 +1,30 @@
+# Managed databases will be handled automatically.
+return if local_postgresql?
+
+database = param_value('foreman', 'db_database') || 'foreman'
+username = param_value('foreman', 'db_username') || 'foreman'
+password = param_value('foreman', 'db_password')
+host = param_value('foreman', 'db_host')
+port = param_value('foreman', 'db_port') || 5432
+
+# If postgres is the owner of the DB, then the permissions will not matter.
+return if username == 'postgres'
+
+check_evr_owner_sql = "SELECT CASE" \
+                      " WHEN r.rolname = 'postgres' THEN 1" \
+                      " ELSE 0" \
+                      " END AS evr_owned_by_postgres" \
+                      " FROM pg_extension e" \
+                      " JOIN pg_roles r ON e.extowner = r.oid" \
+                      " WHERE e.extname = 'evr';"
+
+command = "PGPASSWORD='#{password}' psql -U #{username} -h #{host} -p #{port} -d #{database} -t -c \"#{check_evr_owner_sql}\""
+logger.debug "Checking if the evr extension is owned by the postgres user via #{command}"
+output, = execute_command(command, false, true)
+unless output.nil?
+  if output.strip == '1'
+    fail_and_exit("The evr extension is owned by postgres and not the foreman DB owner. Please run the following command to fix it: " \
+                  "UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='#{username}');")
+  end
+end
+
--- a/hooks/pre_commit/42-evr_extension_permissions.rb
+++ b/hooks/pre_commit/42-evr_extension_permissions.rb
@@ -7,6 +7,13 @@ config = load_db_config('foreman')
 # If postgres is the owner of the DB, then the permissions will not matter.
 return if config[:username] == 'postgres'
 
+evr_existence_command = pg_sql_statement("SELECT 1 FROM pg_extension WHERE extname = 'evr';")
+logger.debug "Checking if the evr extension exists via #{evr_existence_command}"
+evr_existence_output, = execute_command(evr_existence_command, false, true, pg_env(config))
+
+# If the evr extension does not exist, then we can skip this check.
+return if evr_existence_output&.strip != '1'
+
 check_evr_owner_sql = "SELECT CASE" \
                       " WHEN r.rolname = '#{config[:username]}' THEN 0" \
                       " ELSE 1" \
--- a/hooks/pre_commit/42-evr_extension_permissions.rb
+++ b/hooks/pre_commit/42-evr_extension_permissions.rb
@@ -1,8 +1,8 @@
 # Managed databases will be handled automatically.
 return if local_postgresql?
+return unless katello_enabled?
 
-database = param_value('foreman', 'db_database') || 'foreman'
-config = load_db_config(database)
+config = load_db_config('foreman')
 
 # If postgres is the owner of the DB, then the permissions will not matter.
 return if config[:username] == 'postgres'
@@ -24,7 +24,7 @@ when '0'
   # The evr extension is owned by the foreman DB owner, so we can skip this check.
   return
 when '1'
-  fail_and_exit("The evr extension is not owned by the #{database} DB owner. Please run the following command to fix it: " \
+  fail_and_exit("The evr extension is not owned by the foreman DB owner. Please run the following command to fix it: " \
                 "UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='#{config[:username]}') WHERE extname='evr';")
 else
   fail_and_exit("Failed to check the ownership of the evr extension.")
--- a/hooks/boot/01-kafo-hook-extensions.rb
+++ b/hooks/boot/01-kafo-hook-extensions.rb
@@ -101,6 +101,50 @@ module HookContextExtension
     Kafo::KafoConfigure.logger.send(level, message) if do_log
   end
 
+  def load_db_config(db)
+    case db
+    when 'foreman'
+      module_name = 'foreman'
+      user_param = 'username'
+      db_param = 'database'
+      param_prefix = 'db_'
+    when 'candlepin'
+      module_name = 'katello'
+      user_param = 'user'
+      db_param = 'name'
+      param_prefix = 'candlepin_db_'
+    when 'pulpcore'
+      module_name = 'foreman_proxy_content'
+      user_param = 'user'
+      db_param = 'db_name'
+      param_prefix = 'pulpcore_postgresql_'
+    else
+      raise "installer module unknown for db: #{db}"
+    end
+
+    {
+      host: param_value(module_name, "#{param_prefix}host") || 'localhost',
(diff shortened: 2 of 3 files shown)


ianballou:
To-do: skip the entire hook if the database is local. 
Edit: done

hooks/pre_commit/42-evr_extension_permissions.rb
```diff
@@ -0,0 +1,27 @@
...
+                      " WHEN r.rolname = 'postgres' THEN 1" \
+                      " ELSE 0" \
+                      " END AS evr_owned_by_postgres" \
+                      " FROM pg_extension e" \
+                      " JOIN pg_roles r ON e.extowner = r.oid" \
+                      " WHERE e.extname = 'evr';"
+
+command = "PGPASSWORD='#{password}' psql -U #{username} -h #{host} -p #{port} -d #{database} -t -c \"#{check_evr_owner_sql}\""
+logger.debug "Checking if the evr extension is owned by the postgres user via #{command}"
+output = execute!(command, false, true).strip
```
ehelms: For me this `execute!` returns an empty string which causes this `.strip` to be run on `nil` and throws an error:

```
  stderr: |-
    /usr/share/foreman-installer/hooks/pre_commit/42-evr_extension_permissions.rb:23:in `block (4 levels) in load': undefined method `strip' for nil:NilClass (NoMethodError)
            from /usr/share/gems/gems/kafo-7.6.0/lib/kafo/hooking.rb:36:in `instance_eval'
            from /usr/share/gems/gems/kafo-7.6.0/lib/kafo/hooking.rb:36:in `block (4 levels) in load'
```

ehelms:
Tested and I see the proper error message:

```
    2024-10-24 13:55:54 [NOTICE] [root] Loading installer configuration. This will take some time.
    2024-10-24 13:55:56 [NOTICE] [root] Running installer with log based terminal output at level NOTICE.
    2024-10-24 13:55:56 [NOTICE] [root] Use -l to set the terminal output log level to ERROR, WARN, NOTICE, INFO, or DEBUG. See --full-help for definitions.
    2024-10-24 13:55:57 [NOTICE] [checks] System checks passed
    The evr extension is owned by postgres and not the foreman DB owner. Please run the following on the foreman DB to fix it: UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='foreman');
    2024-10-24 13:55:59 [ERROR ] [root] The evr extension is owned by postgres and not the foreman DB owner. Please run the following on the foreman DB to fix it: UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='foreman');
```

I ran (at least I think I ran) the recommended update `UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='foreman');` command on the database but it keeps telling me that things are not correct on the remote database. Any ideas?

I tested this using some new Forklift code to make it easier (https://github.com/theforeman/forklift/pull/1871). This is needed locally to get all the PRs:

```
diff --git a/pipelines/external_database.yml b/pipelines/external_database.yml
index 1e0021c4..0f1b85c8 100644
--- a/pipelines/external_database.yml
+++ b/pipelines/external_database.yml
@@ -19,6 +19,11 @@
   become: yes
   vars_files:
     - vars/external_database.yml
+  vars:
+    packit_prs:
+      - https://github.com/theforeman/foreman-installer/pull/984
+      - https://github.com/Katello/katello/pull/11159
+    postgresql_use_evr: false
   roles:
     - role: forklift_versions
       scenario: "{{ pipeline_type }}"
@@ -27,6 +32,7 @@
     - role: foreman_server_repositories
     - role: etc_hosts
     - role: update_os_packages
+    - role: packit
     - role: foreman_installer
       foreman_installer_options_internal_use_only:
         - "--foreman-db-manage false"
```

ehelms:
From discussion with @ianballou, a check that evr extension will be added to this hook since for fresh installations the EVR extension would not be expected to have been installed.

ianballou:
@ehelms I fixed up the PR, it should be okay now. I haven't tested it yet for nil outputs, but the nil check there should do the trick.

ehelms:
@ianballou Can you fix up the rubocop issue?

ehelms:
We should also add a similar check to foreman-maintain to test the ownership before upgrading for the best user experience.

hooks/pre_commit/42-evr_extension_permissions.rb
```diff
@@ -0,0 +1,30 @@
...
+check_evr_owner_sql = "SELECT CASE" \
+                      " WHEN r.rolname = 'postgres' THEN 1" \
+                      " ELSE 0" \
+                      " END AS evr_owned_by_postgres" \
+                      " FROM pg_extension e" \
+                      " JOIN pg_roles r ON e.extowner = r.oid" \
+                      " WHERE e.extname = 'evr';"
+
+command = "PGPASSWORD='#{password}' psql -U #{username} -h #{host} -p #{port} -d #{database} -t -c \"#{check_evr_owner_sql}\""
+logger.debug "Checking if the evr extension is owned by the postgres user via #{command}"
```
ekohl, line 26: This logs the password. If you use the above solution of passing env vars you'll avoid that.

hooks/pre_commit/42-evr_extension_permissions.rb
```diff
@@ -0,0 +1,30 @@
...
+
+check_evr_owner_sql = "SELECT CASE" \
+                      " WHEN r.rolname = 'postgres' THEN 1" \
+                      " ELSE 0" \
+                      " END AS evr_owned_by_postgres" \
+                      " FROM pg_extension e" \
+                      " JOIN pg_roles r ON e.extowner = r.oid" \
+                      " WHERE e.extname = 'evr';"
+
+command = "PGPASSWORD='#{password}' psql -U #{username} -h #{host} -p #{port} -d #{database} -t -c \"#{check_evr_owner_sql}\""
```
ekohl: In 97487cf60719664b3f3c6fa6d5f01779ec4fdcd9 we've moved to passing all variables an environment variables. I think you could even reuse those methods. Probably something like
```ruby
config = load_db_config(db)
command = pg_sql_statement(check_evr_owner_sql)
output, = execute_command(command, false, true, pg_env(config))
```

hooks/pre_commit/42-evr_extension_permissions.rb
```diff
@@ -0,0 +1,30 @@
...
+username = param_value('foreman', 'db_username') || 'foreman'
+password = param_value('foreman', 'db_password')
+host = param_value('foreman', 'db_host')
+port = param_value('foreman', 'db_port') || 5432
+
+# If postgres is the owner of the DB, then the permissions will not matter.
+return if username == 'postgres'
+
+check_evr_owner_sql = "SELECT CASE" \
+                      " WHEN r.rolname = 'postgres' THEN 1" \
```
ekohl: Wouldn't it make more sense to verify that the owner equals the username to connect to the DB?

hooks/pre_commit/42-evr_extension_permissions.rb
```diff
@@ -0,0 +1,30 @@
...
+                      " END AS evr_owned_by_postgres" \
+                      " FROM pg_extension e" \
+                      " JOIN pg_roles r ON e.extowner = r.oid" \
+                      " WHERE e.extname = 'evr';"
+
+command = "PGPASSWORD='#{password}' psql -U #{username} -h #{host} -p #{port} -d #{database} -t -c \"#{check_evr_owner_sql}\""
+logger.debug "Checking if the evr extension is owned by the postgres user via #{command}"
+output, = execute_command(command, false, true)
+unless output.nil?
+  if output.strip == '1'
```
ekohl: If the command somehow fails, you won't capture it. Perhaps assert that it returns 0?
```suggestion
if output&.strip != '0'
```

Or
```ruby
case output&.strip
when '0'
  # All good
when '1'
  fail_and_exit('incorrect owner')
else
  fails_and_exit('Unable to verify owner')
end
```

hooks/pre_commit/42-evr_extension_permissions.rb
```diff
@@ -0,0 +1,30 @@
...
+    fail_and_exit("The evr extension is owned by postgres and not the foreman DB owner. Please run the following command to fix it: " \
```
ekohl:
```suggestion
    fail_and_exit("The evr extension is owned by postgres and not the #{database} DB owner. Please run the following command to fix it: " \
```

ekohl (changes requested):
At least the logging of the password needs to be addressed.

hooks/pre_commit/42-evr_extension_permissions.rb
ianballou: Good catch :+1:

hooks/pre_commit/42-evr_extension_permissions.rb
ianballou: The case logic there looks good to me.

hooks/pre_commit/42-evr_extension_permissions.rb
ianballou: Yeah, and that way we wouldn't have to assume that postgres is the default admin. I'll try it out.

ianballou:
@ekohl I believe I've addressed all of your comments. I had to move some methods around to use them in both  this hook and the database reset one.

hooks/pre_commit/42-evr_extension_permissions.rb
```diff
@@ -0,0 +1,31 @@
...
+logger.debug "Checking if the evr extension is owned by the postgres user via #{command}"
+output, = execute_command(command, false, true, pg_env(config))
+
+case output&.strip
+when '0'
+  # The evr extension is owned by the foreman DB owner, so we can skip this check.
+  return
+when '1'
+  fail_and_exit("The evr extension is not owned by the #{database} DB owner. Please run the following command to fix it: " \
+                "UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='#{config[:username]}');")
```
evgeni: this updates *all* extensions, not only `evr`!
```suggestion
                "UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='#{config[:username]}') WHERE extname='evr';")
```

evgeni:
> We should also add a similar check to foreman-maintain to test the ownership before upgrading for the best user experience.

I consider not having this check a blocker, TBH.

hooks/pre_commit/42-evr_extension_permissions.rb
ianballou: I kept this because it felt like all of the Foreman DB extensions should be owned by Foreman, but that isn't strictly necessary. I can add the WHERE clause.

evgeni: they aren't on a fresh install, so we shouldn't modify them

(the longer version: just doing `UPDATE pg_extension SET extowner = …` is *technically* a super bad idea -- you also gotta update all the data *owned* by that extention, it's just damn luck that we don't have to do it here due to the way how we ended up with the wrong owner to begin with).

ianballou: Makes sense, I've updated it to only update the evr extension.

hooks/pre_commit/42-evr_extension_permissions.rb
```diff
@@ -0,0 +1,31 @@
+# Managed databases will be handled automatically.
+return if local_postgresql?
+
+database = param_value('foreman', 'db_database') || 'foreman'
+config = load_db_config(database)
```
evgeni:
```suggestion
config = load_db_config('foreman')
```

While you can deploy the Foreman DB with a different DB name (no idea why you would, but you can), `load_db_config` only accepts `foreman` (and `candlepin` and `pulpcore`) keys.

hooks/pre_commit/42-evr_extension_permissions.rb
```diff
@@ -0,0 +1,31 @@
+# Managed databases will be handled automatically.
+return if local_postgresql?
```
evgeni, line 2:
```suggestion
return if local_postgresql?
return unless katello_enabled?
```

Otherwise this will run on Proxies and plain Foreman installs, which do not have the EVR extension, which will make the SQL below return no rows and run into the `else` branch, raising `fail_and_exit("Failed to check the ownership of the evr extension.")`

ianballou:
I realized that we don't need to check the `evr` permissions if the extension does not exist, so I added a check for that earlier in the script.

ianballou:
Foreman maintain side: https://github.com/theforeman/foreman_maintain/pull/953

evgeni (approved):
Should be squashed

ehelms (approved):

hooks/pre_commit/42-evr_extension_permissions.rb
```diff
@@ -0,0 +1,38 @@
...
```
ekohl, line 2: You can't return outside of a function and it broke nightly pipelines
```
[2024-11-20T22:10:35.810Z]     /usr/share/foreman-installer/hooks/pre_commit/42-evr_extension_permissions.rb:2:in `block (4 levels) in load': unexpected return (LocalJumpError)
[2024-11-20T22:10:35.810Z]             from /usr/share/gems/gems/kafo-7.6.0/lib/kafo/hooking.rb:36:in `instance_eval'
[2024-11-20T22:10:35.810Z]             from /usr/share/gems/gems/kafo-7.6.0/lib/kafo/hooking.rb:36:in `block (4 levels) in load'
[2024-11-20T22:10:35.810Z]             from /usr/share/gems/gems/kafo-7.6.0/lib/kafo/hook_context.rb:19:in `instance_eval'
[2024-11-20T22:10:35.810Z]             from /usr/share/gems/gems/kafo-7.6.0/lib/kafo/hook_context.rb:19:in `execute'
[2024-11-20T22:10:35.810Z]             from /usr/share/gems/gems/kafo-7.6.0/lib/kafo/hooking.rb:67:in `block in execute'
[2024-11-20T22:10:35.810Z]             from /usr/share/gems/gems/kafo-7.6.0/lib/kafo/hooking.rb:65:in `each'
[2024-11-20T22:10:35.810Z]             from /usr/share/gems/gems/kafo-7.6.0/lib/kafo/hooking.rb:65:in `execute'
[2024-11-20T22:10:35.810Z]             from /usr/share/gems/gems/kafo-7.6.0/lib/kafo/kafo_configure.rb:219:in `execute'
[2024-11-20T22:10:35.810Z]             from /usr/share/gems/gems/clamp-1.3.2/lib/clamp/command.rb:66:in `run'
[2024-11-20T22:10:35.810Z]             from /usr/share/gems/gems/kafo-7.6.0/lib/kafo/kafo_configure.rb:184:in `run'
[2024-11-20T22:10:35.810Z]             from /usr/share/gems/gems/clamp-1.3.2/lib/clamp/command.rb:140:in `run'
[2024-11-20T22:10:35.810Z]             from /usr/share/gems/gems/kafo-7.6.0/lib/kafo/kafo_configure.rb:54:in `run'
[2024-11-20T22:10:35.810Z]             from /sbin/foreman-installer:8:in `<main>'
```

ianballou: Maybe this will solve it: https://github.com/theforeman/foreman-installer/pull/999

PR #988, opened 2024-10-17
Do not use tilde specifier for PUPPET_VERSION read from environment

dosas:
when specifing a specific version this version should be used

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ gem 'rdoc', '< 6.4'
 
 gem 'kafo', '>= 7.3', '< 8'
 gem 'librarian-puppet', '>= 3.0'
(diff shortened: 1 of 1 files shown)


ekohl (changes requested):
:-1: on this because I don't see any good reason for this. IMHO you want both a lower and upper bound and `~>` gives you that (most of the time).

At first when I saw the summary I thought you'd simplify it to something like:
```ruby
gem 'puppet', ENV.fetch('PUPPET_VERSION', '~> 7.0')
```

That I could be open to since it gives complete control, but `<=` is just not the right way.

Also, it now breaks because this part doesn't work anymore:
https://github.com/theforeman/foreman-installer/blob/67702be227056c29bbe9e34089009d69bb3a6758/.github/workflows/test.yml#L28-L33
Suddenly `~> 8.0` turned into `<= 8.0` which means you no longer get the latest 8.x but just 8.0 or worse, 7.x.

ekohl:
ping @dosas

PR #990, opened 2024-10-18
Pass all PG settings as env vars

evgeni:
needs https://github.com/theforeman/kafo/pull/378

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,7 @@ source 'https://rubygems.org'
 # We don't want to list psych since that updates the bundled version
 gem 'rdoc', '< 6.4'
 
-gem 'kafo', '>= 7.3', '< 8'
+gem 'kafo', '>= 7.6', '< 8'
 gem 'librarian-puppet', '>= 3.0'
 gem 'puppet', ENV.key?('PUPPET_VERSION') ? "~> #{ENV['PUPPET_VERSION']}" : '~> 7.0'
 gem 'facter', '>= 3.0', '!= 4.0.52'
--- a/hooks/pre/10-reset_data.rb
+++ b/hooks/pre/10-reset_data.rb
@@ -57,7 +57,11 @@ def reset_candlepin
 end
 
 def pg_command_base(config, command, args)
-  "PGPASSWORD='#{config[:password]}' #{command} -U #{config[:username]} -h #{config[:host]} -p #{config[:port]} #{args}"
+  "#{command} -U #{config[:username]} -h #{config[:host]} -p #{config[:port]} #{args}"
+end
+
+def pg_env(config)
+  { 'PGPASSWORD' => config[:password] }
 end
 
 def pg_sql_statement(config, statement)
@@ -67,7 +71,7 @@ end
 # WARNING: deletes all the data owned by the user. No warnings. No confirmations.
 def empty_database!(config)
   delete_statement = 'DROP OWNED BY CURRENT_USER CASCADE;'
-  execute!(pg_sql_statement(config, delete_statement), false, true)
+  execute!(pg_sql_statement(config, delete_statement), false, true, pg_env(config))
 end
 
 def clear_pulpcore_content(content_dir)
--- a/hooks/pre/10-reset_data.rb
+++ b/hooks/pre/10-reset_data.rb
@@ -56,22 +56,24 @@ def reset_candlepin
   empty_db_in_postgresql('candlepin')
 end
 
-def pg_command_base(config, command, args)
-  "#{command} -U #{config[:username]} -h #{config[:host]} -p #{config[:port]} #{args}"
-end
-
 def pg_env(config)
-  { 'PGPASSWORD' => config[:password] }
+  {
+    'PGHOST' => config.fetch(:host, 'localhost'),
+    'PGPORT' => config.fetch(:port, '5432').to_s,
+    'PGUSER' => config[:username],
+    'PGPASSWORD' => config[:password],
+    'PGDATABASE' => config[:database],
+  }
 end
 
-def pg_sql_statement(config, statement)
-  pg_command_base(config, 'psql', "-d #{config[:database]} -t -c \"#{statement}\"")
+def pg_sql_statement(statement)
+  "psql -t -c \"#{statement}\""
 end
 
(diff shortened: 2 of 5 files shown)


hooks/pre/10-reset_data.rb
```diff
@@ -57,7 +57,11 @@ def reset_candlepin
 end
 
 def pg_command_base(config, command, args)
-  "PGPASSWORD='#{config[:password]}' #{command} -U #{config[:username]} -h #{config[:host]} -p #{config[:port]} #{args}"
+  "#{command} -U #{config[:username]} -h #{config[:host]} -p #{config[:port]} #{args}"
+end
+
+def pg_env(config)
+  { 'PGPASSWORD' => config[:password] }
```
ekohl: Should we pass all `PG*` params to make the command really simple?

evgeni: yes, I was just doing this :D

hooks/pre/10-reset_data.rb
```diff
@@ -56,18 +56,24 @@ def reset_candlepin
...
+    'PGUSER' => config[:username],
+    'PGPASSWORD' => config[:password],
+    'PGDATABASE' => config[:database],
+  }
 end
 
-def pg_sql_statement(config, statement)
-  pg_command_base(config, 'psql', "-d #{config[:database]} -t -c \"#{statement}\"")
+def pg_sql_statement(statement)
+  "psql -t -c \"#{statement}\""
```
ekohl, line 70: I'd be tempted to pass in the statement via stdin but wonder if that works well with runuser (if we do that)

hooks/pre/10-reset_data.rb
```diff
@@ -56,18 +56,24 @@ def reset_candlepin
   empty_db_in_postgresql('candlepin')
 end
 
-def pg_command_base(config, command, args)
-  "PGPASSWORD='#{config[:password]}' #{command} -U #{config[:username]} -h #{config[:host]} -p #{config[:port]} #{args}"
+def pg_env(config)
+  {
+    'PGHOST' => config.fetch(:host, 'localhost'),
```
ekohl, line 61: Do we need a default or can we trust psql to figure it out?

ekohl (commented):
Btw, if we don't need to ensure we've escaped the Puppet Ruby env we don't need to format the command and you could use `Open3` directly

hooks/pre/10-reset_data.rb
evgeni: runuser is only used for local sql, and there we don't need statements.

hooks/pre/10-reset_data.rb
evgeni: No idea, copied from your fm patch ;)

evgeni:
> Btw, if we don't need to ensure we've escaped the Puppet Ruby env we don't need to format the command and you could use `Open3` directly

Yeah,  but our helpers incorporate the escaping and I like helpers ;)

hooks/pre/10-reset_data.rb
ekohl: There I had it because `local?` compares to `localhost` and I didn't want to change that. Perhaps we should make it more robust so it can parse any postgresql url. I'd like to be able to use unix sockets and ident auth but now we'd break foreman-maintain

ekohl:
You can pass an array and avoid any escaping

hooks/pre/10-reset_data.rb
evgeni: Here we have `remote_host?(hostname)` which does something very similar.

The old code just assumed `config[:host]` is set (the command would be wrong if it were unset), which we can do here too?

hooks/pre/10-reset_data.rb
evgeni: that said, our `execute` helpers don't do stdin right now, so it'd need more refactoring.

evgeni:
> You can pass an array and avoid any escaping

I meant the "escaping from Puppet ruby env".

My point is:
* We have those helpers exactly to escape the Puppet ruby env, because we need it sometimes.
* They also have logging and stuff wired in, which is nice
* Our code is standardized on using those helpers
* I do not want to have a special code that does not use the helpers, just because I don't *need* puppet-env-escaping.

So if we can drop the "puppet-env-escaping" all together, cool (but I think we can't, as we call f-m and f-rake, which need system ruby), but scope-creep. Here I just want to pass in PG things via env :)

spec/postgresql_upgrade_hook_context_extension_spec.rb
```diff
@@ -70,12 +70,12 @@
 
     it 'runs postgresql-setup --upgrade' do
       expect(subject).to be_nil
-      expect(context).to have_received(:'execute!').with("runuser -l postgres -c 'PGSETUP_INITDB_OPTIONS=\"--lc-collate=en_US.UTF-8 --lc-ctype=en_US.UTF-8 --locale=en_US.UTF-8\" postgresql-setup --upgrade'", false, true)
+      expect(context).to have_received(:'execute!').with("runuser -l postgres -c 'PGSETUP_INITDB_OPTIONS=\"--lc-collate=en_US.UTF-8 --lc-ctype=en_US.UTF-8 --locale=en_US.UTF-8\" postgresql-setup --upgrade'", false, true, {})
```
evgeni, line 73: `runuser -l` clears the environment, so that's why I did *not* use the env stuff here too.

Gemfile
```diff
@@ -4,7 +4,7 @@ source 'https://rubygems.org'
 # We don't want to list psych since that updates the bundled version
 gem 'rdoc', '< 6.4'
 
-gem 'kafo', '>= 7.3', '< 8'
+gem 'kafo', github: 'theforeman/kafo'
```
evgeni: This needs updating once Kafo 7.6 is released with the feature (see https://github.com/theforeman/kafo/pull/380), but other than that this is ready for review

hooks/pre/10-reset_data.rb
ekohl: I think keeping a default for localhost is OK for now then.

ekohl (approved):
I would appreciate a Redmine issue for this.

evgeni:
@ekohl done

ekohl (approved):

PR #991, opened 2024-11-01
Make katello-certs-check openssl 1.1 (EL8) compatible again

frostygresh:
Removed openssl command option that isn't supported by openssl 1.1.1

The `-no-CAstore` option is not supported in EL8 and will cause foreman-installer to fail on EL8 servers.

### DIFF
--- a/bin/katello-certs-check
+++ b/bin/katello-certs-check
@@ -157,7 +157,7 @@ function check-priv-key () {
 function check-ca-bundle () {
     printf "Checking CA bundle against the certificate file: "
     ERROR_PATTERN="error [0-9]+ at"
-    CHECK=$(openssl verify -no-CApath -no-CAstore -CAfile $CA_BUNDLE_FILE -purpose sslserver -verbose $CERT_FILE 2>&1)
+    CHECK=$(openssl verify -no-CApath -CAfile $CA_BUNDLE_FILE -purpose sslserver -verbose $CERT_FILE 2>&1)
     CHECK_STATUS=$?
 
     if [[ $CHECK_STATUS != "0" || $CHECK =~ $ERROR_PATTERN ]]; then


evgeni:
cc @ehelms 

Can we make this dynamic based on the OpenSSL version?

ehelms:
> Can we make this dynamic based on the OpenSSL version?

I'm sure we can... how many version of OpenSSL are we dealing with?

evgeni:
I think just two: 1 (el8) and 3 (el9)

frostygresh:
```
version=$(openssl version | awk '{print $2}')
if [[ ${version:0:2} == "1." ]]; then
    CHECK=$(openssl verify -no-CApath -CAfile $CA_BUNDLE_FILE -purpose sslserver -verbose $CERT_FILE 2>&1)
else
    CHECK=$(openssl verify -no-CApath -no-CAstore -CAfile $CA_BUNDLE_FILE -purpose sslserver -verbose $CERT_FILE 2>&1)
fi
```

ehelms:
> I think just two: 1 (el8) and 3 (el9)

Huh... oh this is against 3.11. That makes sense, and the code snippet above works for me.

evgeni:
Yeah. But now that you spelled it out. Maybe just dropping that offending option from 3.11 (and 3.12!) is safer, avoiding any version parsing and then being super strict in 3.13+ that is el9 only?

Your call

ekohl:
In https://github.com/theforeman/foreman-installer/pull/986#issue-2581744235 I noted that I actually didn't mean to cherry pick this, but messed up. I thought it couldn't hurt, but clearly it does so I'm leaning to what @evgeni: revert 914c3cb35f499862be6e6a79833c9563937d90cc (3.12) and a25324c1948ddc4766b1724a18bc4dd47b267c46 (3.11). Version parsing is messy and can IMHO only cause more problems.

ekohl:
I've reverted it in c5cae20f82d5c0fcc8bb4e21d040a85455970bca and we'll work on releasing it soon. You can follow the meeting notes in https://community.theforeman.org/c/development/releases/20.

PR #994, opened 2024-11-13
Fixes #38012 - Handle Java versions newer than 8

ekohl:
Java < 11 uses version 1.X where X is the major while Java 11 drops the old 1. prefix.

It also refactors the code so it can be tested. I've only tested this using unit tests, but the `java -version` output was retrieved from a real CentOS Stream 9 system.

### DIFF
--- a/hooks/boot/01-kafo-hook-extensions.rb
+++ b/hooks/boot/01-kafo-hook-extensions.rb
@@ -157,6 +157,14 @@ module HookContextExtension
     end
     mountpoint[:available_bytes]
   end
+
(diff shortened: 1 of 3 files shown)


ehelms (approved):

Odilhao:
:eyes:

evgeni (approved):

jameerpathan111 (approved):
Verified the fix. I don't see `Detected Java 0 which is too old for Puppetserver 8` in installer log with patch applied (https://dashboard.packit.dev/jobs/copr/2014269)

PR #999, opened 2024-11-20
Refs #37883 - wrap evr remote checks in method

ianballou:
Attempts to fix the issue mentioned here https://github.com/theforeman/foreman-installer/pull/984/files#r1851069746

Wraps all the code in a method and calls the method at the end.

Untested since I'm posting this right at the end of the day in case it can merge early.

### DIFF
--- a/hooks/pre_commit/42-evr_extension_permissions.rb
+++ b/hooks/pre_commit/42-evr_extension_permissions.rb
@@ -2,35 +2,34 @@ def check_remote_evr_extension_permissions
   # Managed databases will be handled automatically.
   return if local_postgresql?
   return unless katello_enabled?
-  
+
   config = load_db_config('foreman')
-  
+
   # If postgres is the owner of the DB, then the permissions will not matter.
   return if config[:username] == 'postgres'
-  
+
   evr_existence_command = pg_sql_statement("SELECT 1 FROM pg_extension WHERE extname = 'evr';")
   logger.debug "Checking if the evr extension exists via #{evr_existence_command}"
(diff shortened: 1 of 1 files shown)


hooks/pre_commit/42-evr_extension_permissions.rb
```diff
@@ -1,38 +1,42 @@
...
+    return
+  when '1'
+    fail_and_exit("The evr extension is not owned by the foreman DB owner. Please run the following command to fix it: " \
+                  "UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='#{config[:username]}') WHERE extname='evr';")
+  else
+    fail_and_exit("Failed to check the ownership of the evr extension.")
+  end
 end
+
+check_remote_evr_extension_permissions
```
ekohl, line 41: I wonder if we should have guards here
```suggestion
check_remote_evr_extension_permissions if !local_postgresql? && katello_enabled?
```

ehelms: The guards are in the method currently -- https://github.com/theforeman/foreman-installer/pull/999/files#diff-690025616ae0636f98279c9e319051813fa73ee28f489541b4dd89b2662756acR3-R4

ekohl: I wasn't quite sure what's nicer. It probably doesn't really matter.

What does matter is the trailing whitespace failure that breaks RuboCop.

ehelms (approved):

ianballou:
Thanks for cleaning  the PR up @ekohl !

PR #1004, opened 2024-11-26
Add check for sha1 CA certificate

ehelms:


bin/katello-certs-check
```diff
@@ -249,6 +249,16 @@ function check-shortname () {
     fi
 }
 
+function check-ca-signing-algorithm () {
+    printf "Checking CA signing algorithm for sha1: "
+    CHECK=$(openssl x509 -noout -text -in $CA_BUNDLE_FILE | grep 'Signature Algorithm: sha1WithRSAEncryption')
+    if [[ $? == "0" ]]; then
+        error 4 "The server CA certificate has been signed with sha1 and will break installation."
```
evgeni: As with the maintain PR, I think the text should hint the user more to obtain a CA (!) that doesn't use SHA1

bin/katello-certs-check
```diff
@@ -249,6 +249,16 @@ function check-shortname () {
     fi
 }
 
+function check-ca-signing-algorithm () {
+    printf "Checking CA signing algorithm for sha1: "
+    CHECK=$(openssl x509 -noout -text -in $CA_BUNDLE_FILE | grep 'Signature Algorithm: sha1WithRSAEncryption')
```
ekohl: Should this be `grep -q` to swallow the output??
```suggestion
    CHECK=$(openssl x509 -noout -text -in $CA_BUNDLE_FILE | grep -q 'Signature Algorithm: sha1WithRSAEncryption')
```

bin/katello-certs-check
ehelms: "Please acquire a CA and server certificates signed with sha256 or greater." ?

evgeni: Something like that, yeah

ekohl: Don't forget to mention any intermediates. The whole chain needs to be free of SHA1.

ehelms: Check messaging now

ekohl (commented):
I just saw http://sven.stormbind.net/blog/posts/pkix_look_into_cert_chains/ as a nice way to inspect a full chain

ekohl (changes requested):
Please also test this where the chain is one certificate signed by SHA256 and one by SHA1. In reality I think the intermediate is more likely to be SHA256 with a SHA1 root. I think your current code only verifies the first certificate (intermediate) and it will miss the root, which the JIRA issue told us was the actual cause.

ehelms:
> I just saw http://sven.stormbind.net/blog/posts/pkix_look_into_cert_chains/ as a nice way to inspect a full chain

I'll take a look at this method - this isn't the first issue we've had with not checking the full chain as it's not been an easy thing to do.

ehelms:
Updated with a bundle and checks on the bundle.

ehelms:
@ekohl @evgeni Can you re-review please?

bin/katello-certs-check
```diff
@@ -249,6 +249,16 @@ function check-shortname () {
     fi
 }
 
+function check-ca-signing-algorithm () {
+    printf "Checking CA signing algorithm for sha1: "
+    CHECK=$(openssl crl2pkcs7 -nocrl -certfile $CA_BUNDLE_FILE  | openssl pkcs7 -print | grep algorithm | grep -q 'sha1WithRSAEncryption')
+    if [[ $? == "0" ]]; then
+        error 4 "The server CA certificate has been signed with sha1 and will break installation. Update the server CA certificate with one signed by sha256 or stronger."
```
ekohl: I struggle with phrasing, but I'd like to make it obvious the whole chain needs to be signed with SHA256.
This check only checks the `CA_BUNDLE_FILE` so it's not even checking the certificate itself. Would this help?
```suggestion
        error 4 "The file '$CA_BUNDLE_FILE' contains a certificate signed with sha1 and will break installation. Update the server CA certificate and its chain with one signed by sha256 or stronger."
```

bin/katello-certs-check
```diff
@@ -249,6 +249,16 @@ function check-shortname () {
     fi
 }
 
+function check-ca-signing-algorithm () {
+    printf "Checking CA signing algorithm for sha1: "
+    CHECK=$(openssl crl2pkcs7 -nocrl -certfile $CA_BUNDLE_FILE  | openssl pkcs7 -print | grep algorithm | grep -q 'sha1WithRSAEncryption')
```
ekohl: Let's quote the filename just to be sure.
```suggestion
    CHECK=$(openssl crl2pkcs7 -nocrl -certfile "$CA_BUNDLE_FILE" | openssl pkcs7 -print | grep algorithm | grep -q 'sha1WithRSAEncryption')
```

ekohl:
Tests aren't happy with my suggestions :)

ekohl (approved):
@evgeni mind giving it a final read through?

spec/fixtures/katello-certs-check/certs/.crt
evgeni, line 1: What is this file? :)

spec/fixtures/katello-certs-check/create_cert.sh
```diff
@@ -38,6 +38,29 @@ else
...
+fi
+
+CERT_NAME=foreman-sha1.example.com
+if [[ ! -f "$CERTS_DIR/$CERT_NAME.key" || ! -f "$CERTS_DIR/$CERT_NAME.crt" ]]; then
+  echo "Generate server certificate"
+  openssl genrsa -out $CERTS_DIR/$CERT_NAME.key 2048
+  openssl req -new -key $CERTS_DIR/$CERT_NAME.key -out $CERTS_DIR/$CERT_NAME.csr -subj "/CN=foreman.example.com"
+  openssl x509 -req -in $CERTS_DIR/$CERT_NAME.csr -CA $CERTS_DIR/$CA_SHA1_CERT_NAME.crt -CAkey $CERTS_DIR/$CA_SHA1_CERT_NAME.key -CAcreateserial -out $CERTS_DIR/$CERT_NAME.crt -days 3650 -sha256 -extfile extensions.txt -extensions extensions
+else
+  echo "Server certificate with bad SAN exists. Skipping."
```
evgeni: This doesn't sound like the correct message ;)

evgeni (approved):
nits in the test fixtures, but nothing serious

PR #1007, opened 2025-02-12
Add hammer foreman_resource_quota package

bastian-src:
This PR aims on integrating the hammer cli plugin for `foreman_resource_quota` into the foreman installer:

https://github.com/ATIX-AG/hammer-cli-foreman-resource-quota

Relates to this `foreman-puppet` PR: https://github.com/theforeman/puppet-foreman/pull/1209

### DIFF
--- a/config/foreman-answers.yaml
+++ b/config/foreman-answers.yaml
@@ -19,6 +19,7 @@ foreman::cli::kubevirt: false
 foreman::cli::openscap: false
 foreman::cli::puppet: true
 foreman::cli::remote_execution: false
+foreman::cli::resource_quota: false
 foreman::cli::ssh: false
 foreman::cli::tasks: false
 foreman::cli::templates: false
--- a/config/foreman.migrations/20250218135439_add_hammer_cli_resource_quota.rb
+++ b/config/foreman.migrations/20250218135439_add_hammer_cli_resource_quota.rb
@@ -1,0 +1,1 @@
+answers['foreman::cli::resource_quota'] ||= false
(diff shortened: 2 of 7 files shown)


bastian-src:
Could it be the case that the puppet-foreman PR must be merged first before this succeeds?

https://github.com/theforeman/foreman-installer/actions/runs/13283117473/job/37085732694?pr=1007#step:7:687

ehelms:
> Could it be the case that the puppet-foreman PR must be merged first before this succeeds?
> 
> https://github.com/theforeman/foreman-installer/actions/runs/13283117473/job/37085732694?pr=1007#step:7:687

Yep!

config/katello.migrations/20250212105233_add_hammer_cli_resource_quota.rb
ekohl, line 1: Please keep the timestamps the same as from `foreman.migrations` to indicate they're effectively the same migration.

ekohl (changes requested):
Minor comment, but overall :+1:. We will need a new release of puppet-foreman to make sure it actually works when we branch & release so I'll also get on that.

config/katello.migrations/20250212105233_add_hammer_cli_resource_quota.rb
bastian-src: Nice, thanks for the review! I've renamed the `katello.migrations` file to have the same timestamp as the `foreman.migrations`.

Let me know if I can do anything else to support you in the process!

ekohl (changes requested):
The file names don't follow the conventions. Could you please use the helper script:
```sh
./bin/create-migration add-hammer-cli-resource-quota foreman.migrations katello.migrations
```

Or if you want to directly add the content:
```sh
./bin/create-migration add-hammer-cli-resource-quota foreman.migrations katello.migrations <<EOF
answers['foreman::cli::resource_quota'] ||= false
EOF
```

bastian-src:
Used the helper script as suggested, thanks!

ekohl (approved):
Thanks!

PR #1012, opened 2025-02-24
Drop register-in-foreman option from install message output

ehelms:
The default value for `--foreman-proxy-register-in-foreman` is true in the puppet module, so this option is not needed in the output.


ekohl (commented):
:+1: for removal, but it's also mentioned in https://github.com/theforeman/foreman-installer/blob/develop/KATELLO.md. Even though we don't really maintain it, I'd prefer to remove any mentions. Once merged, we can revisit https://github.com/theforeman/foreman-documentation/pull/3080.

ekohl (approved):

PR #1011, opened 2025-02-24
use Ansible/Quadlet to deploy the main app

evgeni:
# what is this?

a playground to test out deploying a Foreman/Katello with the following "adjustments":
- certs generated by `puppet-certs`
- Foreman, Pulp, Candlepin, PostgreSQL, Redis deployed by `foreman-quadlet`
- Foreman Proxy deployed by `puppet-foreman_proxy`

# details, gory details

- this uses https://github.com/evgeni/pulp_smart_proxy instead of https://github.com/theforeman/smart_proxy_pulp -- see https://github.com/theforeman/foreman-quadlet/commits/pulp-smart for the details of that switch
- this uses a hacked up `foreman-quadlet` to allow using `puppet-certs` generated certs straight out of `/root/ssl-build` -- see https://github.com/theforeman/foreman-quadlet/commits/installer-certs for the details of this hack
- this removes any calls to `puppet-foreman` and `puppet-foreman_proxy_content` from the installer, which means no CLI among other things
- this uses [a rather hackish patch to `puppet-katello`](https://github.com/theforeman/puppet-katello/commit/frankenstein) to achieve the following:
    - generate all certs
    - deploy foreman-proxy certs (which was done by `puppet-foreman_proxy_content` in the past)
    - create some shims so that `puppet-certs` can deploy Foreman and Candlepin certs, even tho they are unused on the system
    - check out `foreman-quadlet` and run it's `deploy` playbook at "the right time" so that it sets up Foreman/Katello/Candlepin/Pulp is deployed for us

# problems

- many, probably
- one that I know of: for some reason foreman-tasks gets listed as failed, it gets fixed by `systemctl restart dynflow*`, so probably some ordering issue or something

# testing

```console
# foreman-installer --scenario katello --foreman-proxy-oauth-consumer-key abcdefghijklmnopqrstuvwxyz123456 --foreman-proxy-oauth-consumer-secret abcdefghijklmnopqrstuvwxyz123456 --tuning development
```

PR #1015, opened 2025-03-05
Deploy in three stages - certs, Ansible, puppet

ehelms:
This is built on top of https://github.com/theforeman/foreman-installer/pull/1011

To run this:

```
foreman-installer --scenario katello --foreman-proxy-oauth-consumer-key abcdefghijklmnopqrstuvwxyz123456 --foreman-proxy-oauth-consumer-secret abcdefghijklmnopqrstuvwxyz123456 --tuning development
```


ekohl (commented):
At this point I'm wondering if we shouldn't simply create a content proxy like scenario and then call the installer from Ansible.

ehelms:
> At this point I'm wondering if we shouldn't simply create a content proxy like scenario and then call the installer from Ansible.

That's the tactic @evgeni is exploring.

hooks/pre/40-exacto.rb
```diff
@@ -0,0 +1,41 @@
...
+system('dnf -y install ansible-core git-core podman python3-cryptography python3-libsemanage python3-requests bash-completion nmap python3.12-psycopg2 python3.12-requests')
+
+Dir.chdir('/opt') do
+  system('git clone https://github.com/theforeman/foreman-quadlet')
+end
+
+Dir.chdir('/opt/foreman-quadlet') do
+  system('git checkout installer-certs')
+  system('ansible-galaxy install -r requirements.yml')
+  status = system('ansible-playbook playbooks/deploy.yaml')
```
evgeni:
```suggestion
  status = system('ansible-playbook playbooks/deploy.yaml -e certificate_source=installer')
```

hooks/pre/40-exacto.rb
```diff
@@ -0,0 +1,41 @@
...
+MANIFEST
+
+deploy_manifest = <<~MANIFEST
+  class { 'certs::foreman_proxy':
+    deploy => true,
+  }
+MANIFEST
+
+apply_installer_puppet(generate_manifest)
+apply_installer_puppet(deploy_manifest)
```
evgeni: Ideally this would use #935 instead?
Also, why deploying Proxy certs already? The proxy is not yet installed so this should fail?

ehelms: You are likely right, I added this after I had run it a few times.

In this use case, I think https://github.com/theforeman/foreman-installer/pull/935 is probably overkill for what we need. Unless the tactic was to have the user run the three stages rather than the installer orchestrating it.

ehelms:
> > At this point I'm wondering if we shouldn't simply create a content proxy like scenario and then call the installer from Ansible.
> 
> That's the tactic @evgeni is exploring.

The Ansible approach -- https://github.com/theforeman/foreman-installer/pull/1016

hooks/pre/40-exacto.rb
ehelms: I added a deploy class (https://github.com/theforeman/puppet-certs/pull/449/files#diff-e6327449f108163297d1ce69ca08f3a4d8013b01d772a4be3aba628b223bd8c5) that now handles this within the puppet properly. It's added to the answers file and this part is dropped.

I think you could drop `katello` from your Ansible based version now, and rely on this same mechanism as well.

ehelms:
Testing steps:

```
dnf -y install https://yum.puppetlabs.com/puppet8/el/9/x86_64/puppet8-release-1.0.0-9.el9.noarch.rpm
dnf -y install https://yum.theforeman.org/nightly/el9/x86_64/foreman-release.rpm
dnf -y install https://yum.theforeman.org/katello/nightly/katello/el9/x86_64/katello-repos-latest.rpm
dnf -y copr enable packit/theforeman-foreman-installer-1015 rhel-9-x86_64
dnf install -y foreman-installer-katello
foreman-installer --scenario katello --foreman-proxy-oauth-consumer-key abcdefghijklmnopqrstuvwxyz123456 --foreman-proxy-oauth-consumer-secret abcdefghijklmnopqrstuvwxyz123456 --tuning development
```

PR #1022, opened 2025-03-25
Fixes #38313 - make case insensitive for SAN entries and  Subject CN on certificate

takaelucky:
katello-certs-check will error when the certificate has the same FQDN in CN and SAN, but one is uppercase and another is in lowercase.

With this change, both FQDN will lowercase and check will not fail


ekohl (commented):
I think in the past we did have code which didn't deal well with this so I wonder if we solved all instances

PR #1025, opened 2025-05-06
Require base64 for Ruby 3.4

ehelms:
I am testing if these fix things or not. I noticed Packit switched to Fedora 42 w/ Ruby 3.4 which is where the base64 requirement came from in the Gemfile.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,11 @@ gem 'facter', '>= 3.0', '!= 4.0.52'
 gem 'puppet-strings'
 gem 'rake'
 
+if RUBY_VERSION >= '3.4'
+  gem 'base64'
+  gem 'getoptlong'
+  gem 'syslog'
+end
 gem 'racc' if RUBY_VERSION >= '3.3'
 
 gem 'semverse', groups: [:development, :test]
--- a/Puppetfile
+++ b/Puppetfile
@@ -13,6 +13,9 @@ mod 'puppet/redis',                    '>= 8.5.0'
 # https://github.com/theforeman/puppet-puppet/#git-repo-support
 mod 'puppetlabs/vcsrepo',              '>= 5.2.0'
 
+# https://github.com/southalc/podman/pull/100
+mod 'puppet/systemd', '< 8'
+
 # Soft dependency of theforeman/foreman
 mod 'theforeman/iop_advisor_engine',   :git => 'https://github.com/theforeman/puppet-iop_advisor_engine'
 


Puppetfile
```diff
@@ -13,6 +13,9 @@ mod 'puppet/redis',                    '>= 8.5.0'
 # https://github.com/theforeman/puppet-puppet/#git-repo-support
 mod 'puppetlabs/vcsrepo',              '>= 5.2.0'
 
+#
+mod 'southalc/podman',   :git => 'https://github.com/bastelfreak/podman', :branch => 'systemd'
```
ekohl: This doesn't look like a good thing, especially considering we're branching soon.

I'll also note that originally I proposed to use https://github.com/voxpupuli/puppet-quadlets precisely because we can control the maintenance and the podman module maintainer was at times unresponsive.

I'd be tempted to pin `systemd` explicitly to an older version for now
```suggestion
# https://github.com/southalc/podman/pull/100
mod 'puppet/systemd', '< 8'
```

Gemfile
```diff
@@ -11,6 +11,7 @@ gem 'facter', '>= 3.0', '!= 4.0.52'
 
 gem 'puppet-strings'
 gem 'rake'
+gem 'base64'
```
ekohl:
```suggestion
gem 'base64' if RUBY_VERSION >= '3.4'
```

Puppetfile
ehelms: I can update with the pin.

On what module to use, I think as we noted before https://github.com/voxpupuli/puppet-quadlets does not have everything we need and is "quadlet" specific rather than having all the podman pieces we need now and in the future. How would you proceed with that?

ekohl: I noticed that and ended up suggesting the pin as a short term workaround. Let's review this after branching.

ekohl:
Argh, now puppet-puppetserver_foreman requires a newer version. You just can't win with this. https://github.com/theforeman/puppet-puppetserver_foreman/pull/52/files#r2075602218 and I'll check if we can lower that bound.

ehelms:
> Argh, now puppet-puppetserver_foreman requires a newer version. You just can't win with this. https://github.com/theforeman/puppet-puppetserver_foreman/pull/52/files#r2075602218 and I'll check if we can lower that bound.

Oh yes... I should have called this out more specifically as this was the one I saw in the logs that led me to open the PR.

Gemfile
```diff
@@ -12,6 +12,11 @@ gem 'facter', '>= 3.0', '!= 4.0.52'
 gem 'puppet-strings'
 gem 'rake'
 
+if RUBY_VERSION >= '3.4'
+  gem 'base64'
```
ekohl, line 16: This is likely https://github.com/puppetlabs/forge-ruby/pull/129

Gemfile
```diff
@@ -12,6 +12,11 @@ gem 'facter', '>= 3.0', '!= 4.0.52'
...
+  gem 'getoptlong'
```
ekohl, line 17: Did you figure out why we need this? We should make sure that dependency pulls it in.

ehelms: Puppet does -- https://github.com/puppetlabs/puppet/blob/e227c27540975c25aa22d533a52424a9d2fc886a/puppet.gemspec#L27

ekohl: That already lists and we list `puppet` so why do we need to list it again?

ehelms: Sorry, I linked to the wrong version, 7.34.0 is used in SRPM build and you will note this is missing there:

https://github.com/puppetlabs/puppet/blob/7.34.0/puppet.gemspec

ekohl: Should we move to using Puppet 8 instead of trying to fix it with Puppet 7? I'm a bit worried about other dragons hiding in there.

ehelms: That thought crossed my mind. I thought maybe we should get builds back on track with 7 and then investigate puppet 8 / openvox.

ekohl (approved):

PR #1026, opened 2025-06-11
Fixes #38492 - don't fail to detect local db if there is other output

evgeni:
Sometimes calling `runuser -l` produces output (from files in
/etc/profile.d), and that confuses the strict "output equals 1" check we
had. Instead let's test for a known substring to be present in the
output.


ehelms (approved):
Untested but seems logical.

PR #1028, opened 2025-06-23
Fixes #38516 - Reset Apache & Foreman Proxy on certs-reset

ekohl:
When the user passes `--certs-reset` the intention is clearly that all certificates must be updated. It was reported that adding `--certs-update-server` does work. This makes sure the [documented procedure](https://docs.theforeman.org/nightly/Installing_Server/index-katello.html#resetting-custom-ssl-certificate-to-default-self-signed-certificate-on-foreman_foreman) matches that reality.

### DIFF
--- a/hooks/pre/20-certs_update.rb
+++ b/hooks/pre/20-certs_update.rb
@@ -22,7 +22,7 @@ if module_enabled?('certs')
     hostname = param('certs', 'node_fqdn').value
   end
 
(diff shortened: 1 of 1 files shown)


evgeni (approved):

sayan3296:
Tested and can confirm that it worked great. Thanks much for the fix @ekohl

ekohl:
Thanks for confirming @sayan3296!

PR #1029, opened 2025-06-25
Fixes #38533 - Drop evr extension first when resetting data

evgeni:
Under certain circumstances (12-to-13 upgraded remote DB),
`DROP OWNED BY CURRENT_USER CASCADE;` fails with "cannot drop type evr_t
because extension evr requires it" because it tries to drop the type
before the extension. However, dropping the extension using
`DROP EXTENSION IF EXISTS evr CASCADE;` works just fine and then the
"DROP OWNED" works too.

### DIFF
--- a/hooks/pre/10-reset_data.rb
+++ b/hooks/pre/10-reset_data.rb
@@ -28,7 +28,9 @@ end
 
 # WARNING: deletes all the data owned by the user. No warnings. No confirmations.
 def empty_database!(config)
+  drop_evr_statement = 'DROP EXTENSION IF EXISTS evr CASCADE;'
   delete_statement = 'DROP OWNED BY CURRENT_USER CASCADE;'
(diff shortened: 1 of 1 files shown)


ekohl (commented):
You can pass mutiple statements in a single exec. Would that be better or worse?

evgeni:
You mean something like `execute!(pg_sql_statement(drop_evr_statement + delete_statement), false, true, pg_env(config))`?
In the happy case it'd be "better" (we saved 0.00001µs execution time), in the error case you get a less readable output of what failed, which is IMHO worse.

ekohl (approved):
I was thinking about connection reliability as well in case it's networked, but I suppose that's negligible compared to the better output when something does fail for other reasons.

ekohl (changes requested):
Since it's a bugfix, I am tempted to add a Redmine issue so we have a changelog.

evgeni:
Redmine added.

ekohl (approved):

PR #1031, opened 2025-07-29
Add puppet-iop

ehelms:


ehelms:
This will be required after https://github.com/theforeman/puppet-foreman/pull/1229 is merged

ehelms:
This changes the design of iop enablement to be at the top-level. Users would enable it with `foreman-installer --enable-iop`. This exists only on the katello scenario since it requires both Katello and foreman_rh_cloud.

ShimShtein:
LGTM. Didn't test it though

PR #1045, opened 2025-11-03
Add ovirt plugin to foreman-installer

nadjaheitmann:
Needs https://github.com/theforeman/puppet-foreman/pull/1243

### DIFF
--- a/spec/migrations/250317000666_remove_ovirt_spec.rb
+++ b/spec/migrations/250317000666_remove_ovirt_spec.rb
@@ -1,0 +1,37 @@
+require 'spec_helper'
+
+migration '20250317000666_remove_ovirt' do
+  scenarios %w[foreman katello] do
+    context 'foreman::compute::ovirt was enabled' do
+      let(:answers) do
+        {
+          'foreman::compute::ovirt' => { 'version' => 'installed' },
+        }
+      end
+
+      it 'remove compute resource ovirt' do
+        expect(migrated_answers).not_to(include("foreman::compute::ovirt"))
+      end
+
+      it 'add plugin ovirt' do
+        expect(migrated_answers["foreman::plugin::ovirt"]).to be true
+      end
+    end
+
+    context 'foreman::compute::ovirt was disabled' do
+      let(:answers) do
+        {
+          'foreman::compute::ovirt' => false,
+        }
+      end
+
+      it 'remove compute resource ovirt' do
+        expect(migrated_answers).not_to(include("foreman::compute::ovirt"))
+      end
+
+      it 'does not add plugin ovirt' do
+        expect(migrated_answers["foreman::plugin::ovirt"]).to be false
+      end
+    end
+  end
+end
--- a/config/foreman-answers.yaml
+++ b/config/foreman-answers.yaml
@@ -49,6 +49,7 @@ foreman::plugin::monitoring: false
 foreman::plugin::netbox: false
 foreman::plugin::omaha: false
 foreman::plugin::openscap: false
+foreman::plugin::ovirt: false
 foreman::plugin::proxmox: false
 foreman::plugin::puppet: true
 foreman::plugin::puppetdb: false
--- a/config/foreman.migrations/20250317000666_remove_ovirt.rb
+++ b/config/foreman.migrations/20250317000666_remove_ovirt.rb
@@ -1,2 +1,5 @@
-answers.delete('foreman::compute::ovirt')
+# install foreman_ovirt plugin if oVirt compute resource was enabled before
+compute_ovirt = answers.delete('foreman::compute::ovirt')
+answers['foreman::plugin::ovirt'] = compute_ovirt.is_a?(Hash) || compute_ovirt
+
 answers.delete('foreman::plugin::ovirt_provision')
--- a/config/foreman.migrations/20251103142459_add_ovirt_plugin.rb
+++ b/config/foreman.migrations/20251103142459_add_ovirt_plugin.rb
@@ -1,0 +1,1 @@
+answers['foreman::plugin::ovirt'] ||= false
--- a/config/katello-answers.yaml
+++ b/config/katello-answers.yaml
@@ -64,6 +64,7 @@ foreman::plugin::leapp: false
 foreman::plugin::monitoring: false
 foreman::plugin::netbox: false
 foreman::plugin::openscap: false
+foreman::plugin::ovirt: false
 foreman::plugin::proxmox: false
 foreman::plugin::puppet: false
 foreman::plugin::puppetdb: false
--- a/config/katello.migrations/20250317000666_remove_ovirt.rb
+++ b/config/katello.migrations/20250317000666_remove_ovirt.rb
@@ -1,2 +1,5 @@
-answers.delete('foreman::compute::ovirt')
+# install foreman_ovirt plugin if oVirt compute resource was enabled before
+compute_ovirt = answers.delete('foreman::compute::ovirt')
+answers['foreman::plugin::ovirt'] = compute_ovirt.is_a?(Hash) || compute_ovirt
+
 answers.delete('foreman::plugin::ovirt_provision')
--- a/config/katello.migrations/20251103142419_add_ovirt_plugin.rb
+++ b/config/katello.migrations/20251103142419_add_ovirt_plugin.rb
@@ -1,0 +1,1 @@
+answers['foreman::plugin::ovirt'] ||= false
--- a/spec/fixtures/pulpcore-migration-dont-use-content-plugins-on-upgrades/katello-answers-after.yaml
+++ b/spec/fixtures/pulpcore-migration-dont-use-content-plugins-on-upgrades/katello-answers-after.yaml
@@ -31,6 +31,7 @@ foreman::plugin::kubevirt: false
 foreman::plugin::leapp: false
 foreman::plugin::monitoring: false
 foreman::plugin::netbox: false
+foreman::plugin::ovirt: false
 foreman::plugin::proxmox: false
 foreman::plugin::puppet: false
 foreman::plugin::remote_execution: true
--- a/spec/fixtures/pulpcore-migration-rpm-only/katello-answers-after.yaml
+++ b/spec/fixtures/pulpcore-migration-rpm-only/katello-answers-after.yaml
@@ -31,6 +31,7 @@ foreman::plugin::kubevirt: false
 foreman::plugin::leapp: false
 foreman::plugin::monitoring: false
 foreman::plugin::netbox: false
+foreman::plugin::ovirt: false
 foreman::plugin::proxmox: false
 foreman::plugin::puppet: false
 foreman::plugin::remote_execution: true
--- a/spec/fixtures/pulpcore-migration/katello-answers-after.yaml
+++ b/spec/fixtures/pulpcore-migration/katello-answers-after.yaml
@@ -31,6 +31,7 @@ foreman::plugin::kubevirt: false
 foreman::plugin::leapp: false
 foreman::plugin::monitoring: false
 foreman::plugin::netbox: false
+foreman::plugin::ovirt: false
 foreman::plugin::proxmox: false
 foreman::plugin::puppet: false
 foreman::plugin::remote_execution: true


nadjaheitmann:
@evgeni Puppet 7 tests are failing because of missing dependency. I guess this is unrelated?

evgeni:
Rebase, I just merged the Puppet 7 drop.

I wonder if we should adjust the https://github.com/theforeman/foreman-installer/blob/develop/config/foreman.migrations/20250317000666_remove_ovirt.rb migration to enable the plugin if ovirt was enabled in the past.

@ekohl thoughts?

nadjaheitmann:
> I wonder if we should adjust the https://github.com/theforeman/foreman-installer/blob/develop/config/foreman.migrations/20250317000666_remove_ovirt.rb migration to enable the plugin if ovirt was enabled in the past.

This would force the oVirt plugin onto people who are just happy to remove it (even if it was enabled before).

If I remember correctly from the Puppet drop to foreman_puppet, the installation of the new plugin was optional and had to be done manually.

evgeni:
> This would force the oVirt plugin onto people who are just happy to remove it (even if it was enabled before).

Wasn't that the plan? But I can live with whatever, I don't use oVirt :)

nadjaheitmann:
> > This would force the oVirt plugin onto people who are just happy to remove it (even if it was enabled before).
> 
> Wasn't that the plan? But I can live with whatever, I don't use oVirt :)

I did not know THAT was part of the plan :) @sbernhard, any opinion :arrow_double_up: ?

sbernhard:
> > > This would force the oVirt plugin onto people who are just happy to remove it (even if it was enabled before).
> > 
> > 
> > Wasn't that the plan? But I can live with whatever, I don't use oVirt :)
> 
> I did not know THAT was part of the plan :) @sbernhard, any opinion ⏫ ?

I would set `answers['foreman::plugin::ovirt']` to the 'old' value of `foreman::compute::ovirt`

like: 
```
answers['foreman::plugin::ovirt'] = answers['foreman::compute::ovirt'] || false
```

evgeni:
`answers['foreman::compute::ovirt']` can be a Hash, so you probably want:
```ruby
compute_ovirt = answers.delete('foreman::compute::ovirt')
answers['foreman::plugin::ovirt'] = compute_ovirt.is_a?(Hash) ? true : compute_ovirt
```

nadjaheitmann:
> `answers['foreman::compute::ovirt']` can be a Hash, so you probably want:
> 
> ```ruby
> compute_ovirt = answers.delete('foreman::compute::ovirt')
> answers['foreman::plugin::ovirt'] = compute_ovirt.is_a?(Hash) ? true : compute_ovirt
> ```

So we will also backport it to 3.16, then? Otherwise this would not make sense because the migration has already run? :thinking:

evgeni:
Well, we can? :)

It's more a call for @ogajduse who does 3.16 releases, but I wouldn't be against it.

(And even if we do not, 3.15→3.17 upgrades are *possible*, even if we don't encourage them)

nadjaheitmann:
> Well, we can? :)
> 
> It's more a call for @ogajduse who does 3.16 releases, but I wouldn't be against it.
> 
> (And even if we do not, 3.15→3.17 upgrades are _possible_, even if we don't encourage them)

I heard he wants to do the release today, so that gives this a bit of pressure :)

evgeni:
The next point release will come ;)

nadjaheitmann:
> The next point release will come ;)

Then let's give it a shot... Any hint how I can properly test the installer change if I want to enable the plugin if the compute resource was enabled? Is it enough to add the flag to the answers.yaml file and see if the plugin gets installed?

ogajduse:
@nadjaheitmann @evgeni What are the pieces that need to be released?
* new release of puppet-foreman
* package the release

Question: Does https://github.com/theforeman/foreman-packaging/pull/12539 and the DEB counterpart, too, need to be cherry-picked to `/3.16` branches in foreman-packaging?

evgeni:
Yeah, we need:
- package fixes from https://github.com/theforeman/foreman-packaging/pull/12539 and https://github.com/theforeman/foreman-packaging/pull/12540
- puppet-foreman fixes from https://github.com/theforeman/puppet-foreman/pull/1243 (probably picked to a stable branch, as the next puppet-foreman release will be a breaking one)
- the migration update from here

I don't think we should rush to get this into todays release.

ogajduse:
Agreed with Evgeni.
I can release this as part of 3.16.2 once everything is ready. Feel free to ping me once things are ready, I'll do the release.

nadjaheitmann:
I have adopted the oVirt removal migration to install the new plugin if the compute resource was active before. I did some tests but it would be good if someone could double-check. 

I was thinking about removing the migrations that add the oVirt plugin but in case someone has already run the removal migrations, I guess we need to keep them.

nadjaheitmann:
@evgeni Do you have any concerns here? It's the only piece we are missing for the installer integration of the plugin :crossed_fingers:

evgeni:
Looks good to me on a first glance and I have no technical concerns, no.

And yes, you can't remove a migration, it might have been run already and no idea how much that would confuse kafo.

If you want, you could add a spec test like https://github.com/theforeman/foreman-installer/blob/develop/spec/migrations/20250930000666_remove_grub1_spec.rb or https://github.com/theforeman/foreman-installer/blob/develop/spec/migrations/20240715095211_normalize_dns_forwarders_spec.rb

(I wouldn't block on this, so just tell me if you want)

nadjaheitmann:
> Looks good to me on a first glance and I have no technical concerns, no.
> 
> And yes, you can't remove a migration, it might have been run already and no idea how much that would confuse kafo.
> 
> If you want, you could add a spec test like https://github.com/theforeman/foreman-installer/blob/develop/spec/migrations/20250930000666_remove_grub1_spec.rb or https://github.com/theforeman/foreman-installer/blob/develop/spec/migrations/20240715095211_normalize_dns_forwarders_spec.rb
> 
> (I wouldn't block on this, so just tell me if you want)

I added a test, thanks for the suggestion. I think it makes sense to have a test here.

spec/migrations/20250317000666_remove_ovirt.rb
```diff
@@ -0,0 +1,21 @@
...
+          'foreman_compute_ovirt' => {},
+        }
+      end
+
+      it 'remove compute resource ovirt' do
+        expect(migrated_answers.not_to(include("foreman_compute_ovirt")))
+      end
+
+      it 'add plugin ovirt' do
+        expect(migrated_answers.to(include("foreman_plugin_ovirt")))
```
evgeni:
```suggestion
        expect(migrated_answers['foreman::plugin::ovirt']).to be true
```

this is probably the better assertion here (as you check it's true, not only present)

spec/migrations/20250317000666_remove_ovirt.rb
```diff
@@ -0,0 +1,21 @@
...
+  scenarios %w[foreman] do
+    context 'tftp_dirs' do
+      let(:answers) do
+        {
+          'foreman_compute_ovirt' => {},
+        }
+      end
+
+      it 'remove compute resource ovirt' do
+        expect(migrated_answers.not_to(include("foreman_compute_ovirt")))
```
evgeni:
```suggestion
        expect(migrated_answers.not_to(include("foreman::compute::ovirt")))
```

I think you can assert on both, but I prefer the `::` as in the asnwers.

spec/migrations/20250317000666_remove_ovirt.rb
```diff
@@ -0,0 +1,21 @@
+require 'spec_helper'
+
+migration '20250317000666_remove_ovirt' do
+  scenarios %w[foreman] do
```
evgeni: I think you can extend this to foreman and katello, as the latter also has that migration

spec/migrations/20250317000666_remove_ovirt.rb
```diff
@@ -0,0 +1,21 @@
...
+    context 'tftp_dirs' do
+      let(:answers) do
+        {
+          'foreman_compute_ovirt' => {},
```
evgeni: The answers file I have here has:
```
foreman::compute::ovirt:
  version: installed
```

So:
```
      let(:answers) do
        {
          'foreman::compute::ovirt' => { 'version' => 'installed' }
        }
      end
```

spec/migrations/20250317000666_remove_ovirt.rb
```diff
@@ -0,0 +1,21 @@
+require 'spec_helper'
+
+migration '20250317000666_remove_ovirt' do
+  scenarios %w[foreman] do
+    context 'tftp_dirs' do
```
evgeni:
```suggestion
    context 'foreman::compute::ovirt was enabled' do
```

evgeni:
Probably worth adding a test with
```
      let(:answers) do
        {
          'foreman::compute::ovirt' => false
        }
      end
```
too?

(sorry for the escallation ;) )

nadjaheitmann:
> Probably worth adding a test with
> 
> ```
>       let(:answers) do
>         {
>           'foreman::compute::ovirt' => false
>         }
>       end
> ```
> 
> too?
> 
> (sorry for the escallation ;) )

Thanks for the review, I guess :see_no_evil: :heart:

spec/migrations/250317000666_remove_ovirt_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+          'foreman::compute::ovirt' => { 'version' => 'installed' },
+        }
+      end
+
+      it 'remove compute resource ovirt' do
+        expect(migrated_answers.not_to(include("foreman::compute::ovirt")))
+      end
+
+      it 'does not add plugin ovirt' do
+        expect(migrated_answers["foreman::plugin::ovirt"]).to be false
```
evgeni, line 33: this should fail if you pass `version: installed`, but it did not? confusing.

spec/migrations/250317000666_remove_ovirt_spec.rb
evgeni, line 1: this file needs to be called `spec/migrations/20250317000666_remove_ovirt_spec.rb`, without the `_spec` it's not executed

spec/migrations/250317000666_remove_ovirt_spec.rb
nadjaheitmann: Let me rename the file and see if the test fails

evgeni:
```
diff --git spec/migrations/20250317000666_remove_ovirt_spec.rb spec/migrations/20250317000666_remove_ovirt_spec.rb
index 97541f0..a13e4a4 100644
--- spec/migrations/20250317000666_remove_ovirt_spec.rb
+++ spec/migrations/20250317000666_remove_ovirt_spec.rb
@@ -10,7 +10,7 @@ migration '20250317000666_remove_ovirt' do
       end
 
       it 'remove compute resource ovirt' do
-        expect(migrated_answers.not_to(include("foreman::compute::ovirt")))
+        expect(migrated_answers).not_to(include("foreman::compute::ovirt"))
       end
 
       it 'add plugin ovirt' do
@@ -21,12 +21,12 @@ migration '20250317000666_remove_ovirt' do
     context 'foreman::compute::ovirt was disabled' do
       let(:answers) do
         {
-          'foreman::compute::ovirt' => { 'version' => 'installed' },
+          'foreman::compute::ovirt' => false,
         }
       end
 
       it 'remove compute resource ovirt' do
-        expect(migrated_answers.not_to(include("foreman::compute::ovirt")))
+        expect(migrated_answers).not_to(include("foreman::compute::ovirt"))
       end
 
       it 'does not add plugin ovirt' do
```

this and the rename made it work

spec/migrations/250317000666_remove_ovirt_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+      it 'add plugin ovirt' do
+        expect(migrated_answers["foreman::plugin::ovirt"]).to be true
+      end
+    end
+
+    context 'foreman::compute::ovirt was disabled' do
+      let(:answers) do
+        {
+          'foreman::compute::ovirt' => {},
```
evgeni:
```suggestion
          'foreman::compute::ovirt' => false,
```

an empty hash counts as enabled

evgeni (approved):

nadjaheitmann:
Thanks @evgeni ! If we want it to be part of 3.16.2, how is the best practice to proceed with the puppet-foreman change?

evgeni:
Irgh, right. We should've added a redmine for this to track it :see_no_evil: 

Anyway. I've created a `27-stable` branch in `puppet-foreman`. You pick your changes there, we can release a 27.3.0 with ovirt support for 3.16 (which currently has 27.1.0, 27.2.0 only has depdendency range bumps, which is fine)

nadjaheitmann:
> Anyway. I've created a `27-stable` branch in `puppet-foreman`. You pick your changes there, we can release a 27.3.0 with ovirt support for 3.16 (which currently has 27.1.0, 27.2.0 only has depdendency range bumps, which is fine)

Alright, here is the PR: https://github.com/theforeman/puppet-foreman/pull/1246 (puppet 7 tests failing) and here is the one for foreman-installer: https://github.com/theforeman/foreman-installer/pull/1047

PR #1046, opened 2025-11-05
Reset puppet package parameters if nil or empty

ehelms:
This results from changes in https://github.com/theforeman/puppet-puppet/pull/935

@evgeni I believe this should address the Debian upgrade issue since those two parameters are no longer optional.

### DIFF
--- a/config/foreman-proxy-content.migrations/251105202301-reset-puppet-package-params.rb
+++ b/config/foreman-proxy-content.migrations/251105202301-reset-puppet-package-params.rb
@@ -1,0 +1,13 @@
+if answers["puppet"].is_a?(Hash)
+  puppet = answers["puppet"]
+
+  if puppet.has_key?("server_package") &&
+     (puppet["server_package"].nil? || puppet["server_package"].empty?)
+    puppet.delete('server_package')
+  end
+
+  if puppet.has_key?("package") &&
+     (puppet["package"].nil? || puppet["package"].empty?)
+    puppet.delete('package')
+  end
+end
--- a/config/foreman.migrations/20251105194319_reset_puppet_package_params.rb
+++ b/config/foreman.migrations/20251105194319_reset_puppet_package_params.rb
@@ -1,0 +1,13 @@
+if answers["puppet"].is_a?(Hash)
+  puppet = answers["puppet"]
+
+  if puppet.has_key?("server_package") &&
+     (puppet["server_package"].nil? || puppet["server_package"].empty?)
+    puppet.delete('server_package')
+  end
+
+  if puppet.has_key?("package") &&
+     (puppet["package"].nil? || puppet["package"].empty?)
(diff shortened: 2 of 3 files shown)


config/foreman-proxy-content.migrations/251105202301-reset-puppet-package-params.rb
```diff
@@ -0,0 +1,15 @@
+if answers["puppet"].is_a?(Hash)
+  puppet = answers["puppet"]
+
+  if puppet.has_key?("server_package") &&
+     (puppet["server_package"].nil? || puppet["server_package"] == "" ||
```
evgeni: empty? works for strings too, so you could just do `.nil? || .empty?`

OTOH, setting it to nil when it already is nil seems odd, but then you have to check that it supports empty (or is not nil 🙃), so the above is probably still the nicest shortcut

evgeni (approved):
I'm fine with the explicit version or my shorter suggestion. Whatever you prefer.

config/foreman-proxy-content.migrations/251105202301-reset-puppet-package-params.rb
```diff
@@ -0,0 +1,15 @@
...
+      (puppet["server_package"].is_a?(Array) && puppet["server_package"].empty?))
+    puppet["server_package"] = nil
```
evgeni: Setting to `nil` is not sufficient. What did work was to use `puppet.delete('server_package')` instead.

evgeni (changes requested):

config/foreman-proxy-content.migrations/251105202301-reset-puppet-package-params.rb
```diff
@@ -0,0 +1,13 @@
+if answers["puppet"].is_a?(Hash)
+  puppet = answers["puppet"]
+
+  if puppet.has_key?("server_package") && (puppet["server_package"].empty? ||
+      (puppet["server_package"].is_a?(Array) && puppet["server_package"].empty?))
```
evgeni: That's not what I suggested. Now if `server_package` is `nil`, it will error out as `nil.empty?` doesn't exist.
```suggestion
  if puppet.has_key?("server_package") && (puppet["server_package"].nil? || puppet["server_package"].empty?)
```

ehelms: Ahh I got confused because of the line you commented on, I get you now.

evgeni: Blame the GitHub mobile app.

evgeni (approved):
Verified working with packit build foreman-installer-3.17.0-0.20251106131809781154.pr1046.860.geb9231d.el9.noarch

PR #1048, opened 2025-12-09
Update modules for 3.17 GA

evgeni:


Puppetfile
```diff
@@ -10,9 +10,9 @@ mod 'theforeman/dns',                  '~> 12.0.0'
 mod 'theforeman/puppetserver_foreman', '~> 4.3.0'
 mod 'theforeman/tftp',                 '~> 10.0.0'
 mod 'katello/candlepin',               '~> 17.0.0'
-mod 'theforeman/pulpcore',             '~> 13.2.0'
+mod 'theforeman/pulpcore',             '~> 13.3.0'
 mod 'theforeman/foreman',              '~> 28.0.0'
-mod 'theforeman/foreman_proxy',        '~> 29.0.0'
+mod 'theforeman/foreman_proxy',        '~> 29.1.0'
 mod 'theforeman/puppet',               '~> 22.0.0'
```
ogajduse: Shouldn't this one be bumped to 22.0.1? Similar bump for theforeman/iop.

evgeni: it can. it doesn't have to, but I'll do it to be sure.

ogajduse (approved):

Puppetfile
evgeni: (can because 22.0.0 to 22.0.1 is allowed by `~>` to be updated, while 29.0.0 to 29.1.0 needs to be explicitly allowed)

PR #1058, opened 2026-05-05
Fix gunicorn control socket migration for modern katello installs

zjhuntin:
## Summary

The migration added in 25e94b83 assumed `foreman_proxy_content` is a Hash, but `katello-answers.yaml` ships `foreman_proxy_content: true` (a boolean) as the default for all installs since ~2021. The `is_a?(Hash)` guard silently skipped every modern katello server, leaving gunicorn using `$HOME/gunicorn.ctl` (`/var/lib/pulp/gunicorn.ctl`) and causing SELinux AVC denials.

- Convert `true` to `{}` before the Hash guard so the migration applies to both legacy Hash-form and modern boolean-true installations
- Update `katello-answers.yaml` default to the Hash form with socket paths so fresh installs also get the fix without relying on the migration
- Add spec context covering the `foreman_proxy_content: true` case which would have caught this

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/config/katello-answers.yaml
+++ b/config/katello-answers.yaml
@@ -109,7 +109,10 @@ foreman_proxy::plugin::openscap: false
 foreman_proxy::plugin::remote_execution::script: false
 foreman_proxy::plugin::salt: false
 foreman_proxy::plugin::shellhooks: false
-foreman_proxy_content: true
+foreman_proxy_content:
+  puppet: false
+  pulpcore_api_control_socket_path: /run/pulpcore-api/gunicorn.ctl
+  pulpcore_content_control_socket_path: /run/pulpcore-content/gunicorn.ctl
 katello: true
 puppet: false
 apache::mod::status: false
--- a/config/katello.migrations/200611220455-dont-proxy-pulp-yum-to-pulpcore-on-upgrades.rb
+++ b/config/katello.migrations/200611220455-dont-proxy-pulp-yum-to-pulpcore-on-upgrades.rb
@@ -1,9 +1,0 @@
-FPC = 'foreman_proxy_content'.freeze
-PROXY_YUM = 'proxy_pulp_yum_to_pulpcore'.freeze
-
-def upgrade
-  answers[FPC] = {} unless answers[FPC].is_a?(Hash)
-  answers[FPC][PROXY_YUM] = false
-end
-
-upgrade if answers[FPC].is_a?(Hash) && answers[FPC][PROXY_YUM].nil?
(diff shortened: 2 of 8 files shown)


Odilhao (approved):

config/katello-answers.yaml
```diff
@@ -109,7 +109,12 @@ foreman_proxy::plugin::openscap: false
 foreman_proxy::plugin::remote_execution::script: false
 foreman_proxy::plugin::salt: false
 foreman_proxy::plugin::shellhooks: false
-foreman_proxy_content: true
+foreman_proxy_content:
+  puppet: false
+  proxy_pulp_yum_to_pulpcore: false
+  proxy_pulp_deb_to_pulpcore: false
```
ehelms: Where did these three above here parameters come from? I do not see them on https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/manifests/init.pp

zjhuntin: This is wrong, it was my agent doing stupid things. Will revert, good catch.

Odilhao: Looks like they are parameters that we never cleaned, stuff from Pulp 2 https://github.com/theforeman/foreman-installer/commit/c214354d4a2a22ba6e4fd8295cc310d34561f9a3

ehelms: I think we just drop the two migrations related to these because they are all built on the principle that if this is a hash and these parameters no longer apply:

config/katello.migrations/200611220455-dont-proxy-pulp-yum-to-pulpcore-on-upgrades.rb
config/katello.migrations/202011084146-dont-proxy-pulp-deb-to-pulpcore-on-upgrades.rb

So delete those two files

config/katello.migrations/180702133937-set-puppet-integration-answer.rb
```diff
@@ -1,4 +0,0 @@
-if answers['foreman_proxy_content'].is_a?(Hash)
```
ehelms, line 1: Keep this migration

zjhuntin: fixed, sorry about that

ianballou:
I tested this on a fresh Katello box using the staging RPMs. I added the COPR repo here and installed foreman-installer from it. Lastly I added back in the migration that was missing before the latest push:

```bash
  Success!
  * Foreman is running at https://toads.org
      Initial credentials are admin / toads
  * To install an additional Foreman proxy on separate machine continue by running:

      foreman-proxy-certs-generate --foreman-proxy-fqdn "$FOREMAN_PROXY" --certs-tar "/root/$FOREMAN_PROXY-certs.tar.gz"
  * Foreman Proxy is running at https://toads:9090

[root@toads ~]# ausearch -m avc
<no matches>

[root@toads ~]# grep control_socket /etc/foreman-installer/scenarios.d/katello-answers.yaml
  pulpcore_api_control_socket_path: "/run/pulpcore-api/gunicorn.ctl"
  pulpcore_content_control_socket_path: "/run/pulpcore-content/gunicorn.ctl"
```

ehelms (approved):

PR #1059, opened 2026-05-18
improved the spelling and grammar of hooks/boot/02-message-helpers.rb

PascalKont:
After installing multiple servers, i found the below message increasingly annoying and now i changed it and some other lines as well :)

`  * To install an additional Foreman proxy on separate machine continue by running:
`

### DIFF
--- a/hooks/boot/02-message-helpers.rb
+++ b/hooks/boot/02-message-helpers.rb
@@ -8,14 +8,16 @@ module MessageHookContextExtension
   end
 
   def dev_server_success_message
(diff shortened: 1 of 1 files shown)


ianballou (approved):
I haven't run this, but it looks fine to me.

chris1984 (approved):

chris1984:
Might be worth looking to see if there is bats test that check for wording, I am guessing downstream we assert the wording somewhere too.

ehelms (approved):

PR #1060, opened 2026-06-08
Update modules for 3.19.0

evgeni:


evgeni:
@ogajduse you made us do module updates for rc2 but never updated them here? :)

Puppetfile
```diff
@@ -4,16 +4,16 @@ mod 'puppetlabs/apache',               '>= 8.3'
...
+mod 'theforeman/dhcp',                 '~> 9.5.0'
+mod 'theforeman/dns',                  '~> 12.2.0'
 mod 'theforeman/puppetserver_foreman', '~> 4.3.0'
 mod 'theforeman/tftp',                 '~> 10.1.0'
 mod 'katello/candlepin',               '~> 18.1.0'
 mod 'theforeman/pulpcore',             '~> 14.2.0'
-mod 'theforeman/foreman',              '~> 29.2.0'
-mod 'theforeman/foreman_proxy',        '~> 30.1.0'
-mod 'theforeman/puppet',               '~> 22.3.0'
+mod 'theforeman/foreman',              '~> 29.3.0'
```
evgeni, line 13: forge is slow, so this is not yet installable

ogajduse:
> @ogajduse you made us do module updates for rc2 but never updated them here? :)

Good point. I wonder where in the procedure I missed this. And how can we make sure I (we) don't miss it again?

ogajduse (approved):

PR #1062, opened 2026-06-24
Fixes #39446 - Reject non-PEM content in katello-certs-check

Satellite-RedHat:
Certificate files exported from PKCS#12/PFX bundles may contain OpenSSL metadata between PEM blocks that Pulp rejects at sync time. Add an early PEM structure check for the server cert and CA bundle.

## Summary

- Added  `check-pem-content` to reject certificate files with content outside PEM CERTIFICATE blocks (e.g. PKCS#12 Bag Attributes from Windows PFX exports)
- Added test fixture and RSpec coverage
## Test cases

1. **PKCS#12 metadata in CA bundle** — `Bag Attributes` / `subject=` / `issuer=` between PEM blocks (`ca-bundle-bag-attributes.crt`), exit 12
2. **Comment lines in certificate files** — `# RETRAITE AC Intermédiaire TLS` before server cert and CA bundle PEM blocks, exit 12 on both `-c` and `-b`

- Prevents Pulp sync failures: `Must be PEM encoded X.509 certificate` and `(cadata should be an ASCII string or a bytes-like object)`

## Related issues

- https://projects.theforeman.org/issues/39446
- SAT-46871
- SAT-22472


ianballou:
Is this change still needed with https://github.com/pulp/pulpcore/pull/6533 merged?

Edit: perhaps it is, that PR seems to be a revert of the previous fix. The bugfix file was removed.

sayan3296:
One small thing:

When an external CA entity signs a cert, and the customer gets it along with the bundle, they mostly always have a few things mentioned ( in commented form ) aside from the actual cert blocks.

Sometimes it just shows the Subject and Issuer for each cert. And sometimes it just calls out the name of the Root and Intermediate CA. So this is pretty common. And usually do not cause issues unless untill some weird characters or information is present in those extra segments. 

If the goal is to detect them and inform end-users about fixing the certs, it may trigger a new support case each time this activity is done by someone. We either should include a user-facing  **suggestion of an awk/sed combination** that can fix this in the certs, or else have a KB readily available to share with CUs and have them sanitize the cert files on their own.

Satellite-RedHat:
Thanks for the feedback.

I still think #1062 is worth merging. Today katello-certs-check can pass files with non-PEM content between blocks (e.g. PKCS#12 Bag Attributes, bare subject=/issuer= lines). Those can fail later in production — Pulp sync was one example (#39446). Server/CA certs are used in many other places that do not sanitize PEM the same way.

On “comments are common and usually harmless”

Agreed that CA bundles from external CAs often include human-readable labels or #-prefixed comments. In many environments those never cause trouble — until they do, and the failure happens after install, often as an obscure sync error rather than at cert validation time.

I agree we should minimize support impact. Along with the validation, adding a short doc **Note** in the Katello/Satellite certificates guide on what a valid cert/CA bundle should contain (only BEGIN/END CERTIFICATE blocks and blank lines) should help.

ianballou:
Since this is more of a feature, this change is less likely to be backported. As such it might make more sense for it to happen in foremanctl https://github.com/theforeman/foremanctl since the future of our installation is containerized.
We can keep this PR around and merge both, but these days new installer functionality needs to go into both places (or just foremanctl).

sayan3296:
> Since this is more of a feature, this change is less likely to be backported. As such it might make more sense for it to happen in foremanctl https://github.com/theforeman/foremanctl since the future of our installation is containerized. We can keep this PR around and merge both, but these days new installer functionality needs to go into both places (or just foremanctl).

So @ianballou , we need the same fix to be populated in https://github.com/theforeman/foremanctl/blob/master/src/roles/certificate_checks/files/foreman-certificate-check right ?

@Satellite-RedHat jfyi ^^

ianballou:
> > Since this is more of a feature, this change is less likely to be backported. As such it might make more sense for it to happen in foremanctl https://github.com/theforeman/foremanctl since the future of our installation is containerized. We can keep this PR around and merge both, but these days new installer functionality needs to go into both places (or just foremanctl).
> 
> So @ianballou , we need the same fix to be populated in https://github.com/theforeman/foremanctl/blob/master/src/roles/certificate_checks/files/foreman-certificate-check right ?
> 
> @Satellite-RedHat jfyi ^^

That looks like the right file to start with, yep!

Satellite-RedHat:
@ianballou Same check is now in foremanctl: https://github.com/theforeman/foremanctl/pull/676

It ports `check-pem-content` into `foreman-certificate-check` with the same validation, fixtures, and unit tests.

PR #1067, opened 2026-07-21
Fixes #39546 - Fix glob expansion crash in katello-certs-check loops

sbernhard:
Wrap both DNS_LIST for-loops with set -o noglob / set +o noglob to prevent wildcard SANs (e.g. *.example.com) from being expanded as filesystem globs, causing a "too many arguments" error in the [ ] test.

### DIFF
--- a/bin/katello-certs-check
+++ b/bin/katello-certs-check
@@ -201,15 +201,18 @@ function check-cert-san () {
         success
         printf "Checking if any Subject Alt Name on certificate matches the Subject CN"
         SUBJECT_CN=$(openssl x509 -in $CERT_FILE -noout -subject -nameopt multiline|grep commonName|cut -d '=' -f 2|tr -d ' ')
+        set -o noglob
         for DNS in ${DNS_LIST}
           do
           DNS_VALUE="$( echo ${DNS//DNS:/} | tr -d ',')"
           if [ $SUBJECT_CN == $DNS_VALUE ]
           then
+            set +o noglob
             success
             return
           fi
         done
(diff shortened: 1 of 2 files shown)


sbernhard:
Steps to reproduce the issue:
Have wildcard cert with:
```
# openssl x509 -in rocky_custom_certs.crt -noout -subject -ext subjectAltName
subject=C=DE, ST=Bavaria, O=Company AG, OU=Test, CN=*.rocky.name.nice, emailAddress=test@company.de
X509v3 Subject Alternative Name: 
    DNS:*.rocky.name.nice
```

In the directory in which you start to run "katello-certs-check", create at least two files:
```
touch foo.rocky.name.nice
touch blub.rocky.name.nice
```

without the fix, it fails with

```
katello-certs-check: line 213: [: too many arguments

[FAIL]

The /root/rocky_custom_certs.crt does not have a Subject Alt Name matching the Subject CN
```

With this fix, it works.

sbernhard:
Want to have a look at this change @ehelms ?

ekohl (commented):
In `spec/katello_certs_check_spec.rb` there are tests for this with fixtures. Any chance you could add a test for this?

Also wondering: do we actually need globbing anywhere or can we disable it globally?

ehelms:
Generally looks fine to me -- I assume https://github.com/theforeman/foremanctl/blob/master/src/roles/certificate_checks/files/foreman-certificate-check needs the same treatment?

ehelms (approved):

PR #1068, opened 2026-09-17
Refs #39777 - Set registration admission control in tuning profiles

pablomh:
## Summary

- Adds `foreman::registration_admission_max` overrides to each tuning
  size profile, scaling the admission control pool with system capacity

| Profile           | Value |
|-------------------|------:|
| default (no profile) | 150 |
| medium            |   300 |
| large             |   600 |
| extra-large       |   800 |
| extra-extra-large |   800 |

Depends on theforeman/puppet-foreman#1292 which introduces the
`registration_admission_max` parameter (default: 150).

Values match [foremanctl tuning profiles](https://github.com/theforeman/foremanctl/tree/main/src/vars/tuning).

## Test plan

- [ ] Verify `satellite-installer --tuning medium` sets `registration_admission_max` to 300
- [ ] Verify each profile applies the expected value
- [ ] Confirm no tuning profile selected uses the puppet module default (150)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jakduch (commented):
Keeping this as a draft makes sense until the admission-control semantics in theforeman/puppet-foreman#1292 are resolved. The current values do not represent global Apache connection limits because `BalancerMember max` is scoped per httpd child process.

The RPM build itself is green. The Redmine check only fails because #39777 belongs to the Foreman project; it needs to be moved to the Installer project (or replaced with an Installer ticket). After moving it, this can use `Fixes #39777`.

jakduch:
The Installer ticket is now https://projects.theforeman.org/issues/39797. It replaces the misplaced Foreman #39777. Please retarget the PR title and commit from `Refs #39777` to `Fixes #39797`.

I would keep this draft until the admission-control semantics in theforeman/puppet-foreman#1292 are validated.

